Issue #21, opened 2016-07-15
Use logging framework


sebastianhaas:
I suggest log4js, but it's really up to you. Again, I would love to hear @petrosalema's input on this. Might be good to use the same technology.

I used log4js in various places and found it super intuitive.
This will close #9

oliver-h:
If it's easy to use, I agree on using a logging framework. For the node version of log4js, there are also typings available (which maybe doesn't make sense to use it, because there is less browser specific code in it...)

PR #55, opened 2016-08-24
Added loglevel as logging framework

sebastianhaas:
See #21 
Added [loglevel](https://www.npmjs.com/package/loglevel)

Easy to use:
```javascript
    log.trace(msg);
    log.debug(msg);
    log.info(msg);
    log.warn(msg);
    log.error(msg);
```
 You can get different loggers, set log levels application-wide or per logger, etc.

Will resolve both #21 and #9.

### DIFF
--- a/demo/demo.js
+++ b/demo/demo.js
@@ -93,7 +93,7 @@
                 canvas.width = width;
                 try {
                 sheet.render();
-                } catch (e) {};
+                } catch (e) {}
                 enable();
             }
         );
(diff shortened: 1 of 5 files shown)


sebastianhaas:
@bukaznik @matt-uib could you review and merge?
We should open another issue for adding proper logging and debug messages to opensheetmusicdisplay. Also, the output of internal classes like the MusicSheetReader is currently logged to the console. This should be redirected to our logging framework as well.

Issue #140, opened 2017-06-07
Can OSMD use vexflow in SVG mode?


joaocc:
The reason for the question is that it seems to be easier to build interactions if SVG mode is used instead of Canvas.
Thx

sebastianhaas:
Not at the moment, but it should be really easy to do. This is also one possible fix for #69.
I'm not so sure about whether it's _easier_ to interact with a static vector graphic than with an HTML5 canvas.
PRs welcome, this should really require just a few lines of code.

sebastianhaas:
SVG rendering was introduced in #154.

Issue #69, opened 2016-09-23
Vexflow Rendering is blurry on retina display


matt-uib:

sebastianhaas:
SVG support was added in #154, apparently also the blurriness of the canvas demo was removed. Still have to investigate why that is the case.

muratcorlu:
Maybe we should fix by setting canvas bigger and make it correct size by styling again: https://coderwall.com/p/vmkk6a/how-to-make-the-canvas-not-look-like-crap-on-retina

bneumann:
Is this only a retina display issue?

muratcorlu:
I think it's an issue on all high density screens.

matt-uib:
It works fine in SVG mode, which is now default mode.

PR #154, opened 2017-08-14
Adds SVG Rendering option

jportela:
In order to solve the "blurriness" on Retina displays, I've added an option to render the music drawer by using the SVG backend from VexFlow.

I've also encapsulated logic related to the backend (ie: fillRect, translate, etc) to a separate class, so different strategies for custom rendering can be implemented, for SVG and Canvas. To test this, go to the demo page and select the SVG backend: http://jportela-osmd.netlify.com/

Currently, there are a few outstanding issues, which I think make sense to solve before merging this:

* Text Styles are not rendered
* Re-rendering a SVG-backend score will duplicate the "rendering artifacts" (try using the demo page with SVG, and resize it)
* I haven't added any documentation explaining that the new backend could be used (once it's stable, I can write it)
* I haven't really tested this through with all the features OSMD offers, as I'm still fairly new to the library.

Let me know what you think, I'll try to fix the issues I stated above, but I'd appreciate your feedback on this solution.

This solves #69 and #140

### DIFF
--- a/demo/demo.js
+++ b/demo/demo.js
@@ -40,7 +40,8 @@
         nextCursorBtn,
         resetCursorBtn,
         showCursorBtn,
-        hideCursorBtn;
+        hideCursorBtn,
+        backendSelect;
 
     // Initialization code
     function init() {
@@ -59,6 +60,7 @@
         resetCursorBtn = document.getElementById("reset-cursor-btn");
         showCursorBtn = document.getElementById("show-cursor-btn");
         hideCursorBtn = document.getElementById("hide-cursor-btn");
+        backendSelect = document.getElementById("backend-select");
 
         // Hide error
         error();
@@ -87,7 +89,7 @@
         };
 
         // Create OSMD object and canvas
-        OSMD = new opensheetmusicdisplay.OSMD(canvas, false, "svg");
+        OSMD = new opensheetmusicdisplay.OSMD(canvas, false);
         OSMD.setLogLevel('info');
         document.body.appendChild(canvas);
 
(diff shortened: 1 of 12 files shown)


sebastianhaas:
This is so awesome, I love it. ❤️ It looks so clean and beautiful on my display, it's totally amazing.

I haven't had the time to get granular on the actual code changes, but it looks as you also improved the canvas quality? Did you increase the size of the canvas?

Re your issues, I can certainly help out on 3) and 4). Just let me know. I would love to see this in a release soon :)

matt-uib:
Looks also good to me - thanks for your engangement! Please just continue as you planned!

jportela:
Thanks for the feedback, I'll continue working on this, hopefully this will be ready for merge soon!

jportela:
@sebastianhaas @matt-uib text styles are now rendered properly for the SVG renderer, and I've also fixed the re-rendering issue. From navigating the demo page, everything looks spot on, but as I've said, there might be some more hidden features that are not working properly, I'd appreciate if you tested it a bit. Again you can look at: http://jportela-osmd.netlify.com/

For some reason, this also fixed the blurriness on the Canvas engine, which was my main motivation to use the SVG one. That's super strange, I'm not sure what I did to cause it, but if I did something, it was a nice side-effect ¯\_(ツ)_/¯

sebastianhaas:
@jportela I guess you increased the size of the canvas to be rendered on. If you compare it to the demo at http://opensheetmusicdisplay.com, it's almost twice the size. We would need to find a way to decide on the device's pixel density though, because now, in your solution, every device would have to render the canvas twice as large as needed and then scale it down again.

jportela:
@sebastianhaas I've checked and the difference happens because the demo page has a different CSS that limits the width of the page (to 896px) and a different zoom set (55%). The demo page lives in another repo, and I also updated it in this PR: https://github.com/opensheetmusicdisplay/opensheetmusicdisplay.github.io/pull/3

You can check what it looks like here: https://jportela-osmd-demo.netlify.com/demo/

sebastianhaas:
Thanks again for the great PR and sorry for the delay! I've just merged and I will publish a release asap.

Issue #109, opened 2017-03-17
Extend key signature enum to match MusicXML standard


sebastianhaas:
According to http://usermanuals.musicxml.com/MusicXML/MusicXML.htm#EL-MusicXML-mode.htm, the `mode` attribute on the element `key` in the traditional keys group can be one of the following values:
```
major, minor, dorian, phrygian, lydian, mixolydian, aeolian, ionian, locrian and none
```

Albeit non of these values besides major and minor are currently in use, `KeyEnum` at `KeyInstruction.ts:121` should reflect the MusicXML specs.

PR #170, opened 2017-10-13
Completing the Key Mode enum according to MusicXML

jportela:
This one solves: #109


sebastianhaas:
Thanks for the changes!

PR #178, opened 2017-11-14
Remove global dependency on rimraf 

vinzentt:
Hi,

When installing opensheetmusicdisplay on a new project, if we don't have rimraf installed, installation fails. I know my solution is UNIX only but maybe we could just keep the typings folder?

Thanks.

-- 
Vincent

### DIFF
--- a/package.json
+++ b/package.json
@@ -11,7 +11,6 @@
     "test": "npm run lint && grunt test",
     "tslint": "tslint --project tsconfig.json \"src/**/*.ts\" \"test/**/*.ts\"",
     "typedoc": "typedoc --out ./build/docs --name OpenSheetMusicDisplay --module commonjs --target ES5 --mode file ./src/**/*.ts",
-    "postinstall": "rm -rf typings",
     "prepublish": "grunt build:dist",
     "start": "http-server build/demo"
   },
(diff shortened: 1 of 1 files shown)


sebastianhaas:
Thanks for your input! Not sure I get this right, but `rimraf` is currently being [installed as a npm dependency](https://github.com/opensheetmusicdisplay/opensheetmusicdisplay/blob/develop/package.json#L90), so it should work without having to install it globally. In fact, I don't have it installed globally and `npm run postinstall` works just fine.

The only reason I added `rimraf` in the first place was to maintain cross-platform compatibility, so I'd rather not switch to `rm` here. 

However, if you could just remove that entire postinstall line and the `rimraf` dependency, I would be happy to accept your PR. The `typings` folder should not be present on any developers' machines anymore, so no need to keep this. It was just for the transition from typings to @types.

vinzentt:
Done.
Yes I saw rimraf in the dependencies but I still had the error "rimraf not found" when installing opensheetmusicdisplay from a project. And the only fix I saw was to add rimraf as a global package.
Anyway if it's not needed anymore let's remove this.

sebastianhaas:
Thanks for your time! I've just merged it.

PR #194, opened 2018-01-27
Added prefer-const rule and fixed all let variables

bneumann:
This one might be tricky. I touched so many files where "let" should have been "const" that it might come to merge conflicts among all branches. But right now the commit frequency is quite low so this would be a fantastic opportunity to do it :)


sebastianhaas:
Uuuuh I love this 😍 I will just merge it right away.

PR #198, opened 2018-02-11
Improvement/bounding box draw

bneumann:
Adding possibility to draw bounding boxes around graphical objects. Can be controled by environment variables over webpack, which is also another feature coming with this PR.
 also implemented the "style to css" converter which basically assigns randomly chosen colors (by me) to the id number.

### DIFF
--- a/demo/index.js
+++ b/demo/index.js
@@ -135,7 +135,7 @@ import { OSMD } from '../src/OSMD/OSMD';
             var value = e.target.value;
             // clears the canvas element
             canvas.innerHTML = "";
-            osmdObj = new opensheetmusicdisplay.OSMD(canvas, false, value);
+            osmdObj = new OSMD(canvas, false, value);
             osmdObj.setLogLevel('info');
             selectOnChange();
 
--- a/src/MusicalScore/Graphical/DrawingEnums.ts
+++ b/src/MusicalScore/Graphical/DrawingEnums.ts
@@ -1,3 +1,5 @@
+import Dictionary from "typescript-collections/dist/lib/Dictionary";
+
 /**
  * The supported styles to draw a rectangle on the music sheet
  */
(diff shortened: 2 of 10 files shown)


src/MusicalScore/Graphical/MusicSheetDrawer.ts
```diff
@@ -24,6 +24,7 @@ import {Instrument} from "../Instrument";
 import {MusicSymbolDrawingStyle, PhonicScoreModes} from "./DrawingMode";
 import {GraphicalOctaveShift} from "./GraphicalOctaveShift";
 import {GraphicalObject} from "./GraphicalObject";
+// import { unitInPixels } from "./VexFlow/VexFlowMusicSheetDrawer";
```
sebastianhaas: maybe remove this?

src/MusicalScore/Graphical/MusicSheetDrawer.ts
```diff
@@ -417,6 +419,28 @@ export abstract class MusicSheetDrawer {
...
         }
+        // Draw bounding boxes for debug purposes. This has to be at the end because only
+        // then all the calculations and recalculations are done
+        if (process.env.DRAW_BOUNDING_BOXES === "1") {
+            this.drawBoundingBoxes(page.PositionAndShape, 0, process.env.BOUNDING_BOX_TYPE);
+        }
+
+    }
+
+    private drawBoundingBoxes(startBox: BoundingBox, layer: number = 0, type: string = undefined): void {
```
sebastianhaas: You might want to add a docstring explaining the parameters, as they are not self-explanatory. At least to me.

src/MusicalScore/Graphical/DrawingEnums.ts
```diff
@@ -1,3 +1,5 @@
+import Dictionary from "typescript-collections/dist/lib/Dictionary";
```
sebastianhaas: We might want to use
```ts
import * as Collections from "typescript-collections";
...
new Collections.Dictionary<...>();
```

instead, as this is the preferred usage according to https://github.com/basarat/typescript-collections#usage.
If they change the directory structure of their bundle, we're doomed with that path-dependent imports :D

sebastianhaas:
@matt-uib @bneumann @sebastianhaas
**Please perform a squash merge with a commit message like:**

`feat: added bounding boxes for debug purposes`

bneumann:
OK I will incorporate your review and push again. Thank you.

webpack.config.js
```diff
@@ -36,6 +36,12 @@ module.exports = {
       $: 'jquery',
       jQuery: 'jquery'
     }),
+    new webpack.EnvironmentPlugin({
+      NODE_ENV: 'development', // use 'development' unless process.env.NODE_ENV is defined
+      DEBUG: false,
+      DRAW_BOUNDING_BOXES: '0',
+      BOUNDING_BOX_TYPE: 'GraphicalLabel', // Set to undefined for all bounding boxes otherwise put in the name of the object
```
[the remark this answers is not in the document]
bneumann: Loving it, will do!

src/MusicalScore/Graphical/VexFlow/VexFlowMusicSheetDrawer.ts
```diff
@@ -118,8 +118,8 @@ export class VexFlowMusicSheetDrawer extends MusicSheetDrawer {
      * @param layer is the current rendering layer. There are many layers on top of each other to which can be rendered. Not needed for now.
      * @param styleId the style id
      */
```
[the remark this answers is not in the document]
bneumann: whoops yes

demo/index.js
```diff
@@ -135,7 +135,7 @@ import { OSMD } from '../src/OSMD/OSMD';
             var value = e.target.value;
             // clears the canvas element
             canvas.innerHTML = "";
-            osmdObj = new opensheetmusicdisplay.OSMD(canvas, false, value);
+            osmdObj = new OSMD(canvas, false, value);
```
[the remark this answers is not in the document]
bneumann: Du to the change from grunt build to webpack.

src/MusicalScore/Graphical/MusicSheetDrawer.ts
bneumann: another whoops

src/MusicalScore/Graphical/DrawingEnums.ts
bneumann: Damn Visual Studio Code autoimporter. So damn convenient and such a b*** :)

src/MusicalScore/Graphical/MusicSheetDrawer.ts
```diff
@@ -417,6 +418,34 @@ export abstract class MusicSheetDrawer {
                 this.drawLabel(label, <number>GraphicalLayers.Notes);
             }
         }
+        // Draw bounding boxes for debug purposes. This has to be at the end because only
+        // then all the calculations and recalculations are done
+        if (process.env.DRAW_BOUNDING_BOX_ELEMENT !== undefined) {
```
sebastianhaas: Does this work? If `process.env.DRAW_BOUNDING_BOX_ELEMENT` is `false` as in the default configuration below, will this evaluate as intended? Webpack docs says you need to put `null` for an env variable to appear undefined.

bneumann: It actual works because false !== undefined is not true BUT the type check for object === false is not. So the funtion is called but does not render.
I will change it to null just to be sure

bneumann:
I just leave it to javascript to check if it is not false now. Instead of all the undefine, null, bla bla checks

src/MusicalScore/Graphical/VexFlow/CanvasVexFlowBackend.ts
```diff
@@ -52,11 +52,22 @@ export class CanvasVexFlowBackend extends VexFlowBackend {
...
+
+    /**
+     * Renders a rectangle with the given style to the screen.
+     * It is given in screen coordinates.
+     * @param rectangle the rect in screen coordinates
+     * @param layer is the current rendering layer. There are many layers on top of each other to which can be rendered. Not needed for now.
+     * @param styleId the style id
+     * @param alpha alpha value between 0 and 1
+     */
+    public renderRectangle(rectangle: RectangleF2D, styleId: number, alpha: number = 1): void {
```
sebastianhaas: Hopefully the last annoying comment now, I think you can omit the docstring here, as it will be inherited from the parent class(es). Typedoc should do so, IntelliSense in Visual Studio Code as well. Could you check that? If not, you might be able to put a `@inheritdoc`, but it should never be required to have the same docstring twice. So I guess, just put the comment in `VexFlowBackend` and everything should be fine?

bneumann:
Yeah I was wondering if it would be sufficient only in one place. I will
delete it later

Sebastian Haas <notifications@github.com> schrieb am Di., 13. Feb. 2018,
09:36:

bneumann:
Merge, merge, merge, merge!!! 👯‍♂️

sebastianhaas:
@matt-uib please review and squash merge with a commit message like mentioned in my previous comment :)

bneumann:
Bedankt

PR #212, opened 2018-03-04
Feature/instrument brace

bneumann:
Brackets are done. Also the "layering" works now. Bounding boxes should also be there but are not displayed with the debugging option. Still trying to figure out why.


bneumann:
@matt-uib don't forget to merge ;)

PR #222, opened 2018-03-22
Greenkeeper/webpack 4.2.0

bneumann:
I have updated all webpack related dependencies to the latest version. We could argue about the vexflow update but the 84 version has some nice fixes that we could use anyways.


bneumann:
Node version 5 is breaking the tests.

PR #251, opened 2018-05-13
Fix bug when endingIndices are undefined. Only add space for title on…

LearnedLately:
… first page.

### DIFF
--- a/src/MusicalScore/Graphical/EngravingRules.ts
+++ b/src/MusicalScore/Graphical/EngravingRules.ts
@@ -154,11 +154,11 @@ export class EngravingRules {
         // Staff sizing Variables
         this.pagePlacementEnum = PagePlacementEnum.Down;
         this.pageHeight = 100001.0;
-        this.pageTopMargin = 5.0;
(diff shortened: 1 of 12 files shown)


bneumann:
Hi, thank you for the fix. Can you enlighten me a bit about the bug it fixes? Code looks fine to me

LearnedLately:
Hi, yeah, apparently sometimes endingIndices are undefined, but the code checks for the 'length' property, causing an error. I noticed this when I made another change that created multiple pages.

matt-uib:
@LearnedLately, thanks very much for the investigation and the fix! The change in the calculator makes totally sense! For now we didn't really use the pages, as a page has currently a very large height and therefore only one is produced as all fits on it.
For the repetition fix: yes this is of course needed, thanks! i just think of one more thing that might be useful. Can you add a check that the two lengths are equal (if the arrays exist) so that the last check won't fail with a possible index-out-of-bounds error? That would be great and then we merge it :-)

sebastianhaas:
Changes were incorporated in 9c7e164

PR #265, opened 2018-06-01
Migration to loglevel for all logging statements

sebastianhaas:
Changed all missing debug and error outputs so that they print using the `loglevel `framework.

- How can we put OSMD components into logical groups in terms of logging? I can imagine something, where it would be possible to mute all errors and warnings coming from the MusicXML reading parts but displaying the ones that result from rendering etc.
- Do we really need class names etc in the logging statement as @bneumann pointed out? loglevel does not alter the source of the call so with sourcemaps activated, you can easily find the origin of the log statement.

### DIFF
--- a/src/MusicalScore/Graphical/BoundingBox.ts
+++ b/src/MusicalScore/Graphical/BoundingBox.ts
@@ -1,3 +1,4 @@
+import * as log from "loglevel";
 import {ArgumentOutOfRangeException} from "../Exceptions";
 import {PointF2D} from "../../Common/DataObjects/PointF2D";
 import {SizeF2D} from "../../Common/DataObjects/SizeF2D";
@@ -201,7 +202,7 @@ export class BoundingBox {
     public set Parent(value: BoundingBox) {
         this.parent = value;
         if (this.parent.ChildElements.indexOf(this) > -1) {
-            console.error("BoundingBox of " + (this.dataObject.constructor as any).name +
+            log.warn("BoundingBox of " + (this.dataObject.constructor as any).name +
             " already in children list of " + (this.parent.dataObject.constructor as any).name + "'s BoundingBox");
         } else {
             this.parent.ChildElements.push(this);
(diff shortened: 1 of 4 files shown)


src/MusicalScore/Graphical/BoundingBox.ts
```diff
@@ -201,7 +202,7 @@ export class BoundingBox {
     public set Parent(value: BoundingBox) {
         this.parent = value;
         if (this.parent.ChildElements.indexOf(this) > -1) {
-            console.error("BoundingBox of " + (this.dataObject.constructor as any).name +
+            log.warn("BoundingBox of " + (this.dataObject.constructor as any).name +
```
bneumann: why not error? This is actually a "this should not happen case"

src/MusicalScore/Graphical/VexFlow/VexFlowStaffEntry.ts
```diff
@@ -28,7 +29,7 @@ export class VexFlowStaffEntry extends GraphicalStaffEntry {
                 const staveNote: Vex.Flow.StaveNote = tickable as Vex.Flow.StaveNote;
                 tickablePosition += staveNote.getNoteHeadEndX() - staveNote.getGlyphWidth() / 2;
             } else {
-                console.log(tickable);
+                log.trace(tickable);
```
bneumann: Can be removed

src/MusicalScore/VoiceData/SourceMeasure.ts
```diff
@@ -214,7 +215,7 @@ export class SourceMeasure extends BaseIdClass {
                 }
             }
         }
-        //log.debug("created new container: ", staffEntry, this.verticalSourceStaffEntryContainers);
+        log.trace("created new container: ", staffEntry, this.verticalSourceStaffEntryContainers);
```
bneumann: Should be debug (trace always includes the whole stack which would result in shitloads of output information). BUT I think this should be removed also as we are done implementing voice entries and ghostnotes

matt-uib: yes, better remove it - we just needed it for debugging

sebastianhaas:
Hey @matt-uib and @bneumann sorry for the delay, I incorporated the changes you requested.

Would you like to comment on the two initial questions I asked :) I could offer some help here, but I guess we need to agree on an approach we want to take.

PR #277, opened 2018-06-26
chore(package): update webpack-cli to version 3.0.8

sebastianhaas:
Closes #266

@bneumann do you think we get in troubles with that?


bneumann:
I tried it recently and I had some problems. I will double check tomorrow

PR #296, opened 2018-07-20
chore(package): update eslint to version 5.0.1

sebastianhaas:
Closes #274 

> eslint now throws a deprecation warning for the "standard" file. I think we can use the 5.0.1 version!

@bneumann  So you're talking about:
```
(node:1696) [ESLINT_LEGACY_OBJECT_REST_SPREAD] DeprecationWarning: The 'parserOptions.ecmaFeatures.experimentalObjectRestSpread' option is deprecated. Use 'parserOptions.ecmaVersion' instead. (found in "standard")
```

I can live with that, maybe I'll ping https://github.com/standard/eslint-plugin-standard so that they can fix it? Do we need eslint-plugin-standard after all? Or would we rather use our own ruleset?

> do you agree updating to 5.0.1?

@sschmidTU If we can all live with the deprecation warning above, yes.

There is also
```
no-unused-variable is deprecated. Since TypeScript 2.9. Please use the built-in compiler checks instead.
```
which can't be fixed as it comes from the standards rule set too.

### DIFF
--- a/package.json
+++ b/package.json
@@ -63,7 +63,7 @@
     "clean-webpack-plugin": "^0.1.18",
     "cross-env": "^5.1.3",
     "cz-conventional-changelog": "^2.0.0",
-    "eslint": "^4.18.1",
+    "eslint": "^5.0.1",
     "eslint-config-standard": "^11.0.0",
     "eslint-plugin-import": "^2.9.0",
     "eslint-plugin-node": "^6.0.1",


sschmidTU:
review: ok, builds and runs demo fine

sebastianhaas:
what happend here?

sschmidTU:
@sebastianhaas this is already merged into develop by our [formatting merge](https://github.com/opensheetmusicdisplay/opensheetmusicdisplay/commit/e3fa44fa3b178db006c17a5c7f26c6980bcf1016).
@matt-uib and I were working on a branch that already had the eslint updated to 5.0.1. when we changed the formatting and decided to merge into develop.
I could have taken the eslint update out of the merge, but we have tested extensively with it and it seems to work fine. So, sorry for the unorthodox merge, but if all agree, the case seems to be indeed closed. I wanted to comment on this issue immediately, but have been busy.

PR #305, opened 2018-07-25
Fix key signature without mode, common/alla breve time

sschmidTU:
2 fixes:
fix(KeySignatureParse): key signature displayed correctly when mode not set in XML
fix(TimeSignatures): common time and alla breve display correctly

image

Previously, common time(C) was displayed as 4/4 and alla breve(C|) as 2/2,
and the D major signature in An die Musik was read as no key signature because mode attribute was missing:

image


sschmidTU:
@sebastianhaas I added tests for both fixes. The tests for time signature are in the key tests file, but if we want to separate those, we need quite some refactoring to make the useful helper functions in the key tests file available to other test files.

matt-uib:
To me as well! @sschmidTU please squash and merge it!

PR #315, opened 2018-08-01
Update vexflow to new version 1.2.85 (fixes bugs)

sschmidTU:
We finally got a new Vexflow release, and it fixes many layouting bugs:
* Delayed ornament x-shift
* Grace notes x-shift
* NoteSubGroup, In-Measure Clef x-shift
* (Dotted note dot y-shift, in previous master builds)

I used this version extensively, looked at most samples, and didn't notice breaking changes (that i didn't already fix).

(do npm install to use the new version)

### DIFF
--- a/package.json
+++ b/package.json
@@ -53,7 +53,7 @@
     "loglevel": "^1.5.0",
     "shortid": "^2.2.6",
     "typescript-collections": "^1.1.2",
-    "vexflow": "^1.2.84"
+    "vexflow": "^1.2.85"
   },
   "devDependencies": {
     "@types/chai": "^4.0.3",


sebastianhaas:
Great, I restarted the CI. As soon as it passes we can merge that. Please squash merge with

```
chore(package): update vexflow to version 1.2.85
```

BTW: If you remove the `^` in front of the version number in `package.json`, the new version won't be used automatically anymore and greenkeeper will pick it up and offer a PR exactly as you did here.

Keep in mind that greenkeeper will actually run a build with the new version as well, as anyone who runs `npm i` on our project will also get `1.2.85` even if we have `^1.2.84` in our `package.json`. In case that build fails, it will open an issue to track down the problem.

Sorry I was being verbose here, just wanted to make sure that everyone gets the mechanics of gk. :)

sschmidTU:
@sebastianhaas that's actually good to know. I'm still not quite sure about how to handle this in theory though. If i remove the ^ in front of the version number as you say, i also need to make a commit. and greenkeeper won't update to further versions until 2.0.
But i think it's fine for me to make my own PR for a critical version update.

sebastianhaas:
> If i remove the ^ in front of the version number as you say, i also need to make a commit. and greenkeeper won't update to further versions until 2.0.

Nah, if you remove the ^ gk will create a PR for any version change, also for patches and minors.

sschmidTU:
@sebastianhaas but i still have to make a commit removing the ^, not a PR, so that greenkeeper can detect it and make a PR, right? otherwise greenkeeper would need to be running locally on my machine. is it? ^^

sebastianhaas:
If you remove the `^`, gk will start detecting any version changes from that point on and, since none of them is in the auto-update range anymore, will create a PR for every single one of them.

Issue #314, opened 2018-08-01
Improve Lyrics Dashes


sschmidTU:
How about using fewer dashes in general? Setting minimumDistanceBetweenDashes to 10 instead of 5 makes the Hü-gel only use one dash instead of 5, and it looks better to me. I checked 4 physical sheet music editions (from IMSLP) and Musescore and they all only have one dash.
![image](https://user-images.githubusercontent.com/33069673/43515839-e090d736-9583-11e8-826a-a415efee0d3f.png)

Musescore also uses much fewer dashes:
![image](https://user-images.githubusercontent.com/33069673/43517612-a82964d4-9589-11e8-945c-198493b1fcf3.png)

which looks like this in our demo:
![image](https://user-images.githubusercontent.com/33069673/43517652-be2ee5a6-9589-11e8-808a-c8455d2ffa3a.png)

* **We can also improve dashes positioning** by making them equidistant between other dashes and lyrics, like in Musescore. Right now, the first dash starts very close to the word before, which is unnecessary and doesn't look good.

Issue #304, opened 2018-07-24
Optimize (decrease) lyrics spacing


sschmidTU:
the [lyrics spacing fix](https://github.com/opensheetmusicdisplay/opensheetmusicdisplay/commit/622f3465a152975db32c639e702e74578f31a424)
overreached on the issue of eliminating lyrics/labels overlap, and the extension of measure width we give in case of overlapping lyrics could be lowered a bit.

Before/After the lyrics spacing fix:
![lyrics_spacing_vorher_nachher](https://user-images.githubusercontent.com/33069673/43154826-f45fab52-8f75-11e8-84c6-742b24b92780.png)
(it's a pretty big Hü - - - - - gel now.)

Related to #299, more info there.

sschmidTU:
Actually, decreasing minLyricsSpacing pretty much handled this specific issue, now lyrics spacing causes two measures less to fit in one line instead of 4. One parameter, minLyricsSpacing is the core of this issue.
The rest of the issue depends not really on lyrics spacing, but on balancing measure widths (#299), and improving lyrics dashes (#314).
This issue could almost be closed. Though i might have to make another issue for the comma positioning.

sschmidTU:
The last commit fixed the comma spacing, improved lyrics spacing while (still) using less space so we only lose two measures in a line compared to no lyrics spacing, and reduced lyrics dashes (#314), PR to come.

PR #316, opened 2018-08-01
Fix Lyrics spacing and Dashes

sschmidTU:
Fixes:
* Space first lyrics in measure (see #304)
image
* Reduce dashes by increasing minimum distance between them as engraving rule (#314)
* Improve spacing (#304), add space for dashes between syllable, while still using little space, only showing 2-3 measures less on the first line compared to no lyrics spacing

Before:
image
After:
image

Only the last commit is new for this PR, the rest is staggered from PRs #313 and #315.
Those should probably be merged first. I wasn't sure what the best way to have these multiple PRs was.

### DIFF
--- a/demo/demo.css
+++ b/demo/demo.css
@@ -9,7 +9,7 @@ select {
 }
 
 #selectSample {
-  width: 250px;
+  width: 310px;
 }
 
 .bignum {
--- a/demo/demo.css
+++ b/demo/demo.css
@@ -9,7 +9,7 @@ select {
 }
 
 #selectSample {
-  width: 310px;
+  width: 320px;
 }
 
 .bignum {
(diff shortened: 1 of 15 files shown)


sebastianhaas:
> I wasn't sure what the best way to have these multiple PRs was.

Didn't look into the technical details of these changes, just wanted to offer an idea on how to solve the multiple PR problem. You could create a new feature branch which will host all the required elements for that feature and re-target your existing PRs so that they don't merge into `develop` but into the new feature branch. Then we get everything cleaned up and done there, and merge it into `develop` as soon as you're happy with the results :)

sschmidTU:
@sebastianhaas good idea, i'll try that next time.
for now i'll merge these PRs, as they've been up for some time and make huge improvements.
If someone should request changes in the future, i'll assign myself to implement them.

Issue #312, opened 2018-07-31
Canvas background color in saved image


kkjeer:
When the render type is set to "canvas", the user can right-click the canvas to save the image. However, the saved image has a transparent/gray background:
![example](https://user-images.githubusercontent.com/6687333/43438576-c188d5aa-9443-11e8-99a0-1697b0a4bf84.png)

A white background would match the expectation of sheet music. In the browser, this can be achieved by setting background: white on the container, but the saved canvas image doesn't have the white background. 

Is there a way to specify the background color of the canvas so that it gets saved into the resulting image?

sschmidTU:
Good catch, thanks for making the issue!
There should be a setting for this in Vexflow, i'm already looking at it, should be pretty easy to fix.

sschmidTU:
It seems to be more difficult than i thought. We should be able to simply set the background fill color in CanvasVexFlowBackend.initialize() as in this [demo](https://jsfiddle.net/7yaykcjp/8/) from [the Vexflow Tutorial](https://github.com/0xfe/vexflow/wiki/The-VexFlow-Tutorial):
```this.ctx.setBackgroundFillStyle("#eed");```
or
```(this.canvasRenderingCtx as any).setBackgroundFillStyle("#eed");```

But somehow we still get a transparent background with this.

It should be noted that VexFlow prefers SVGContext, and the CanvasContext is much less developed.

For now, you can save the canvas image and set the background to white in an image editor.
Or just take a screenshot. (though you'll lose some quality)

kkjeer:
@sschmidTU thanks for looking into this! Can the canvas draw a full screen white rectangle underneath the music? I tried a workaround that:

1. Creates a second canvas
2. Draws the contents of the music canvas onto the second canvas
3. Clears the music canvas
4. Draws a white rectangle onto the music canvas
5. Draws the contents of the second canvas onto the music canvas

This did set the image background to white; however, it also resulted in some odd sizing issues:
![example](https://user-images.githubusercontent.com/6687333/43504143-0deba3fe-9517-11e8-9026-df026a592b35.png)

sschmidTU:
@kkjeer thanks for experimenting with this as well!

I also tried drawing a white rectangle before the music is drawn, but found the color settings in vexflow and OSMD a bit convoluted. And drawing a white rectangle after clear() didn't do anything for me.

[...]

PR #318, opened 2018-08-03
Draw colored Background Layer

sschmidTU:
We now draw a colored rectangle in VexflowBackend.clear(), which we call in OpenSheetMusicDisplay.render().

This enables directly [right-click copying or saving the rendered score](https://user-images.githubusercontent.com/33069673/43636825-7ad672b6-9713-11e8-8dcb-3b222c558f5a.png) when using the Canvas backend, which is pretty convenient. Now we don't have to take a screenshot or edit image transparency anymore.

Previously, the canvas image had a transparent background:
image

Now it's white by default (saved as engraving rule).

The commits displayed here aren't that relevant, just look at the [File changed](https://github.com/opensheetmusicdisplay/opensheetmusicdisplay/pull/318/files). I'll squash the messages.

### DIFF
--- a/external/vexflow/vexflow.d.ts
+++ b/external/vexflow/vexflow.d.ts
@@ -203,6 +203,9 @@ declare namespace Vex {
             public setPosition(position: number): Modifier;
         }
 
+        export class NoteSubGroup extends Modifier {
+            constructor(notes: Object);
+        }
 
         export class StaveModifier extends Modifier {
             public static get Position() {
@@ -240,6 +243,8 @@ declare namespace Vex {
         
         export class ClefNote  extends Note {
             constructor(type: string, size: string, annotation: string);
+
+            public type: string;
         }
 
         export class Renderer {
--- a/external/vexflow/vexflow.d.ts
+++ b/external/vexflow/vexflow.d.ts
@@ -320,13 +320,28 @@ declare namespace Vex {
         }
 
         export class CanvasContext extends RenderContext {
-            public vexFlowCanvasContext: CanvasRenderingContext2D;
+            public vexFlowCanvasContext: any;
+            public background_fillStyle: { fill: string };
+            public setBackgroundFillStyle(style: string);
+        }
+
+        export class CanvasRenderingContext2D { 
+            public setBackgroundFillStyle(style: string): void;
+            public canvas: HTMLElement;
+            public fillStyle: Object;
+            public font: Object;
+            public globalAlpha: Object;
+            public globalCompositeOperation: Object;
+            public imageSmoothingEnabled: boolean;
+            public lineCap: Object;
+            public lineDashOffset: Object;
         }
 
         export class SVGContext extends RenderContext {
             public svg: SVGElement;
             public attributes: any;
             public state: any;
+            public setBackgroundFillStyle(style: string): void;
         }
 
         export class StaveConnector {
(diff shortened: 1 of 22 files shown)


src/OpenSheetMusicDisplay/OpenSheetMusicDisplay.ts
```diff
@@ -161,8 +161,13 @@ export class OpenSheetMusicDisplay {
         this.drawer.clear();
         this.drawer.resize(width, height);
         this.drawer.scale(this.zoom);
+
+        // clear and fill with background color
+        this.drawer.clear(0, 0, width, height);
```
bneumann: There are 2 clears. Why?

external/vexflow/vexflow.d.ts
```diff
@@ -321,12 +321,19 @@ declare namespace Vex {
 
         export class CanvasContext extends RenderContext {
             public vexFlowCanvasContext: CanvasRenderingContext2D;
+            // TODO CanvasRenderingContext2D isn't defined anywhere, and defining it leads to problems.
+            // this somehow works, though.
+            public background_fillStyle: { fill: string };
+            public setBackgroundFillStyle(style: string): void;
+            public clear(x: number, y: number, width: number, height: number): void;
         }
```
bneumann: I really don't like the idea that I need to know my canvas size to clear it.

src/OpenSheetMusicDisplay/OpenSheetMusicDisplay.ts
sschmidTU: the first clear actually only removes the SVG children in the SVG backend in this case. I left it in in case the resizing and scaling is slower if we don't remove the SVG children beforehand.
So, clear without arguments behaves like before, clear with position and dimensions makes a colored background (rectangle).
We could separate these methods, of course. But i don't think removing children from a list or drawing a single rectangle costs much performance.

external/vexflow/vexflow.d.ts
sschmidTU: perhaps you need to know it to fill it with the background color though. I know in any case that canvas.width and canvas.height is often wrong in the canvas backend, too small.

bneumann:
Alright rest looks good to me then

sschmidTU:
@bneumann so is there anything else i need to review? I checked SkyBottomLineCalculator and it works as before.

bneumann:
I can't think of anything else right now. That function is pretty low level but has not really been used before so should not be a biggie

sschmidTU:
@bneumann just asking because your review still requests changes. But i've noted your LGTM ;)

bneumann:
Si señor, problem is I cannot reopen the review on my phone

sschmidTU:
Seems like i broke the cursor, so right now this should not be merged. I'm fixing it among other things.

sschmidTU:
The cursor was broken because it's drawn on z-layer -1 in HTML, behind everything else, so it needs transparency on the next layers to be visible. So right now, we can't have the cursor and a colored background.
I thought about drawing the background on a deeper z-layer similar to the cursor, but the problem is that the cursor does not appear on the image if you copy it because it's not on the same canvas, so the background will have the same problem.
The comments in the Cursor class do say that it should be implemented differently. But it won't be easy to draw it as a background layer like it is now.

We can actually have Cursor and background right now if we move the Cursor to the z=0-layer, but then the notes won't paint over the cursor anymore:
image
This is the current state of the PR, but i'm not sure we want to make this tradeoff right now.

sschmidTU:
Currently implemented is an optional background color:
```js
osmd.setOptions({pageBackgroundColor: '#FFFFFF'})
```
this will make the cursor invisible.

This PR still has a better compromise, but always changes the look (see above).
(also see #312)

Issue #320, opened 2018-08-10
natural accidental not shown


sschmidTU:
It seems like all natural accidentals in xml are not shown right now.
e.g. in [this text-xml](https://github.com/opensheetmusicdisplay/opensheetmusicdisplay/files/2277899/accidental_test_natural_beethoven.zip), bottom version from musescore:
![image](https://user-images.githubusercontent.com/33069673/43959033-b022cb0a-9cad-11e8-968a-347100adab15.png)

This is from Beethoven - An die ferne Geliebte, though it seems to affect all accidentals, e.g. Bach Air Violins bar 3, c natural.

I will look into it.

sschmidTU:
It's a regression, naturals are shown in our last release demo.

PR #321, opened 2018-08-10
fix natural accidental not showing (#320)

sschmidTU:
This fixes the regression discussed in #320.

This PR comes in two alternative versions: Blue pill fix and Red pill fix.

Blue pill fix represented by the older commit 002d12a8d271f598e5f46c94c2c807948156b156 will simply eliminate the bug i introduced in 7032fdc484d7c3f28481feba92c8d9294e561b86. Everything will look rosy like before, and we will maintain the illusion that our AccidentalEnum.NONE is not ambiguous.

Red pill fix represented by the current PR commit 2ab12073b539faccd27b2f2eb95759e02093e3a9 adds an AccidentalEnum.NATURAL = 0 = AccidentalEnum.NONE, and it will show you that sometimes none is natural.

Make your choice. Or suggest another pill.

If you search the current develop branch for AccidentalEnum.NONE, you will see that for AccidentalCalculator and ChordSymbolContainer, NONE always means no accidental, while for VexFlowConverter and ArticulationReader it means NATURAL. So i think this should be made clear via the red pill fix.

I know that because of transposing and pitch calculation it's not always easy to differentiate between accidentals, especially natural/no accidental, but the blue pill version is what confused me into adding a wrong NATURAL enum in 7032fdc484d7c3f28481feba92c8d9294e561b86, causing the glitch in the matrix.

Note that the AccidentalEnums are also used for pitch calculation (sharp = 1, flat = -1), especially when transposing.

image


sschmidTU:
After discussion with @matt-uib we decided to refactor the AccidentalEnums to not be used directly for Pitch calculation anymore. This only worked because Typescript can use floats as Enum values, and even assign the same value to two Enum members (NONE = 0, NATURAL = 0).
Now we can differentiate between AccidentalEnum.NONE and AccidentalEnum.NATURAL, etc.
See 002d12a8d271f598e5f46c94c2c807948156b156

Issue #317, opened 2018-08-01
Draw LyricsExtends


sschmidTU:
We are currently not drawing lyrics extends, "long underscores" after lyrics:
![image](https://user-images.githubusercontent.com/33069673/43543871-94c13a5a-95d1-11e8-8d7b-2884f4e3160d.png)

This is the only one in Beethoven - An die Ferne Geliebte, but there are many more in the Mozart pieces and especially Schubert - An die Musik (search the xml for "extend").

We actually have MusicSheetCalculator.calculateLyricExtend(), which looks good. But we don't draw the GraphicalLines that we get out of it.
I added drawLyricLines() to MusicSheetDrawer in [ab34430](ab34430c2c3c9116e20071ecae0545ee60695431), but didn't implement it yet.

If you implemented calculateLyricExtend() or already tried drawing them, i'd appreciate help.
Otherwise i'll try to implement a drawGraphicalLine function similar to drawLabel in MusicSheetDrawer.

sschmidTU:
LyricExtends are now drawn correctly in the feature/lyricExtend branch:
![image](https://user-images.githubusercontent.com/33069673/44034272-fc0c13a8-9f0c-11e8-98e5-ed10b9487866.png)

x-wise they are drawn to the end of the following note-head, like in Musescore.

Now we can see how many extends some pieces have, and the difference between dashes and extends, dashes connecting syllables and extends for further notes on an already complete word (Schubert):
![image](https://user-images.githubusercontent.com/33069673/44034473-7501aec6-9f0d-11e8-807d-096e4c578245.png)

We sometimes get quite long extends, but that's because the spacing is quite wide then:
![image](https://user-images.githubusercontent.com/33069673/44034629-de782272-9f0d-11e8-9f2e-301ab9698d86.png)
Seems proportional to this physical edition:
![image](https://user-images.githubusercontent.com/33069673/44034668-f970270a-9f0d-11e8-8fd4-e0e51788f5cd.png)
By the way, physical sheet music editions basically do whatever they want with lyrics, dashes and extend formatting, it's interesting to compare them.

One potential refactor/architectural improvement is the location of the y and x-calculation. It would be nice to do it in the calculation methods in MusicSheetCalculator, but for now it seems we have to do it in MusicSheetDrawer:
(this is what fixed the positioning issue, see previous comment)
 ```
protected drawLyricLines(lyricLines: GraphicalLine[], staffLine: StaffLine): void {
[...]

PR #322, opened 2018-08-13
feat: draw lyricsExtends, fix: lyricsSpacing

sschmidTU:
LyricExtends are now drawn correctly, previously weren't at all:
image

see #317, which discusses some of the changes.

lyricsSpacing was reduced, less necessary as lyricsEntries are now correctly sorted before spacing
also contains some refactors and improved comments.

No breaking changes as far as i can see, tested all demo samples with lyrics.

### DIFF
--- a/src/MusicalScore/Graphical/EngravingRules.ts
+++ b/src/MusicalScore/Graphical/EngravingRules.ts
@@ -122,6 +122,7 @@ export class EngravingRules {
     private wedgeLineWidth: number;
     private tupletLineWidth: number;
     private lyricUnderscoreLineWidth: number;
+    private lyricUnderscoreLineWidthVexflow: number;
     private systemThinLineWidth: number;
     private systemBoldLineWidth: number;
     private systemRepetitionEndingLineWidth: number;
@@ -292,6 +293,7 @@ export class EngravingRules {
         this.wedgeLineWidth = 0.12;
         this.tupletLineWidth = 0.12;
         this.lyricUnderscoreLineWidth = 0.12;
+        this.lyricUnderscoreLineWidthVexflow = 1.3;
         this.systemThinLineWidth = 0.12;
         this.systemBoldLineWidth = EngravingRules.unit / 2.0;
         this.systemRepetitionEndingLineWidth = 0.12;
@@ -1020,6 +1022,12 @@ export class EngravingRules {
     }
     public set LyricUnderscoreLineWidth(value: number) {
         this.lyricUnderscoreLineWidth = value;
+    }
+    public get LyricUnderscoreLineWidthVexflow(): number {
+        return this.lyricUnderscoreLineWidthVexflow;
+    }
+    public set LyricUnderscoreLineWidthVexflow(value: number) {
+        this.LyricUnderscoreLineWidthVexflow = value;
     }
     public get SystemThinLineWidth(): number {
         return this.systemThinLineWidth;
--- a/src/MusicalScore/Graphical/EngravingRules.ts
+++ b/src/MusicalScore/Graphical/EngravingRules.ts
@@ -128,6 +128,7 @@ export class EngravingRules {
     private systemRepetitionEndingLineWidth: number;
     private systemDotWidth: number;
     private distanceBetweenVerticalSystemLines: number;
+    private noteHeadWidth: number; // estimated
     private distanceBetweenDotAndLine: number;
     private octaveShiftLineWidth: number;
     private octaveShiftVerticalLineLength: number;
@@ -275,7 +276,7 @@ export class EngravingRules {
         this.lyricsHeight = 2.0; // actually size of lyrics
         this.lyricsYOffsetToStaffHeight = 3.0; // distance between lyrics and staff. could partly be even lower/dynamic
         this.verticalBetweenLyricsDistance = 0.5;
-        this.horizontalBetweenLyricsDistance = 0.3;
+        this.horizontalBetweenLyricsDistance = 0.1;
         this.betweenSyllabelMaximumDistance = 10.0;
         this.betweenSyllabelMinimumDistance = 0.4;
         this.minimumDistanceBetweenDashes = 10;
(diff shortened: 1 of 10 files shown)


src/MusicalScore/Graphical/EngravingRules.ts
```diff
@@ -122,11 +122,13 @@ export class EngravingRules {
     private wedgeLineWidth: number;
     private tupletLineWidth: number;
     private lyricUnderscoreLineWidth: number;
+    private lyricUnderscoreLineWidthVexflow: number;
```
matt-uib: We should not have here extra parameters for Vexflow. Please just set the paramters simply such, that it works for Vexflow. If we later need it, we might add a method setEngravingRulesForVexFlow() or make VexFlowEngravingRules or something like that.

src/MusicalScore/Graphical/EngravingRules.ts
```diff
@@ -292,11 +294,13 @@ export class EngravingRules {
         this.wedgeLineWidth = 0.12;
         this.tupletLineWidth = 0.12;
         this.lyricUnderscoreLineWidth = 0.12;
+        this.lyricUnderscoreLineWidthVexflow = 1.3;
```
matt-uib: as i see it now: can there again be the factor 10 (pixels per unit) missing? Looks like it...

src/MusicalScore/Graphical/EngravingRules.ts
```diff
@@ -292,11 +294,13 @@ export class EngravingRules {
...
         this.systemThinLineWidth = 0.12;
         this.systemBoldLineWidth = EngravingRules.unit / 2.0;
         this.systemRepetitionEndingLineWidth = 0.12;
         this.systemDotWidth = EngravingRules.unit / 5.0;
         this.distanceBetweenVerticalSystemLines = 0.35;
+        this.noteHeadWidth = this.distanceBetweenVerticalSystemLines * 2; // estimated
```
matt-uib: This should actually not be needed, as we get the bounding box from Vexflow including the note heads. So you only need to consider the right border of the BB of the end staff entry. When i look at the rendering of the extend, it seems actually slightly too long (could be the 0.7 that you add)
We can still add this amount, but let's use a different/new name for it.

src/MusicalScore/Graphical/VexFlow/VexFlowMusicSheetCalculator.ts
```diff
@@ -144,20 +144,35 @@ export class VexFlowMusicSheetCalculator extends MusicSheetCalculator {
...
     let lastStaffEntryXPosition: number = 0;
+    // let lastLyricsEntryText: string = ""; // for easier debug purposes
+    // let lyricsEntryText: string = ""; // for easier debug purposes
+    // let lyricExtend: boolean = false; // for easier debug purposes
     let elongationFactorMeasureWidth: number = 1;
     for (const measure of measuresVertical) {
-      for (let i: number = 0; i < measure.staffEntries.length; i++) {
-        const staffEntry: GraphicalStaffEntry = measure.staffEntries[i];
+      // sort staffEntries by position, so that we always compare neighboring staffEntries
+      const staffEntriesSorted: GraphicalStaffEntry[] = measure.staffEntries.sort(
```
matt-uib: Have you tried, that it doesn't work without sorting? They should be sorted actually already..

src/MusicalScore/Graphical/EngravingRules.ts
sschmidTU: the problem is that the lyrics extend positioning is done in MusicSheetCalculator before any VexFlow drawing, so we don't have a bounding box for the notes yet. I checked that with a debugging session as well.
I could decrease the length of the extend. They seemed fine to me, but i'll check.

src/MusicalScore/Graphical/VexFlow/VexFlowMusicSheetCalculator.ts
sschmidTU: Yes, i checked that with a debugging session, they can be unsorted. For example in An die Musik, for "war-mer Lieb' ent-zun-den", i had the StaffEntries with lyricEntries ["war","mer","ent","Lieb'"] and so on.
I also thought they should be sorted, because that's done somewhere in the code, i'd have to look where that is. But i'm pretty sure the entries weren't sorted, and that's why i got much better spacing with much less minDistance (0.3 -> 0.1) after sorting.

matt-uib: interesting... maybe that's the reason for other problems. Let's keep an eye on it..

src/MusicalScore/Graphical/EngravingRules.ts
matt-uib: ..But how can you position the line and know it's length, when there was no VexFlow layouting?

sschmidTU: Good question, not sure. Maybe the Vexflow layouting is done by adding notes to voices and formatting them, but the notes aren't drawn yet. In any case, the GraphicalNote doesn't have valid BoundingBox information yet (which i checked by debugging).
But the width of the note head should pretty much always be the same, unless a rest note follows perhaps. The length of the extend also shouldn't be that critical, i think.

src/MusicalScore/Graphical/EngravingRules.ts
sschmidTU: done.

sschmidTU:
43efa1a63c4cac715dfef3188cf28ae5f094631:
* Removed the Vexflow engraving rule.
* Lyrics spacing is now done for each verse.
* Added the Austrian National Anthem to the demo samples (it's public domain).
It's interesting for spacing, i will add an issue for spacing the last lyric in a measure ("zukunftsreich"), maybe shouldn't go into this PR.

54af65a906e171b4cd81026162971afdc7a5f50c:
* sorting staffEntries/lyricEntries removed from lyrics spacing, not necessary after:
* fix double lyrics width calculation
(accidentally called in both MusicSheetCalculator and VexFlowMusicSheetCalculator)
* shortened lyric extend
not extended by noteHeadWidth anymore. Length is still fine, we have more room for spacing, and sheet music editions do all kinds of extend lengths anyways, mostly to the right edge of the next note head, sometimes to the middle of it.

@matt-uib i think this PR is ready now, i fixed all the issues you mentioned. We could decide not to include the Austrian National Anthem (public domain) in this PR, but it already looks pretty nice, except for the one interesting little spacing issue.

PR #332, opened 2018-08-22
fix(demo): fix console warning on init

sschmidTU:
Without the fix right now we get an error in the console when starting the demo, which is a bit annoying:
image
(see StackTrace at bottom)

This is because in index.js.init(), we call index.js.Resize(...,endCallback), which calls opensheetmusicdisplay.render(), before we loaded a sample.

Of course the demo doesn't have to be perfectly coded, but it's a bit strange to call new Resize() (index.js:194) on a function Resize(), which gets executed immediately, and the error is annoying. (We only recently started showing this error with console.warn() in index.js:133)

In any case, i used a sampleLoaded boolean and this fixes the error on initialization.

<details>
  <summary>
    StackTrace
  </summary>
```
Error: OpenSheetMusicDisplay: Before rendering a music sheet, please load a MusicXML file
    at OpenSheetMusicDisplay../src/OpenSheetMusicDisplay/OpenSheetMusicDisplay.ts.OpenSheetMusicDisplay.render (OpenSheetMusicDisplay.ts:137)
    at index.js:131
(anonymous) @ index.js:133
setTimeout (async)
Resize @ index.js:194
init @ index.js:123
(anonymous) @ index.js:277
load (async)
(anonymous) @ index.js:276
./demo/index.js @ index.js:314
__webpack_require__ @ bootstrap:19
2 @ demo.js:82981
__webpack_require__ @ bootstrap:19
(anonymous) @ bootstrap:83
(anonymous) @ bootstrap:83
webpackUniversalModuleDefinition @ universalModuleDefinition:9
(anonymous) @ universalModuleDefinition:10

### DIFF
--- a/demo/index.js
+++ b/demo/index.js
@@ -4,6 +4,7 @@ import { OpenSheetMusicDisplay } from '../src/OpenSheetMusicDisplay/OpenSheetMus
 (function () {
     "use strict";
     var openSheetMusicDisplay;
+    var sampleLoaded = false;
     // folder of the sample files
     var sampleFolder = process.env.STATIC_FILES_SUBFOLDER ? process.env.STATIC_FILES_SUBFOLDER + "/" : "",
     samples = {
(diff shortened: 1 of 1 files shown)


sschmidTU:
i think resize is broken in this PR, skips enable. don't merge yet.

sschmidTU:
@bneumann when resizing the window, the screen was stuck on being disabled, because i put a third function parameter into Resize(startCallback,endCallback). Just fixed it. Now i would approve this PR ;)
image

bneumann:
Yes go for it

Issue #333, opened 2018-08-23
add build .js files to repo


supersational:
Requesting that the `build/opensheetmusicdisplay.min.js` and `build/opensheetmusicdisplay.js` files be added to the repo.

This would allow easy linking, (and to different versions) using https://rawgit.com/ 

Hopefully a simple request, this is a fantastic project btw!

bneumann:
Hi, thanks for the request but usually build files should not be included in a repository.
Anyways there if there is a demand for pre-built libraries you can use a CDN service. I just checked and it seems jsdelivr has our npm package:

https://www.jsdelivr.com/package/npm/opensheetmusicdisplay

We should check if we can also provide it over cdnjs

sschmidTU:
Vexflow uses both rawgit and npmcdn for jsfiddles (interactive demos). They also host the builds on their website and include it in the repo. They're a bit all over the place.
Of course we don't need to follow Vexflow in that regard.

@supersational if you type in https://unpkg.com/opensheetmusicdisplay, it will direct you to https://unpkg.com/opensheetmusicdisplay@0.3.1/build/opensheetmusicdisplay.min.js.
That should be all you need, right?
By the way, we should be making a new release with big improvements soon.

supersational:
Thanks, that solves most of my problems! 
Is there any link to the unminified library? I'm getting errors rendering certain musicXML files and being able to link to either unminified source or a source mapped library would help debugging. I am currently injecting into pages I don't have control of so can't use webpack.

supersational:
looking into it I think it needs to go into the 'files' field, see https://docs.npmjs.com/files/package.json#files
```
"files": [
...
    "build/opensheetmusicdisplay.js",
```

should add it to unpkg

bneumann:
You are right that would be an easy fix

supersational:
Sweet, look forward to the new release!

sschmidTU:
@supersational I created a PR for this, #335. If you have any input, go ahead. This is your chance to influence the debug.js you'll receive ;)

sebastianhaas:
Would be interesting to learn about how people are actually using/including OSMD. I guess we should look into the cdn option anyway.
@bneumann let's discuss that in another issue? Afaik cdnjs, Google Hosted Libraries, Fastly, jsdelivr etc. all offer free hosting for Open Source libraries, we just need to apply there.

[...]

PR #335, opened 2018-08-24
chore(debug-release): add osmd.debug.js to release

sschmidTU:
As discussed in #333, it would be nice to have a non-minified debug release of OSMD, where people can inspect the code running in their installed node module of OSMD if they decide to use the debug.js.

For a new release, we can run `npm run stage`, which builds the minified and the debug release, like `grunt stage` in Vexflow. When we push the release to npm, the debug.js (2.8MB) should be included, along with the min.js (1.09MB).

An alternative to the debug release would be using the source map (3.8MB) and including it with the min.js in the release.

Vexflow releases a debug version too and it has helped me quickly test things by changing the debug release code in the node module folder.

I would have liked to selectively minify Vexflow (880KB) and/or other node modules (1.5MB total), but i didn't figure out how. The babel-minify-webpack-plugin has an exclude option, but i couldn't get it to work. I'm not really an expert on Webpack.

If you enable Webpack Visualizer in webpack.dev.js, you can [see space usage stats for the (debug) release](https://user-images.githubusercontent.com/33069673/44587772-3df1aa80-a7b4-11e8-943f-f179a1295b6d.png) in build/Statistics.html (12MB).

### DIFF
--- a/package.json
+++ b/package.json
@@ -16,6 +16,8 @@
     "build:doc": "cross-env STATIC_FILES_SUBFOLDER=sheets npm run build",
     "build:webpack": "webpack --progress --colors --config webpack.prod.js",
     "build:webpack-dev": "webpack --progress --colors --config webpack.dev.js",
+    "build:webpack-common": "webpack --progress --colors --config webpack.common.js",
+    "stage": "npm run build && npm run build:webpack-dev",
     "start": "webpack-dev-server --progress --colors --config webpack.dev.js",
     "fix-memory-limit": "increase-memory-limit"
   },
@@ -25,6 +27,7 @@
   "files": [
     "build/dist/src",
     "build/opensheetmusicdisplay.min.js",
+    "build/opensheetmusicdisplay.debug.js",
     "AUTHORS",
     "CHANGELOG.md",
     "README.md",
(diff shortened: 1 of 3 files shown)


sschmidTU:
The only problem is that the map files are 2x3.5MB and everyone installing our package via npm will need to download more than 7MB extra. Our current release (0.3.1) npm package is 1.2MB.
Maybe it's enough just to have the min.js and its source map in the release, which would be 5.07MB. Still a pretty big size increase. Webpack is ~1MB, Vexflow is ~12MB.

supersational:
If size is a concern (fair enough) then I would be happy with just the `opensheetmusicdisplay.min.js.map` since that would link to the `dist` folder which is already included. You could even omit the `opensheetmusicdisplay.debug.js` files in that case.

supersational:
`.debug.js` files will still be useful for those debugging on older browsers without source-maps, but admittedly not worth including in the npm package.

sschmidTU:
Yes. min.js + map seems good. I'm just not sure how to consider the size of our npm package, i'll have to talk to the other contributors. Personally, i'd be fine with an 8MB package size.

sebastianhaas:
I left a comment on #333.

Will check other Typescript libraries that are delivered as minified JS over npm and CDNs such as Angular or rxjs and see how they tackle this issue.

I tend to think that a non-minified version will probably only be used in non-npm scenarios where people just include OSMD via script tags from our CDN-to-be and then may want to replace it with the non-minified version for debugging. In npm scenarios, project setup and debugging is usually more sophisticated and complex.

Issue #355, opened 2018-08-29
TextAlignment and BoundingBoxes


sschmidTU:
There seems to be a general issue for alignments of GraphicalLabels and how we deal with their position and coordinates. We currently do not modify the position of the GraphicalLabel's bounding box, but simply shift the screen coordinates that we got from the label's original position. This example is not from me, though i followed it for left-alignment:
```javascript
const screenPosition: PointF2D = this.applyScreenTransformation(graphicalLabel.PositionAndShape.AbsolutePosition);
(...)
case TextAlignmentEnum.CenterTop:
                screenPosition.x -= <number>bitmapWidth / 2;
```
from MusicSheetdrawer.ts/drawLabel()

For center alignment, this seems to work out for the bounding boxes, though i'm not sure why.
![image](https://user-images.githubusercontent.com/33069673/44810946-9f11e780-abd3-11e8-99b3-2dd21d152782.png)

For left-alignment, in order for lyrics to appear left-aligned, i had to shift them by about 1 unit to the left (our coordinates, 10 pixels in Vexflow).
```javascript
case TextAlignmentEnum.LeftBottom:
  screenPosition.y -= bitmapHeight;
  // without the following, lyrics are not optically left-aligned. influences lyrics, dash positioning.
  // TODO unify alignment shifts and our bbox position, which does not correspond to screenposition
  screenPosition.x -= unitInPixels;
```

This means that our label bounding boxes do not fit what's on the screen for left-alignment:
![image](https://user-images.githubusercontent.com/33069673/44809593-6f60e080-abcf-11e8-883f-9e65261f5f8f.png)

I'll try to fix this by shifting the label's position instead of the screenPosition. I just didn't figure out how to do that yet. The RelativePosition of the label gets overwritten at some point (tried setting it to x=-1), and i'm not sure i should set the AbsolutePosition in the layout calculation.

sschmidTU:
I changed back the placement of non-lyrics (LeftTop, LeftCenter) to how they were before (c13432c03e6b4d685f2b4eda870e1e0bb3a5b3c3). It affected the placement of the Instrument name and the lyricist (see Land der Berge), and i'm not sure that's intended.
This is because lyrics were not optically left-aligned with the settings we also used for labels like instrument name. For now, LeftBottom (lyrics) gets more x-shift (-1) again.

[...]

PR #356, opened 2018-08-29
Lyrics left aligned instead of center, fix lyricsSpacing

sschmidTU:
Lyrics are left-aligned now, which makes spacing easier, and aligns multiple verses:
image

I discuss a lot of this in a new issue #357, because it might be relevant in the future.

One thing still needs to be changed, the bounding boxes for labels are not correct at the moment. I made an issue for that too, #355. [fixed]

I would suggest that here we can discuss specific parts of the code in these commits, the rest in the issues.

I made further changes and general improvements to lyric spacing also for center alignment, which you can see in the commit message and code of 57ec2c011850a3fb78217afa99e71c21911449d1.


samuelbradshaw:
I think it might look nice if lyrics were left-aligned at the beginning of each line (beginning of each system, or beginning of each lyric phrase, or both), but centered in between.

Issue #360, opened 2018-08-30
Add Rerender button to demo


sschmidTU:
I'd like to add a rerender button to the demo. It would often be useful for debugging.

Right now for a rerender we would have to drag&drop an xml into the browser, resize the window, or type window.osmd.render() or something in the browser console, though that doesn't work for me.

E.g. in #300 i used zoom to trigger a rerender, which isn't ideal.

sschmidTU:
done.

Issue #106, opened 2017-02-25
Angular 2 Module


sebastianhaas:
Create an Angular 2 module that is

- installable over npm,
- exports opensheetmusicdisplay as an Angular 2 component
- takes xml/mxl data as `@Input`

bohoffi:
IMHO the process to use OSMD right now in an Angular application isn't a very big deal. One could easily create a custom component and include all the stuff required to get OSMD to work. So IMHO I don't see the requirement to make OSMD available as an Angular module.

But 1 scenario could be that the module import (`OSMDModule.forRoot()`) takes some kind of config.

Could you explain your thoughts on the PROs using OSMD as an Angular module?

sebastianhaas:
> OSMD right now in an Angular application isn't a very big deal.

I agree.

> One could easily create a custom component and include all the stuff required to get OSMD to work.

I agree too.

The reason/motivation for having an Angular (2.x-4.x) module is simply to address a greater number of people and to lower entry barriers. I think that, once set up, maintaining such a module does not cause a big workload plus it might be useful for me to discover (and resolve) challenges when using OSMD withing Angular applications before others have to do so :) The same argumentation goes for the wordpress plugin #25.

However, as we advance further with this project, by using ES6 imports Angular might be able to remove not required parts of OSMD during bundling (tree shaking). The same goes for AOT compiled modules. Modules ready for AOT compiling need to be published with some fancy module descriptors, so you need to compile stuff with ngc rather than tsc etc. so Angular might be able to optimize OSMD, but that probably requires a certain setup.
I just thought it wouldn't hurt to have it done once in a centralized manner, rather than lots of people having to do it on their own stumbling over the same problems over and over again.

Obviously non of this would work out-of-the-box right now, so you might be right, apart from saving  Angular users some of their time, there's no real benefit.

bohoffi:
Ok, your argumentation is good.

If I can support you just let me know :)

eamador:
@sebastianhaas what's the state of this? Is there any development in progress where i can contribute?

sebastianhaas:
No, currently there isn't any development in progress regarding this feature. I think it would make proper sense to outsource this to another repository though.

I've created https://github.com/sebastianhaas/opensheetmusicdisplay-angular for this, and I would welcome PRs on the matter. :)

[...]

Issue #25, opened 2016-07-21
Create OSMD Wordpress plugin


acondolu:

sebastianhaas:
I can do this whenever @bukaznik wants me to (please assign milestone), I have some experience in creating WP plugins and php in general. It will take approximately one day including release and documentation.

bukaznik:
should work within page builders (eg for enfold and avada theme)

fredmeister77:
https://wordpress.org/plugins/opensheetmusicdisplay/ 😁

PR #361, opened 2018-08-30
Demo: Add Rerender button, add column to layout

sschmidTU:
Adds a rerender button to our demo, which is useful for debugging, see #360.
It behaves like the zoom button, just without the zoom.

I added a column to the layout, otherwise there would have been three rows.

I can still change the layout if there are objections.

image

Just by trying out the button, i discovered that the first rerender of a piece can effectively shrink or stretch the layout, which i didn't expect.


bneumann:
I would like to object. Why is that really needed on the demo page? What's different to console render() ?

sschmidTU:
The difference is that it's easier and more obvious for developers who don't know the console command, and window.osmd.render() doesn't work for me, as i wrote in #360.

The demo page is only for development anyways, this will probably not be included in our [official demo](https://opensheetmusicdisplay.org/demo/).

sschmidTU:
By the way, you can't open or resize the Chrome Developer Tools without triggering a resize and rerender, and they always take away from the main window's width.

bneumann:
Okay you're right. If you open the dev tools in an extra window it won't trigger anything even if you open them.

sschmidTU:
@bneumann by the way, how do you trigger a rerender in the console? window.osmd.render() doesn't work for me, just returns undefined, though without the () it returns the right function.

Still, i think a good debugging environment is important for development, and right now that's our demo. We could make a separate debugging environment and add overhaul our demo and everything, but for now i'd love to include this for debugging. Writing or copy-pasting something into the Chrome console every time is not very convenient.

bneumann:
I actually never tried just the render function. But the drawBoundingBoxes function also calls render. So I wonder why it wouldn't work.
Instead of copy you can also use arrow up. The history is saved over sessions

sebastianhaas:
Can't really comment on the issue, but a small suggestion: Maybe we want to change the button to re-render, redraw or refresh. Took me a while to actually get the rerender in the first place. I also did some online and offline research.

Three col layout is fine by me, Angular/WP etc. plugins would be highly appreciated. I hope I will find time to do it at some point. (#106, #25)

Sry and out. 🌷

bneumann:
Lgtm I'd say

sschmidTU:
Yes, i'm all for merging this. I thought there was still some dispute about the layout.
I would just change the button to say "re-render" instead of "rerender". "Rerender" is clear to me, but apparently it causes confusion.

sebastianhaas:
shall we publish a minor on npm, so that the demo gets updated on the website?

sschmidTU:
Just for the re-render button that's not necessary I think. It's more useful for debugging.
But we could make a new minor npm and Github release soon with some of the error handling and little other improvements

PR #367, opened 2018-09-04
Fix error loading empty score, improve error handling

### LINKED ISSUE #358: Error rendering empty score
[As supersational reported in #333](https://github.com/opensheetmusicdisplay/opensheetmusicdisplay/issues/333#issuecomment-416592023), loading an empty score causes an error.

StackTrace:
```stacktrace
Uncaught TypeError: Cannot read property 'Staves' of undefined
    at new GraphicalMusicSheet (GraphicalMusicSheet.ts:33)
    at OpenSheetMusicDisplay../src/OpenSheetMusicDisplay/OpenSheetMusicDisplay.ts.OpenSheetMusicDisplay.load (OpenSheetMusicDisplay.ts:126)
    at selectSampleOnChange (index.js:219)
    at FileReader.reader.onload (index.js:311)
```

this is because MusicSheetReader.createMusicSheet() returns undefined in OpenSheetMusicDisplay.ts:125

I'll look into it.

<details>
<summary>XML causing the error</summary>

```xml
<?xml version="1.0" encoding="utf-8"?>
<score-partwise version="3.0">
  <identification>
    <creator type="composer">Music21</creator>
    <encoding>
      <encoding-date>2018-08-22</encoding-date>
      <software>music21 v.5.3.0</software>
    </encoding>
  </identification>
  <defaults>
    <scaling>
      <millimeters>7</millimeters>
      <tenths>40</tenths>
    </scaling>
  </defaults>
  <part-list>
    <score-part id="P1">
      <part-name>Piano</part-name>
      <part-abbreviation>Pno</part-abbreviation>
      <score-instrument id="Ic248cbee2ba04bcf23dfa435ea191da3">
        <instrument-name>Piano</instrument-name>
        <instrument-abbreviation>Pno</instrument-abbreviation>
      </score-instrument>

    </score-part>
  </part-list>
  <part id="P1">
    <measure number="1">
      <attributes>
        <divisions>10080</divisions>
        <time>
          <beats>4</beats>
          <beat-type>4</beat-type>
        </time>
        <clef>
          <sign>G</sign>
          <line>2</line>
        </clef>
      </attributes>
    </measure>
  </part>
</score-partwise>
```

sschmidTU:
Can now render an empty score, Fixes #358 (see top comment there for XML sample)
image

Adds general improvements to error handling (undefined voices, parsing, calling getBoundingBox() on unformatted StaveNotes).

OSMD now returns an error if the MusicSheet read was undefined. Previously, this would cause random Errors and StackTraces in new GraphicalMusicSheet() (e.g. Staves undefined).

Also changed log.warn to log.info for a measure without voices in VexFlowMusicSheetCalculator.
log.warn produces warnings for the end user.

This still doesn't include XML verification, but one incomplete XML i tried simply gave another empty score, another returned an error "Error: given music sheet was incomplete or could not be loaded", as designed.
Discussion of validation in #358.

### DIFF
--- a/demo/index.js
+++ b/demo/index.js
@@ -141,7 +141,7 @@ import { OpenSheetMusicDisplay } from '../src/OpenSheetMusicDisplay/OpenSheetMus
                 try {
                     openSheetMusicDisplay.render();
                 } catch (e) {
-                    console.warn(e.stack);
+                    errorLoadingOrRenderingSheet(e, "rendering");
                 }
                 enable();
             }
@@ -225,18 +225,24 @@ import { OpenSheetMusicDisplay } from '../src/OpenSheetMusicDisplay/OpenSheetMus
                 return openSheetMusicDisplay.render();
             },
             function(e) {
-                console.warn(e.stack);
-                error("Error reading sheet: " + e);
+                errorLoadingOrRenderingSheet(e, "rendering");
             }
         ).then(
             function() {
                 return onLoadingEnd(isCustom);
             }, function(e) {
-                console.warn("Error rendering sheet: " + e + " StackTrace: \n" + e.stack);
-                error("Error rendering sheet: " + process.env.DEBUG ? e.stack : e);
+                errorLoadingOrRenderingSheet(e, "loading");
                 onLoadingEnd(isCustom);
             }
         );
+    }
+
+    function errorLoadingOrRenderingSheet(e, loadingOrRenderingString) {
+        var errorString = "Error " + loadingOrRenderingString + " sheet: " + e;
+        if (process.env.DEBUG) {
+            errorString += "\n" + "StackTrace: \n" + e.stack;
+        }
+        console.warn(errorString);
     }
 
     function onLoadingEnd(isCustom) {
--- a/demo/index.js
+++ b/demo/index.js
@@ -239,9 +239,9 @@ import { OpenSheetMusicDisplay } from '../src/OpenSheetMusicDisplay/OpenSheetMus
 
     function errorLoadingOrRenderingSheet(e, loadingOrRenderingString) {
         var errorString = "Error " + loadingOrRenderingString + " sheet: " + e;
-        if (process.env.DEBUG) {
-            errorString += "\n" + "StackTrace: \n" + e.stack;
-        }
+        // if (process.env.DEBUG) { // people may not set an environment variable for debug.
+        errorString += "\n" + "StackTrace: \n" + e.stack;
+        // }
         console.warn(errorString);
     }
 
--- a/demo/index.js
+++ b/demo/index.js
@@ -239,7 +239,9 @@ import { OpenSheetMusicDisplay } from '../src/OpenSheetMusicDisplay/OpenSheetMus
 
     function errorLoadingOrRenderingSheet(e, loadingOrRenderingString) {
         var errorString = "Error " + loadingOrRenderingString + " sheet: " + e;
-        // if (process.env.DEBUG) { // people may not set an environment variable for debug.
+        // if (process.env.DEBUG) { // people may not set a debug environment variable for the demo.
+        // Always giving a StackTrace might give us more and better error reports.
+        // TODO for a release, StackTrace control could be reenabled
         errorString += "\n" + "StackTrace: \n" + e.stack;
         // }
         console.warn(errorString);
--- a/demo/index.js
+++ b/demo/index.js
@@ -113,7 +113,7 @@ import { OpenSheetMusicDisplay } from '../src/OpenSheetMusicDisplay/OpenSheetMus
             openSheetMusicDisplay.DrawSkyLine = !openSheetMusicDisplay.DrawSkyLine;
         }
 
-        bottomlineDebug .onclick = function() {
+        bottomlineDebug.onclick = function() {
             openSheetMusicDisplay.DrawBottomLine = !openSheetMusicDisplay.DrawBottomLine;
         }
 
@@ -232,7 +232,7 @@ import { OpenSheetMusicDisplay } from '../src/OpenSheetMusicDisplay/OpenSheetMus
             function() {
                 return onLoadingEnd(isCustom);
             }, function(e) {
-                console.warn(e.stack);
+                console.warn("Error rendering sheet: " + e + " StackTrace: \n" + e.stack);
                 error("Error rendering sheet: " + process.env.DEBUG ? e.stack : e);
                 onLoadingEnd(isCustom);
             }
--- a/src/MusicalScore/Graphical/VexFlow/VexFlowMusicSheetCalculator.ts
+++ b/src/MusicalScore/Graphical/VexFlow/VexFlowMusicSheetCalculator.ts
@@ -118,7 +118,7 @@ export class VexFlowMusicSheetCalculator extends MusicSheetCalculator {
             }
         }
         if (voices.length === 0) {
-            log.warn("Found a measure with no voices... Continuing anyway.", mvoices);
+            log.info("Found a measure with no voices. Continuing anyway.", mvoices);
             continue;
         }
         // all voices that belong to one stave are collectively added to create a common context in VexFlow.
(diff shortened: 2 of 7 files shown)


demo/index.js
```diff
@@ -232,7 +232,7 @@ import { OpenSheetMusicDisplay } from '../src/OpenSheetMusicDisplay/OpenSheetMus
             function() {
                 return onLoadingEnd(isCustom);
             }, function(e) {
-                console.warn(e.stack);
+                console.warn("Error rendering sheet: " + e + " StackTrace: \n" + e.stack);
```
sschmidTU: e.stack can be undefined and e.toString can contain important information (UnformattedNoteError from VexFlow when calling getBoundingBox())

src/MusicalScore/Graphical/VexFlow/VexFlowMusicSheetCalculator.ts
```diff
@@ -118,7 +118,7 @@ export class VexFlowMusicSheetCalculator extends MusicSheetCalculator {
             }
         }
         if (voices.length === 0) {
-            log.warn("Found a measure with no voices... Continuing anyway.", mvoices);
```
sschmidTU: log.warn gets printed to console for the end user (standard loglevel setting), and i don't think that's a good idea in this place, e.g. for an empty score.

sschmidTU:
LGTM now. Even if we want to do XML verification, this has some reasonable error handling improvements.

demo/index.js
```diff
@@ -225,20 +225,26 @@ import { OpenSheetMusicDisplay } from '../src/OpenSheetMusicDisplay/OpenSheetMus
...
-                error("Error rendering sheet: " + process.env.DEBUG ? e.stack : e);
+                errorLoadingOrRenderingSheet(e, "loading");
                 onLoadingEnd(isCustom);
             }
         );
     }
 
+    function errorLoadingOrRenderingSheet(e, loadingOrRenderingString) {
+        var errorString = "Error " + loadingOrRenderingString + " sheet: " + e;
+        if (process.env.DEBUG) {
```
bneumann: I think nobody will ever do this. But yeah just leave it there for ppl to lazy to pop open the console 👯‍♂️

src/MusicalScore/Graphical/VexFlow/VexFlowMusicSheetCalculator.ts
bneumann: Yep you're righto

src/MusicalScore/Graphical/VexFlow/VexFlowStaffEntry.ts
```diff
@@ -27,8 +27,17 @@ export class VexFlowStaffEntry extends GraphicalStaffEntry {
         for (const gve of this.graphicalVoiceEntries as VexFlowVoiceEntry[]) {
             if (gve.vfStaveNote) {
                 gve.vfStaveNote.setStave(stave);
-                gve.applyBordersFromVexflow();
-                this.PositionAndShape.RelativePosition.x = gve.vfStaveNote.getBoundingBox().x / unitInPixels;
+                try {
+                    gve.applyBordersFromVexflow();
+                    this.PositionAndShape.RelativePosition.x = gve.vfStaveNote.getBoundingBox().x / unitInPixels;
```
bneumann: I really really really don't like try catch if not necessary. And I think you can acutally get the isformatted information from vexflow (stavenote.js):
```javascript
getBoundingBox() {
if (!this.preFormatted) {
   throw new Vex.RERR('UnformattedNote', "Can't call getBoundingBox on an unformatted note.");
 }
```

src/MusicalScore/ScoreIO/MusicSheetReader.ts
```diff
@@ -486,21 +487,20 @@ export class MusicSheetReader /*implements IMusicSheetReader*/ {
     private pushSheetLabels(root: IXmlElement, filePath: string): void {
         this.readComposer(root);
         this.readTitle(root);
-        if (this.musicSheet.Title === undefined || this.musicSheet.Composer === undefined) {
-            this.readTitleAndComposerFromCredits(root);
-        }
-        if (this.musicSheet.Title === undefined) {
-            try {
+        try {
```
bneumann: Again this can't be the proper solution. I would check ```readTitleAndComposerFromCredits``` for the error first

src/MusicalScore/ScoreIO/MusicSheetReader.ts
```diff
@@ -611,8 +611,13 @@ export class MusicSheetReader /*implements IMusicSheetReader*/ {
         }
         let paperHeight: number = 0;
         let topSystemDistance: number = 0;
-        const defi: string = root.element("defaults").element("page-layout").element("page-height").value;
-        paperHeight = parseFloat(defi);
+        try {
+            const defi: string = root.element("defaults").element("page-layout").element("page-height").value;
```
bneumann: Again: check for if element(xyz) === undefined instead of try catch. It gets a longish expression but not as costly as try catch

demo/index.js
sschmidTU: People will never do what? These errors were already printed just like that, only the code was duplicated 4 times. I just made a common error logging method.

src/MusicalScore/Graphical/VexFlow/VexFlowStaffEntry.ts
sschmidTU: ok, i'll see if i can get the information from VexFlow. But maybe sometimes you can't avoid VexFlow throwing errors.

src/MusicalScore/ScoreIO/MusicSheetReader.ts
sschmidTU: that alternative is checking for undefined 4 times, probably in 4 nested if-checks. we can do that, not sure if it's better or prettier.

src/MusicalScore/ScoreIO/MusicSheetReader.ts
sschmidTU: Why? You want two try catch blocks, just to know which try block the error came from?
If you want to know that, you can look at the error message or debug.
I don't really get what you mean, `readTitleAndComposerFromCredits` is the first thing i do in the try block anyways.

In any case, all of this works and gives proper error messages, so i don't see the problem.

src/MusicalScore/ScoreIO/MusicSheetReader.ts
matt-uib: I would also say: keep it simple now and don't optimize things too early. Let's see if your current changes do the job for us and improve it if we see problems.

src/MusicalScore/Graphical/VexFlow/VexFlowStaffEntry.ts
sschmidTU: done (b54eb2588c5a0c178caf54288b975065671c2359). this adds a dependency on how VexFlow is implemented, but it looks cleaner.

src/MusicalScore/ScoreIO/MusicSheetReader.ts
sschmidTU: I understand now, fixed in 5dc8c80a3137db1e2e0abd4325ad1cd7260efd7f.
If `readTitleAndComposerFromCredits` fails, the second method of reading the title (from filename) wouldn't be executed with just one try block.

src/MusicalScore/ScoreIO/MusicSheetReader.ts
sschmidTU: I'm not even sure the if-checks would perform better. We'd have to repeat each root.element call in each nested if loop (1), or store each in a new variable (2), and then the code really gets ugly.
Also, the try block is only performed once for each score.

method 1:
```javascript
if (root.element("defaults")) {
  if (root.element("defaults").element("page-layout")) {
    // ...
```

method 2:
```javascript
const elementDefaults: IXmlElement = root.element("defaults");
if (elementDefaults !== undefined) {
  const elementPageLayout: IXmlElement = elementDefaults.element("page-layout");
  if (elementPageLayout !== undefined {
    // ...
```

sschmidTU: In Kotlin, this would be easy: root?.elements("defaults")?... which returns null if any of the calls are null.

demo/index.js
sschmidTU: well, i didn't write that part of the error handling. I'd be fine with removing that part and always printing the stack trace.
Of course then it will also be printed for anyone running the local demo and even our public demo if we don't check the DEBUG env variable, i guess that's what it's for.

PR #368, opened 2018-09-04
feat: options interface for osmd constructor

### LINKED ISSUE #288: Global config class
There should be a global class (singleton?) that can be configured on startup for things like:
- debugging (skyline, bounding boxes, ...)
- Autobeam
- Autostem
...

Remakr: Reuse .\MusicalScore\Graphical\DrawingParameters.ts for this

sschmidTU:
Resolves #288

Adds an interface for options that replaces the individual options parameters in the OpenSheetMusicDisplay() constructor, as @sebastianhaas suggested in #288.
Any changes you and @bneumann want to see?

I added the interface to OpenSheetMusicDisplay.ts and exported it, so i hope people using OSMD with TypeScript can import the interface as well.
But this may not be the best way to go about it, maybe it should go into a new file.

```javascript
export interface IOSMDOptions {
    autoBeam?: boolean; // not yet supported. will always autoBeam.
    autoResize?: boolean; // default is false
    autoStem?: boolean; // not yet supported. will always autoStem
    backend?: string; // Backend will be SVG if backend === "svg", otherwise Canvas
    disableCursor?: boolean; // will override this part of drawingParameters
    drawingParametersEnum?: DrawingParametersEnum;
    drawingParametersString?: string; // alternative to using the enum. only need to set one of these.
}
```

Right now we don't have too many working options. Not sure if we could implement disabling autoStem or autoBeam, for example.

In any case, this works for now, and enables disabling the cursor, which @p-himik requested in #5.

### DIFF
--- a/demo/index.js
+++ b/demo/index.js
@@ -121,7 +121,7 @@ import { OpenSheetMusicDisplay } from '../src/OpenSheetMusicDisplay/OpenSheetMus
         openSheetMusicDisplay = new OpenSheetMusicDisplay(canvas, {
             autoResize: false,
             backend: backendSelect.value,
-            drawingParametersString: "default",
+            drawingParameters: "default",
             disableCursor: false,
         });
         openSheetMusicDisplay.setLogLevel('info');
--- a/demo/index.js
+++ b/demo/index.js
@@ -113,12 +113,17 @@ import { OpenSheetMusicDisplay } from '../src/OpenSheetMusicDisplay/OpenSheetMus
             openSheetMusicDisplay.DrawSkyLine = !openSheetMusicDisplay.DrawSkyLine;
         }
 
-        bottomlineDebug .onclick = function() {
+        bottomlineDebug.onclick = function() {
             openSheetMusicDisplay.DrawBottomLine = !openSheetMusicDisplay.DrawBottomLine;
         }
 
         // Create OSMD object and canvas
-        openSheetMusicDisplay = new OpenSheetMusicDisplay(canvas, false, backendSelect.value);
+        openSheetMusicDisplay = new OpenSheetMusicDisplay(canvas, {
+            autoResize: false,
+            backend: backendSelect.value,
+            drawingParametersString: "default",
+            disableCursor: false,
+        });
         openSheetMusicDisplay.setLogLevel('info');
         document.body.appendChild(canvas);
 
@@ -163,7 +168,11 @@ import { OpenSheetMusicDisplay } from '../src/OpenSheetMusicDisplay/OpenSheetMus
             openSheetMusicDisplay.cursor.hide();
         });
         showCursorBtn.addEventListener("click", function() {
-            openSheetMusicDisplay.cursor.show();
+            if (openSheetMusicDisplay.cursor) {
+                openSheetMusicDisplay.cursor.show();
+            } else {
+                console.info("can't show cursor, cursor was disabled by drawingParameters!");
+            }
         });
 
         backendSelect.addEventListener("change", function(e) {
--- a/demo/index.js
+++ b/demo/index.js
@@ -119,7 +119,7 @@ import { OpenSheetMusicDisplay } from '../src/OpenSheetMusicDisplay/OpenSheetMus
 
         // Create OSMD object and canvas
         openSheetMusicDisplay = new OpenSheetMusicDisplay(canvas, {
-            autoResize: false,
+            autoResize: true,
             backend: backendSelect.value,
             drawingParameters: "default",
             disableCursor: false,
@@ -187,7 +187,6 @@ import { OpenSheetMusicDisplay } from '../src/OpenSheetMusicDisplay/OpenSheetMus
     }
 
     function Resize(startCallback, endCallback) {
-
       var rtime;
       var timeout = false;
       var delta = 200;
--- a/demo/index.js
+++ b/demo/index.js
@@ -171,7 +171,7 @@ import { OpenSheetMusicDisplay } from '../src/OpenSheetMusicDisplay/OpenSheetMus
             if (openSheetMusicDisplay.cursor) {
                 openSheetMusicDisplay.cursor.show();
             } else {
-                console.info("can't show cursor, cursor was disabled by drawingParameters!");
+                console.info("Can't show cursor, as it was disabled (e.g. by drawingParameters).");
             }
         });
 
--- a/karma.conf.js
+++ b/karma.conf.js
@@ -81,7 +81,7 @@ module.exports = function (config) {
         client: {
             captureConsole: true,
             mocha: {
-                timeout: process.env.timeout || 2000
+                timeout: process.env.timeout || 6000
             }
         },
 
--- a/src/MusicalScore/Graphical/DrawingParameters.ts
+++ b/src/MusicalScore/Graphical/DrawingParameters.ts
@@ -38,16 +38,14 @@ export class DrawingParameters {
     public set DrawingParametersEnum(drawingParametersEnum: DrawingParametersEnum) {
         this.drawingParametersEnum = drawingParametersEnum;
         switch (drawingParametersEnum) {
-            case DrawingParametersEnum.Default:
-            case DrawingParametersEnum.AllOn:
-                this.setForAllOn();
-                break;
             case DrawingParametersEnum.Thumbnail:
-                this.setForThumbnail();
-                break;
+            this.setForThumbnail();
+            break;
             case DrawingParametersEnum.Leadsheet:
-                this.setForLeadsheet();
-                break;
+            this.setForLeadsheet();
+            break;
+            case DrawingParametersEnum.AllOn:
+            case DrawingParametersEnum.Default:
             default:
                 this.setForAllOn();
         }
--- a/src/MusicalScore/Graphical/DrawingParameters.ts
+++ b/src/MusicalScore/Graphical/DrawingParameters.ts
@@ -1,4 +1,13 @@
+export enum DrawingParametersEnum {
+    AllOn,
+    Default, // default is AllOn for now
+    Leadsheet,
+    Preview,
+    Thumbnail,
+}
+
 export class DrawingParameters {
+    private drawingParametersEnum: DrawingParametersEnum; // will control other settings if changed with set method
     public drawHighlights: boolean;
     public drawErrors: boolean;
     public drawSelectionStartSymbol: boolean;
@@ -8,6 +17,45 @@ export class DrawingParameters {
     public drawScrollIndicator: boolean;
     public drawComments: boolean;
     public drawMarkedAreas: boolean;
+
+    public static DrawingParametersStringToEnum(stringParameter: string): DrawingParametersEnum {
+        switch (stringParameter.toLowerCase()) {
+            case "allOn":
+                return DrawingParametersEnum.AllOn;
+            case "default":
+                return DrawingParametersEnum.Default;
+            case "leadsheet":
+                return DrawingParametersEnum.Leadsheet;
+            case "preview":
+                return DrawingParametersEnum.Preview;
+            case "thumbnail":
+                return DrawingParametersEnum.Thumbnail;
+            default:
+                return DrawingParametersEnum.Default;
+        }
+    }
+
+    public set DrawingParametersEnum(drawingParametersEnum: DrawingParametersEnum) {
+        this.drawingParametersEnum = drawingParametersEnum;
+        switch (drawingParametersEnum) {
+            case DrawingParametersEnum.Default:
+            case DrawingParametersEnum.AllOn:
+                this.setForAllOn();
+                break;
+            case DrawingParametersEnum.Thumbnail:
+                this.setForThumbnail();
+                break;
+            case DrawingParametersEnum.Leadsheet:
+                this.setForLeadsheet();
+                break;
+            default:
+                this.setForAllOn();
+        }
+    }
+
+    public get DrawingParametersEnum(): DrawingParametersEnum {
+        return this.drawingParametersEnum;
+    }
 
     public setForAllOn(): void {
         this.drawHighlights = true;
@@ -21,7 +69,7 @@ export class DrawingParameters {
         this.drawMarkedAreas = true;
     }
 
-    public setForThumbmail(): void {
+    public setForThumbnail(): void {
         this.drawHighlights = false;
         this.drawErrors = false;
         this.drawSelectionStartSymbol = false;
--- a/src/MusicalScore/Graphical/DrawingParameters.ts
+++ b/src/MusicalScore/Graphical/DrawingParameters.ts
@@ -24,6 +24,8 @@ export class DrawingParameters {
     public drawPartName: boolean = true;
     /** Draw notes set to be invisible (print-object="no" in XML). */
     public drawHiddenNotes: boolean = false;
+    public defaultColorNoteHead: string; // TODO not yet supported
+    public defaultColorStem: string; // TODO not yet supported
 
     constructor(drawingParameters: DrawingParametersEnum = DrawingParametersEnum.Default) {
         this.DrawingParametersEnum = drawingParameters;
--- a/src/MusicalScore/Graphical/DrawingParameters.ts
+++ b/src/MusicalScore/Graphical/DrawingParameters.ts
@@ -19,6 +19,10 @@ export class DrawingParameters {
     public drawComments: boolean;
     public drawMarkedAreas: boolean;
 
+    constructor(drawingParameters: DrawingParametersEnum = DrawingParametersEnum.Default) {
+        this.DrawingParametersEnum = drawingParameters;
+    }
+
     public set DrawingParametersEnum(drawingParametersEnum: DrawingParametersEnum) {
         this.drawingParametersEnum = drawingParametersEnum;
         switch (drawingParametersEnum) {
--- a/src/MusicalScore/Graphical/DrawingParameters.ts
+++ b/src/MusicalScore/Graphical/DrawingParameters.ts
@@ -1,5 +1,6 @@
 export enum DrawingParametersEnum {
     AllOn = "allon",
+    Compact = "compact",
     Default = "default", // default is AllOn for now
     Leadsheet = "leadsheet",
     Preview = "preview",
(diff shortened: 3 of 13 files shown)


bneumann:
Note default color, stem default color. I have some more, will give you feedback later today. No pc

bneumann:
All tests failing? Not a good sign :D

src/OpenSheetMusicDisplay/OpenSheetMusicDisplay.ts
```diff
@@ -13,14 +13,20 @@ import {MXLHelper} from "../Common/FileIO/Mxl";
...
 
 export class OpenSheetMusicDisplay {
     /**
      * The easy way of displaying a MusicXML sheet music file
      * @param container is either the ID, or the actual "div" element which will host the music sheet
      * @autoResize automatically resize the sheet to full page width on window resize
      */
-    constructor(container: string|HTMLElement, autoResize: boolean = false, backend: string = "svg") {
+    constructor(container: string|HTMLElement,
+                options: OSMDOptions = {
```
bneumann: Maybe put the creation of the defaults in a static function. There will be tons more options I think and the constructor will get overblown.

src/OpenSheetMusicDisplay/OpenSheetMusicDisplay.ts
```diff
@@ -13,14 +13,20 @@ import {MXLHelper} from "../Common/FileIO/Mxl";
...
+                    autoResize: false,
```
bneumann: I'd leave autoResize to true

src/OpenSheetMusicDisplay/OpenSheetMusicDisplay.ts
sschmidTU: yes, definitely should be somewhere else.
doesn't need to be static though, could also be a private function called in the constructor.

src/OpenSheetMusicDisplay/OpenSheetMusicDisplay.ts
sschmidTU: it was false before, but we can talk about it ;)

bneumann: I know, but I think that was a mistake.

src/OpenSheetMusicDisplay/OpenSheetMusicDisplay.ts
bneumann: Not sure about TS but in other languages accessing a private method before the constructor doesn't work

sschmidTU:
about tests failing, Travis CI passes with Node 8 and fails with Node 6, as usual.
But there's a problem with Karma loading files:
```stacktrace
ERROR [karma]: TypeError: Path must be a string. Received undefined
    at assertPath (path.js:7:11)
    at Object.join (path.js:470:7)
    at Plugin.<anonymous> (C:\projects\opensheetmusicdisplay\node_modules\karma-webpack\lib\karma-webpack.js:262:68)
    at Plugin.readFile (C:\projects\opensheetmusicdisplay\node_modules\karma-webpack\lib\karma-webpack.js:281:5)
    at _combinedTickCallback (internal/process/next_tick.js:73:7)
    at process._tickCallback (internal/process/next_tick.js:104:9)
```
I'll look into it.

src/OpenSheetMusicDisplay/OpenSheetMusicDisplay.ts
sschmidTU: It doesn't need to be before the constructor, i'd call it in the constructor.

src/OpenSheetMusicDisplay/OpenSheetMusicDisplay.ts
```diff
@@ -307,3 +331,13 @@ export class OpenSheetMusicDisplay {
...
     //#endregion
 }
+
+export interface OSMDOptions {
+    autoBeam?: boolean; // not yet supported. will always autoBeam.
+    autoResize?: boolean; // default is false
+    autoStem?: boolean; // not yet supported. will always autoStem
+    backend?: string; // Backend will be SVG if backend === "svg", otherwise Canvas
+    disableCursor?: boolean; // will override this part of drawingParameters
+    drawingParametersEnum?: DrawingParametersEnum;
```
sebastianhaas: I don't really like this, what's the reason not to use something like:

```
drawingParameters?: DrawingParametersEnum | string;
```

src/OpenSheetMusicDisplay/OpenSheetMusicDisplay.ts
```diff
@@ -307,3 +331,13 @@ export class OpenSheetMusicDisplay {
     }
     //#endregion
 }
+
+export interface OSMDOptions {
```
sebastianhaas: We need proper docstrings for Typedoc here. On the interface itself as well as on the options. The comments as written now won't make it into the Typedoc output.

test/MusicalScore/Graphical/VexFlow/VexFlowMusicSheetDrawer_Test.ts
```diff
@@ -9,6 +9,7 @@ import {IXmlElement} from "../../../../src/Common/FileIO/Xml";
 import {Fraction} from "../../../../src/Common/DataObjects/Fraction";
 import {VexFlowBackend} from "../../../../src/MusicalScore/Graphical/VexFlow/VexFlowBackend";
 import {CanvasVexFlowBackend} from "../../../../src/MusicalScore/Graphical/VexFlow/CanvasVexFlowBackend";
+import { DrawingParameters } from "../../../../src/MusicalScore/Graphical/DrawingParameters";
```
sebastianhaas: Maybe no spaces here. I like it better with the spaces but we should be consistent I guess.

src/MusicalScore/Graphical/DrawingParameters.ts
```diff
@@ -9,6 +18,45 @@ export class DrawingParameters {
     public drawComments: boolean;
     public drawMarkedAreas: boolean;
 
+    public static DrawingParametersStringToEnum(stringParameter: string): DrawingParametersEnum {
```
sebastianhaas: Don't know if we need this. Maybe it is possible to use string based enums here? Those can be used from JS as well. Introduced in 2.4: https://www.typescriptlang.org/docs/handbook/release-notes/typescript-2-4.html

src/MusicalScore/Graphical/DrawingParameters.ts
```diff
@@ -9,6 +18,45 @@ export class DrawingParameters {
...
+                return DrawingParametersEnum.Thumbnail;
+            default:
+                return DrawingParametersEnum.Default;
+        }
+    }
+
+    public set DrawingParametersEnum(drawingParametersEnum: DrawingParametersEnum) {
+        this.drawingParametersEnum = drawingParametersEnum;
+        switch (drawingParametersEnum) {
+            case DrawingParametersEnum.Default:
```
sebastianhaas: maybe move this down and group with `default` :)

src/MusicalScore/Graphical/VexFlow/VexFlowMusicSheetDrawer.ts
```diff
@@ -23,6 +23,7 @@ import { PlacementEnum } from "../../VoiceData/Expressions/AbstractExpression";
 import {GraphicalInstantaneousTempoExpression} from "../GraphicalInstantaneousTempoExpression";
 import {GraphicalInstantaneousDynamicExpression} from "../GraphicalInstantaneousDynamicExpression";
 import log = require("loglevel");
+import { DrawingParameters } from "../DrawingParameters";
```
sebastianhaas: see comment above

src/OpenSheetMusicDisplay/OpenSheetMusicDisplay.ts
```diff
@@ -13,14 +13,20 @@ import {MXLHelper} from "../Common/FileIO/Mxl";
 import {Promise} from "es6-promise";
 import {AJAX} from "./AJAX";
 import * as log from "loglevel";
+import { DrawingParametersEnum, DrawingParameters } from "../MusicalScore/Graphical/DrawingParameters";
```
sebastianhaas: spaces, see above

demo/index.js
```diff
@@ -163,7 +168,11 @@ import { OpenSheetMusicDisplay } from '../src/OpenSheetMusicDisplay/OpenSheetMus
             openSheetMusicDisplay.cursor.hide();
         });
         showCursorBtn.addEventListener("click", function() {
-            openSheetMusicDisplay.cursor.show();
+            if (openSheetMusicDisplay.cursor) {
+                openSheetMusicDisplay.cursor.show();
+            } else {
+                console.info("can't show cursor, cursor was disabled by drawingParameters!");
```
sebastianhaas: Ok, this is suuuper picky now, but maybe we want the message to start with a capital letter if it is ended as a sentence 🙈

src/OpenSheetMusicDisplay/OpenSheetMusicDisplay.ts
bneumann: Second that

sschmidTU:
Working on this again. Thinking about putting the interface in an OSMDOptions.ts file.

Talking with @matt-uib, we also thought it would be useful to be able to set some options during runtime. So, we can still have these constructor options, but maybe they should call functions like disableCursor(), which are also callable when OSMD is already running.

Will also look at all your other suggestions.

demo/index.js
sschmidTU: As a fellow word nerd, I agree ;)

src/OpenSheetMusicDisplay/OpenSheetMusicDisplay.ts
sschmidTU: Yes, autoResize true is probably better default, changed in next commit.

src/OpenSheetMusicDisplay/OpenSheetMusicDisplay.ts
sschmidTU: It's because we need to handle a string differently than an Enum,
and [type checks in Typescript type still aren't ideal](https://stackoverflow.com/questions/12789231/class-type-check-with-typescript/12789324#12789324).
But since we use `instanceof` a lot in OSMD, i guess we can do it here, too.

sschmidTU: Also, it seems like instanceof doesn't work with Enums, or something.

image

<details>
<summary>Details</summary>

```typescript
[ts] The left-hand side of an 'instanceof' expression must be of type 'any', an object type or a type parameter.
(property) IOSMDOptions.drawingParameters?: string | DrawingParametersEnum.Default | DrawingParametersEnum.Leadsheet | DrawingParametersEnum.Preview | DrawingParametersEnum.Thumbnail
```

```typescript
        this.drawingParameters = new DrawingParameters();
        if (options.drawingParameters) {
            if (options.drawingParameters instanceof DrawingParametersEnum) {
                this.drawingParameters.DrawingParametersEnum = options.drawingParametersEnum;
            } else if (options.drawingParameters instanceof string) {
                this.drawingParameters.DrawingParametersEnum =
                    DrawingParameters.DrawingParametersStringToEnum(options.drawingParametersString);
            }
        } else {
            this.drawingParameters.DrawingParametersEnum = DrawingParametersEnum.Default;
        }
```

</details>

src/OpenSheetMusicDisplay/OpenSheetMusicDisplay.ts
sschmidTU: done in next commit

src/MusicalScore/Graphical/DrawingParameters.ts
sschmidTU: We can use string based enums, but not if we port it back to C#.
I like enums as numbers (ideally ints), strings are messy.

sschmidTU:
For some reason, npm test (karma) runs only 79 tests in this branch instead of 155. Not sure why.
The only tests i changed were in VexFlowMusicSheetDrawer_Test.

sschmidTU:
Ok, half the tests not running was because autoResize is true now as a default. Which is interesting in itself. But i'll change the tests to set autoResize to false, i'll make a factory method for the OSMD constructor.

bneumann:
I can only guess the the tests check width and height and the resize option would prevent getting repeatable results

sschmidTU:
I found what's causing the error for the load MXL from string test, but i don't know how to fix it.

In TestUtils.ts, window.__raw__ is undefined.
```javascript
    public static getMXL(scoreName: string): string {
        const path: string = "test/data/" + scoreName;
        return ((window as any).__raw__)[path];
    }
```
I don't even know how the function in TestUtils works, i didn't find anything searching for window and __raw__.

This works in develop, before these constructor changes, but not in the PR.
But all the options for the tests should be the same, so i don't know why this doesn't work like in the develop branch. I must have changed something that affects the tests. I also had to increase the Karma timeout for a few other MXL/XML tests.

You can debug this in the fix/osmd-options-mxl-test-error branch. When you click Zoom In in the demo, it will load this MXL test. (Debugging Karma was a pain for me)

karma.conf.js
```diff
@@ -81,7 +81,7 @@ module.exports = function (config) {
         client: {
             captureConsole: true,
             mocha: {
-                timeout: process.env.timeout || 2000
+                timeout: process.env.timeout || 6000
```
bneumann: ENV timout is 10000 by the way

src/OpenSheetMusicDisplay/OpenSheetMusicDisplay.ts
```diff
@@ -36,7 +36,7 @@ export class OpenSheetMusicDisplay {
             throw new Error("Please pass a valid div container to OpenSheetMusicDisplay");
         }
 
-        if (options.backend.toLowerCase() === "svg" || options.backend === undefined) {
```
bneumann: Haha how could this even have worked? :D

test/Common/OSMD/OSMD_Test.ts
```diff
@@ -2,6 +2,14 @@ import chai = require("chai");
 import {OpenSheetMusicDisplay} from "../../../src/OpenSheetMusicDisplay/OpenSheetMusicDisplay";
 import {TestUtils} from "../../Util/TestUtils";
 
+function createOpenSheetMusicDisplay(div: HTMLElement): OpenSheetMusicDisplay {
```
bneumann: Can we move this to TestUtils?

bneumann:
Some searching brought me back to *drumroll* us. https://github.com/opensheetmusicdisplay/karma-base64-to-js-preprocessor 

So this is basically a loader for mxl.base64 files which we don't even have.

sschmidTU:
Interesting. But it works in the develop branch.

bneumann:
Yeah I was wrong there is exactly one file. And it gets loaded by the testutils function alright. But somehow the error message still comes.

karma.conf.js
sschmidTU: strange, 3 tests or so were failing for me until i changed this, but i guess that wasn't the cause.

test/Common/OSMD/OSMD_Test.ts
sschmidTU: yes, good idea. can also be used in Xml_Test.

src/OpenSheetMusicDisplay/OpenSheetMusicDisplay.ts
sschmidTU: just a whoopsie daisy ;)

src/MusicalScore/Graphical/DrawingParameters.ts
sebastianhaas: https://goo.gl/HUdLTh

What about this? It doesn't seem too messy to me, and you can use it directly from JS with the string value.

src/OpenSheetMusicDisplay/OpenSheetMusicDisplay.ts
sebastianhaas: You don't need to do the typecheck if you use string-based enums.
https://goo.gl/HUdLTh

karma.conf.js
sschmidTU: you meant ENV timeout is 10000 on our CI, right?
Because on my machine, it uses the second value, and 3 tests keep failing.
So, i'd like to keep this at 6000, so that tests don't fail on my PC.

sschmidTU:
I fixed the test fail, so the basics of this PR are solid.
Now we can consider integrating other options (Note default color, Stem default color) before merging.

@bneumann you said you might know more OSMD options we could integrate?

Also, these settings should be changeable after the constructor, when OSMD is already running, for example the note color or disabling the cursor, so there should be methods for changing these settings called in the constructor.

I'm a bit impatient to have this merged, e.g. #367 depends on this (add debug option), but i guess it's less work to do it right the first time.

The cause of the test(s) failing was autoResize(), but for this I created another PR #387, because it's complicated. The resize refactors will not be trivial, so I hope we can mostly ignore the changes to resize in this PR, they simply circumvent errors.

src/OpenSheetMusicDisplay/OSMDOptions.ts
```diff
@@ -0,0 +1,29 @@
...
+    /** Don't show/load cursor. Will override disableCursor in drawingParameters. */
+    disableCursor?: boolean;
+    /** Parameters like drawing a Leadsheet or (Thumbnail) Preview, disabling Cursor. */
+    drawingParameters?: string | DrawingParametersEnum;
+}
+
+/** Handles [[IOSMDOptions]], e.g. returning default options with OSMDOptionsStandard() */
+export class OSMDOptions {
+    /** Returns the default options for OSMD used if no options are given in the constructor. */
+    public static OSMDOptionsStandard(): IOSMDOptions {
```
sebastianhaas: Can you maybe use @return here so that docs are generated correctly? :)
http://typedoc.org/guides/doccomments/

sschmidTU: Can do, although e.g. `RegExp.test()` also doesn't have an @return notation.
Are you sure you need @return to get correctly generated docs?

sebastianhaas: Actually I don't.

sebastianhaas: grafik
At least you get a clickable link with a correct section heading which otherwise you don't.

sschmidTU: Right. Maybe one could do a `[[Link]]` instead. I'll do one or the other.
Can you quickly tell me where to find the docs built with `build:doc`? Can't find them right now.

sschmidTU: It seems like it gives you a clickable return type without the @return, which is good to know.
image

sschmidTU: The reason i don't like the @return is that i would need to repeat myself from the description, so if i can describe function and return value in one doc sentence, i like that better.

PR #382, opened 2018-09-11
Fix RepetitionReader RegExes, RepetitionCalculator crash

### LINKED ISSUE #370: Incorrect repetition detection leads to crashes
For example, `RepetitionInstructionReader` uses `StringUtil.StringContainsSeparatedWord(innerText, "d.s.")` which in turn uses `String.prototype.search()` which accepts a RegEx.
Hence, the above check for "d.s." passes also for such text as "dust".

In the end, it crashes at
```typescript
case RepetitionInstructionEnum.DalSegno:
	this.currentMeasure.LastRepetitionInstructions.push(currentRepetitionInstruction);
```
at `RepetitionCalculator.handleRepetitionInstructions` because `this.currentMeasure ` was not set.

sschmidTU:
Fixes #370

Our StringUtil.StringContainsSeparatedWord function was implemented as if it used plain strings instead of RegExes with String.search(). The function should work as designed now, with proper RegExes.

Also fixes a crash in RepetitionCalculator when currentMeasure was not set.

### DIFF
--- a/src/Common/Strings/StringUtil.ts
+++ b/src/Common/Strings/StringUtil.ts
@@ -1,9 +1,9 @@
 ﻿export class StringUtil {
-  public static StringContainsSeparatedWord(str: string, word: string): boolean {
-    if (str === word ||
-      str.search(" " + word) !== -1 ||
-      str.search(word + " ") !== -1 ||
-      str.search(word + ".") !== -1) {
+  public static StringContainsSeparatedWord(str: string, wordRegExString: string): boolean {
+    if (new RegExp("^" + wordRegExString + "$").test(str) ||
+      new RegExp(" " + wordRegExString).test(str) ||
+      new RegExp(wordRegExString + " ").test(str) ||
+      new RegExp(wordRegExString + "\\.").test(str)) {
       return true;
     }
     return false;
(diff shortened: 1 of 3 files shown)


src/MusicalScore/ScoreIO/MusicSymbolModules/RepetitionCalculator.ts
```diff
@@ -42,6 +42,9 @@ export class RepetitionCalculator {
   }
 
   private handleRepetitionInstructions(currentRepetitionInstruction: RepetitionInstruction): boolean {
+    if (!this.currentMeasure) {
+      return false;
+    }
```
sschmidTU: we could instead handle this in calculateRepetitions:28, where the currentMeasure is set.

src/Common/Strings/StringUtil.ts
```diff
@@ -1,9 +1,6 @@
 ﻿export class StringUtil {
-  public static StringContainsSeparatedWord(str: string, word: string): boolean {
-    if (str === word ||
-      str.search(" " + word) !== -1 ||
-      str.search(word + " ") !== -1 ||
-      str.search(word + ".") !== -1) {
+  public static StringContainsSeparatedWord(str: string, wordRegExString: string): boolean {
+    if (new RegExp("( |^)" + wordRegExString + "([ .]|$)").test(str)) {
```
sschmidTU: it has been done: one RegEx to rule them all.

bneumann: Nice! But what is that | at the end of the line? Match jadajadajada or EOL?

sschmidTU: You mean `([ .]|$)`? That's white space, dot, or end of line

bneumann: Aah now I see it. Good

PR #384, opened 2018-09-13
chore(Updating out dependencies): Since CI was failing greekeeper kep…

bneumann:
…t failing too

#381, #365, #364

@sebastianhaas check if you agree. Specifically with those tslint to compiler stuff


sschmidTU:
This may not be relevant, but with npm install I get some warnings. I've seen the one about tsify before, a bit annoying, because even installing browserify manually doesn't help.
```console
npm WARN eslint-config-standard@12.0.0 requires a peer of eslint-plugin-node@>=7.0.0 but none is installed. You must install
peer dependencies yourself.
npm WARN eslint-config-standard@12.0.0 requires a peer of eslint-plugin-promise@>=4.0.0 but none is installed. You must install peer dependencies yourself.
npm WARN eslint-config-standard@12.0.0 requires a peer of eslint-plugin-standard@>=4.0.0 but none is installed. You must install peer dependencies yourself.
npm WARN karma-webpack@3.0.4 requires a peer of webpack@^2.0.0 || ^3.0.0 but none is installed. You must install peer dependencies yourself.
npm WARN tsify@3.0.4 requires a peer of browserify@>= 10.x but none is installed. You must install peer dependencies yourself
```

Other than that, i'm all for making greenkeeper happy.

sebastianhaas:
Ah good catch we should fix peer dependencies then. Won't work otherwise.

sschmidTU:
By the way, there are also the warnings about skipping optional dependencies:
```console
npm WARN optional SKIPPING OPTIONAL DEPENDENCY: fsevents@1.2.4 (node_modules\fsevents):
npm WARN notsup SKIPPING OPTIONAL DEPENDENCY: Unsupported platform for fsevents@1.2.4: wanted {"os":"darwin","arch":"any"} (current: {"os":"win32","arch":"x64"})
```
Long story short: Don't expect help from npm.
People have complained to npm for years that this is unnecessary and should be INFO instead of WARN, but npm is apathetic, so you can either use `npm install --no-optional`, live with the warning, or use yarn instead of npm ;)
https://github.com/npm/npm/issues/11632

bneumann:
8 tried to solve the peer dependencies but that's not so easy without giving up on some cool plugins. E.g. The HTML plugin for webpack which is needed for our demo relies on the old tapable hooks

sschmidTU:
If we can't solve peer dependencies, we should merge this, right?

I tried solving peer dependencies with npm-install-peers, but i think actually the packages need to declare the peer dependencies. This seems like a bit of an unpolished concept still in npm.

https://lexi-lambda.github.io/blog/2016/08/24/understanding-the-npm-dependency-model/#Peer_dependency
https://stackoverflow.com/questions/35207380/how-to-install-npm-peer-dependencies-automatically

But i made a PR #394 for this PR that fixes 4 of the 5 warnings.

bneumann:
As per #394 the dependency clashes are fixed

PR #394, opened 2018-09-18
chore: fix 4 of 5 peer dependency warnings by updating dependencies

sschmidTU:
This is a PR for PR #384.
It fixes 4 of the 5 npm warnings for npm install that tell you to install peer dependencies.

The remaining warning seems unsolvable right now, because we also need a newer version of webpack:
```
npm WARN karma-webpack@3.0.4 requires a peer of webpack@^2.0.0 || ^3.0.0 but none is installed.
You must install peer dependencies yourself.
```
karma-webpack upgrades the webpack dependency in 4.0.0, but there's only a release candidate for that right now.

for tsify's dependency on browserify, i had to add browserify to the devDependencies. It's not ideal, because tsify might change that peer dependency in the future and then we can remove browserify as a dependency. But better than a warning, i think.

everything seems to work fine with this (npm install, karma tests, npm start, demo).


sschmidTU:
We need browserify as a peer dependency for tsify. i wrote about it above. i think tsify won't abandon browserify so soon, so it's a decent tradeoff to add it i think.

PR #425, opened 2018-10-12
fix crescendo spacing

sschmidTU:
Spacing between two crescendos is already fixed here:
image

I couldn't figure out how to fix the spacing between crescendo and e.g. a forte though, like in the Entertainer:
image

It's strange because it works well in the function test Expressions Overlap.
image
The problem may be that in the Entertainer, the crescendo ends at the first note of the next bar.
This is a pretty common setup though, also found in Schubert - An die Musik.

I also added an option to [show wedges bounding boxes](https://user-images.githubusercontent.com/33069673/46909121-b57ac500-cf2d-11e8-8300-55704c98f436.png) to the demo.

### DIFF
--- a/src/MusicalScore/Graphical/GraphicalContinuousDynamicExpression.ts
+++ b/src/MusicalScore/Graphical/GraphicalContinuousDynamicExpression.ts
@@ -290,13 +290,6 @@ export class GraphicalContinuousDynamicExpression extends AbstractGraphicalExpre
                 this.lines[0].End.x += value;
                 this.lines[1].End.x += value;
             }
-        } else {
-            if (value < 0) {
-                this.lines[0].Start.x += value;
-            } else {
-                this.lines[0].End.x += value;
-                this.lines[1].End.x += value;
-            }
         }
         this.calcPsi();
     }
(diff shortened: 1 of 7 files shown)


bneumann:
Yeah you basically removed the right squeeze. That's not correct and leads to collision on the following ps and fs.
The bug is in the detection of collisions between 2 dynamics.

sschmidTU:
Ah, I see. I was hoping you could fix the crescendo-f/p spacing anyways ;)

src/MusicalScore/Graphical/GraphicalContinuousDynamicExpression.ts
```diff
@@ -290,13 +290,6 @@ export class GraphicalContinuousDynamicExpression extends AbstractGraphicalExpre
                 this.lines[0].End.x += value;
                 this.lines[1].End.x += value;
             }
-        } else {
-            if (value < 0) {
-                this.lines[0].Start.x += value;
-            } else {
-                this.lines[0].End.x += value;
-                this.lines[1].End.x += value;
-            }
```
sschmidTU: by the way, this change doesn't do anything. the else block can only be triggered by value = 0 which doesn't do anything.

sschmidTU:
So, we looked at the code together and @bneumann most of the issues.
The Entertainer looks great now.

Wedge into dynamic: (bar 6)
image

Diminuendo over two stafflines: (bars 10-11)
image

Cresc. + decresc. wedges have a better spacing fix now too.

image

Also, autoResize in OpenSheetMusicDisplay was fixed.

PR #428, opened 2018-10-17
set autoResize at runtime, fix autoResize handling, demo null checks

sschmidTU:
osmd: autoResize defaults to true, add clear(), isReadyToRender() methods
osmd: fix autoResize handling: don't attach listeners multiple times, don't set timeouts if autoResize disabled

previously, disabling autoResize at runtime didn't work.

index.html: add clear button, improve re-render button size/layout
index.js (demo): null-check buttons for public demo, remove resize code (handled by osmd)

In the demo, try combinations of clear, re-render and this in the console:
`osmd.setOptions({autoResize: false})` (or true)

closes #403 (null-checks for github.io demo which has less buttons. better solved here)
closes #387 (autoResize code in osmd and index.js isn't inconsistent and duplicate anymore)

### DIFF
--- a/demo/index.html
+++ b/demo/index.html
@@ -78,7 +78,7 @@
     </div>
     <div class="column">
         <h3 class="ui header">Show bounding box for:</h3>
-        <select class="ui selection dropdown" id="selectBounding">
+        <select class="ui selection dropdown" id="selectBounding" style="width: 80%">
             <option value="none">None</option>
             <option value="all">All</option>
             <option value="VexFlowMeasure">Measures</option>
@@ -111,8 +111,14 @@
     </div>
     <div class="column">
         <h3 class="ui header">Debug controls:</h3>
-        <div class="ui vertical buttons">
-            <div class="ui button" id="debug-re-render-btn">Re-render</div>
+        <div>
+            <div class="ui vertical buttons">
+                <div class="ui button" id="debug-re-render-btn" style="width: 100%">Re-render</div>
+                
+            </div>
+            <div class="ui vertical buttons">
+                <div class="ui button" id="debug-clear-btn" style="width: 100%">Clear</div>
+            </div>
         </div>
     </div>
 </div>
--- a/demo/index.html
+++ b/demo/index.html
@@ -78,7 +78,7 @@
     </div>
     <div class="column">
         <h3 class="ui header">Show bounding box for:</h3>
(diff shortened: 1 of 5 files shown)


demo/index.html
```diff
@@ -78,7 +78,7 @@ <h4 class="ui header" id="zoom-str">???</h4>
     </div>
     <div class="column">
         <h3 class="ui header">Show bounding box for:</h3>
-        <select class="ui selection dropdown" id="selectBounding">
+        <select class="ui selection dropdown" id="selectBounding" style="width: 80%">
```
bneumann: should be in CSS file

demo/index.html
```diff
@@ -111,8 +111,14 @@ <h3 class="ui header">Show debug information:</h3>
     </div>
     <div class="column">
         <h3 class="ui header">Debug controls:</h3>
-        <div class="ui vertical buttons">
-            <div class="ui button" id="debug-re-render-btn">Re-render</div>
+        <div>
+            <div class="ui vertical buttons">
+                <div class="ui button" id="debug-re-render-btn" style="width: 100%">Re-render</div>
```
bneumann: Also css file

demo/index.html
```diff
@@ -111,8 +111,14 @@ <h3 class="ui header">Show debug information:</h3>
...
+                
+            </div>
+            <div class="ui vertical buttons">
+                <div class="ui button" id="debug-clear-btn" style="width: 100%">Clear</div>
```
bneumann: Also css file

demo/index.js
```diff
@@ -308,7 +268,11 @@ import { OpenSheetMusicDisplay } from '../src/OpenSheetMusicDisplay/OpenSheetMus
     function rerender() {
         disable();
         window.setTimeout(function(){
-            openSheetMusicDisplay.render();
+            if (openSheetMusicDisplay.IsReadyToRender()) {
```
bneumann: Nice!

src/OpenSheetMusicDisplay/OpenSheetMusicDisplay.ts
```diff
@@ -304,6 +320,10 @@ export class OpenSheetMusicDisplay {
     }
 
     //#region GETTER / SETTER
```
bneumann: Actually these are not getters

PR #445, opened 2018-11-01
RelativePosition.Y => RelativePosition.y

dolphinxx:
correct the variable case


bneumann:
How could that have happened? The tslint comlpiler should captured this

sschmidTU:
I don't know, but it seems like this expression isn't that important.
I set a breakpoint and the only times the if block was triggered was when gve.notes.length was 0.
[Tied_Notes.zip](https://github.com/opensheetmusicdisplay/opensheetmusicdisplay/files/2548077/Tied_Notes.zip)

In any case, this is what the code was supposed to be, and it tests well, so i'll merge it.

@dolphinxx thanks for noticing and quickly creating a pull request!

Issue #448, opened 2018-11-03
Question about extensibility and private variables


jeremija:
Hi there, this is a pretty cool library - thanks for making it open source!

I've been trying to use it in a TypeScript project, but have run into a wall pretty quickly. A lot of the interesting variables are `private` and they _can_ be accessed from a JavaScript project or browser console, but cannot from a `TypeScript` source file (which makes sense). One such example is being able to access notes under the cursor via `osmd.cursor.iterator` variable, but that doesn't work in `.ts` because the `iterator` variable is defined as private:

```
  private iterator: MusicPartManagerIterator;
```

and I get the expected error:

```
2341: Property 'iterator' is private and only accessible within class 'Cursor'.
```

I was also thinking about extending the `OpenSheetMusicDisplay` and `Cursor` classes and exposing a method like `get notesUnderCursor()`, but I cannot do that because the `iterator` in `Cursor` class is `private`, not `protected`. Additionally, the `Cursor` class isn't exported.

Are you planning on exposing some of these variables at some point, or making them `protected`, or at least adding more methods that give access to internal information? Right now the only solution I can think of is to fork the project and add the getters myself, but I'm wondering if you might have any better suggestions.

Thanks!

sschmidTU:
Hi, thanks for your interest!
Of course you can always fork the project or create a local branch with your changes.

That being said, you're not the first to be interested in the Cursor and its iterator.
I think there would be no harm in making it public.
@bneumann thoughts?

jeremija:
Hi @sschmidTU, thanks for a quick reply. Making the iterator public would definitely help a lot! But that was just one example. Another one would be adding on-click handlers (a little related to #262) to the rendered notes in `svg`, for example `osmd.graphic` below is also `private`:

```javascript
let ml = osmd.graphic.MeasureList

let tickables = []
ml.forEach(m => m.forEach(mm =>
  Object.keys(mm.vfVoices).forEach(voiceKey =>
    mm.vfVoices[voiceKey].tickables.forEach(t => tickables.push(t))
  )
))

tickables
.filter(t => !!t.attrs.el)
.forEach(t => t.attrs.el.addEventListener('click', () => {
   console.log('clicked on', t)
}))
```

The above might not be the right way of doing it because the click handler seems to have access only to `Vex.Flow.StaveNote`, and not the OSMD's abstraction (which I believe is `Note` from `src/MusicalScore/VoiceData/Note`).

[...]

PR #453, opened 2018-11-06
WIP: Add methods to retreive notes and voices under cursor

jeremija:
Work in progress, please do not merge yet! Related to #448.

Added new methods in `Cursor` class:

- `get AllVoicesUnderCursor()`
- `VoicesUnderCursor(instrument?: Instrument)`
- `NotesUnderCursor(instrument?: Instrument)`

Also added tests.

### DIFF
--- a/src/OpenSheetMusicDisplay/Cursor.ts
+++ b/src/OpenSheetMusicDisplay/Cursor.ts
@@ -5,6 +5,8 @@ import {VexFlowStaffEntry} from "../MusicalScore/Graphical/VexFlow/VexFlowStaffE
 import {MusicSystem} from "../MusicalScore/Graphical/MusicSystem";
 import {OpenSheetMusicDisplay} from "./OpenSheetMusicDisplay";
 import {GraphicalMusicSheet} from "../MusicalScore/Graphical/GraphicalMusicSheet";
+import {Instrument} from "../MusicalScore/Instrument";
+import {Note} from "../MusicalScore/VoiceData/Note";
 
 /**
  * A cursor which can iterate through the music sheet.
@@ -166,5 +168,22 @@ export class Cursor {
 
   public get Hidden(): boolean {
     return this.hidden;
+  }
+
+  public get AllVoicesUnderCursor(): VoiceEntry[] {
+    return this.VoicesUnderCursor();
+  }
+
+  public VoicesUnderCursor(instrument?: Instrument): VoiceEntry[] {
+    return this.iterator.CurrentVisibleVoiceEntries(instrument);
+  }
+
+  public NotesUnderCursor(instrument?: Instrument): Note[] {
+    const voiceEntries: VoiceEntry[]  = this.VoicesUnderCursor(instrument);
+    const notes: Note[] = [];
+    voiceEntries.forEach(voiceEntry => {
+      notes.push.apply(notes, voiceEntry.Notes);
+    });
+    return notes;
   }
 }
--- a/test/Common/OSMD/OSMD_Test.ts
+++ b/test/Common/OSMD/OSMD_Test.ts
@@ -1,5 +1,8 @@
 import chai = require("chai");
+import {Instrument} from "../../../src/MusicalScore/Instrument";
+import {Note} from "../../../src/MusicalScore/VoiceData/Note";
 import {OpenSheetMusicDisplay} from "../../../src/OpenSheetMusicDisplay/OpenSheetMusicDisplay";
+import {VoiceEntry} from "../../../src/MusicalScore/VoiceData/VoiceEntry";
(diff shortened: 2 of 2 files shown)


sschmidTU:
nice, keep it up!
I'll add some code comments.

src/OpenSheetMusicDisplay/Cursor.ts
```diff
@@ -167,4 +169,21 @@ export class Cursor {
...
     return this.hidden;
   }
+
+  public get AllVoicesUnderCursor(): VoiceEntry[] {
+    return this.VoicesUnderCursor();
+  }
+
+  public VoicesUnderCursor(instrument?: Instrument): VoiceEntry[] {
+    return this.iterator.CurrentVisibleVoiceEntries(instrument);
+  }
```
sschmidTU: First of all, one of these methods is unnecessary because they do the same thing.
Secondly, this method is just syntactic sugar or a simplication for this.iterator.CurrentVisibleVoiceEntries, so it's not really necessary, but it might make some code look better.

sschmidTU: By the way, where is one supposed to get the Instrument parameter from?

src/OpenSheetMusicDisplay/Cursor.ts
```diff
@@ -167,4 +169,21 @@ export class Cursor {
...
+
+  public NotesUnderCursor(instrument?: Instrument): Note[] {
+    const voiceEntries: VoiceEntry[]  = this.VoicesUnderCursor(instrument);
+    const notes: Note[] = [];
+    voiceEntries.forEach(voiceEntry => {
+      notes.push.apply(notes, voiceEntry.Notes);
+    });
+    return notes;
+  }
```
sschmidTU: this method is a bit more complex, so while not strictly necessary, if you use it often, it makes sense to put it into a method.
On the other hand, OSMD itself doesn't need these utility methods, so one could ask why we need it in OSMD and people couldn't just implement this as a static utility function themselves.
But we've seen multiple requests for more Cursor utility functions, so why not.

jeremija: You're right. I didn't realize everything necessary is now available via public getters. One could just write something like:

```typescript
function getVoicesUnderCursor (instrument?: Instrument): VoiceEntry[] {
    osmd.cursor.Iterator.CurrentVisibleVoiceEntries(instrument)
}

function getNotesUnderCursor (instrument?: Instrument) {
    const voiceEntries: VoiceEntry[]  = this.VoicesUnderCursor(instrument);
    const notes: Note[] = [];
    voiceEntries.forEach(voiceEntry => {
      notes.push.apply(notes, voiceEntry.Notes);
    });
    return notes;
}
```

However, the above definitions would fail because `Instrument`, `VoiceEntry` and `Note` are not exported AFAIK. To define these methods, we need access to the `Instrument` type, but I'd also export `Note` and `VoiceEntry`, because without it we'd have to solely rely on TypeScript's type inference when calling these methods. For example, we'd be able to write:

```typescript
const notes = getNotesUnderCursor()
```

but not:

```typescript
const notes: Note[] = getNotesUnderCursor()
```

src/OpenSheetMusicDisplay/Cursor.ts
jeremija: > First of all, one of these methods is unnecessary because they do the same thing.
Secondly, this method is just syntactic sugar or a simplication for this.iterator.CurrentVisibleVoiceEntries, so it's not strictly necessary. Though it might make some code look better.

You're right. I didn't realize we have access to all that internal stuff from the outside now.

> By the way, where is one supposed to get the Instrument parameter from?

I used the `osmd.Sheet.Instruments` array in the tests.

jeremija:
Closing this as Iterator is now publicly accessible via `osmd.cursor.Iterator`, and there's no need for these helper methods.

sschmidTU:
Thanks! Just to be clear, I don't think these helper methods are a bad idea. If people use functions like getting the notes under the cursor a lot, it makes sense to make an easy and clean method out of it. So i personally wouldn't mind merging this, with a few adjustments. Though on the other hand we'll have to see if there's demand for this, and if people don't just write these functions themselves.

src/OpenSheetMusicDisplay/Cursor.ts
sschmidTU: actually, these are exported now.

sschmidTU:
Reopening because I myself had to remember how to get the notes under the cursor, and these methods are short and probably quite useful to have.

sschmidTU:
Merged with a few changes. Thanks again, @jeremija !

PR #460, opened 2018-11-17
Add fix for cursor.next() when an instrument or voice is hidden

jeremija:
Calling `cursor.next()` might throw an error when an instrument is invisible (i.e. `osmd.Sheet.Instruments[0].Visible = false` is set and `osmd.render()` is called):

```
TypeError: Cannot read property 'PositionAndShape' of undefined
    at Cursor../src/OpenSheetMusicDisplay/Cursor.ts.Cursor.update:73
```

This happens if all of the voice entries under cursor are hidden. In other words, when the currently visible voices under cursor have no notes, but the hidden ones do.

I solved this problem by modifying the `Cursor#update()` method to use `iterator.CurrentVisibleVoiceEntries()` instead of `iterator.CurrentVoiceEntries`.

Also added a test which will fail if changes in `Cursor.ts` are reverted.

### DIFF
--- a/src/OpenSheetMusicDisplay/Cursor.ts
+++ b/src/OpenSheetMusicDisplay/Cursor.ts
@@ -64,7 +64,11 @@ export class Cursor {
     let x: number = 0, y: number = 0, height: number = 0;
 
     // get all staff entries inside the current voice entry
-    const gseArr: VexFlowStaffEntry[] = iterator.CurrentVoiceEntries.map(ve => this.getStaffEntriesFromVoiceEntry(ve));
+    const gseArr: VexFlowStaffEntry[] = iterator.CurrentVisibleVoiceEntries().map(ve => this.getStaffEntriesFromVoiceEntry(ve));
+    if (!gseArr.length) {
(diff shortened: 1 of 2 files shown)


sschmidTU:
Great change, thanks!
One issue still is that the cursor stays on the visible note for multiple steps if there's notes with smaller note values in the invisible part. Because if everything is visible, you want to go to the next smallest time step in any part. But that may be another issue in the code. I don't think the cursor was made for extensive visibility changes.

jeremija:
Glad I was able to contribute! 

> One issue still is that the cursor stays on the visible note for multiple steps if there's notes with smaller note values in the invisible part. 

True, I was using this to skip the hidden parts:

```
cursor.Iterator.moveToNextVisibleVoiceEntry(noRest)
if (!cursor.Hidden) {
    cursor.show()
}
```

Do you think `cursor.next()` should be updated to use `iterator.moveToNextVisibleVoiceEntry()` instead of `iterator.moveToNext()`? We might want to be able to hide some tracks from the view, but still be able to play the notes from the hidden tracks.

Issue #464, opened 2018-11-23
Chord symbols are not to be taken into account for system bounding box size


matt-uib:
then the distance to the previous system is too small.
i guess it also is not taken into account for in-system staffline distances

bukaznik:
This should also fix the composer
![screenshot 2018-11-23 16 10 52](https://user-images.githubusercontent.com/14196717/48950099-610a3280-ef3a-11e8-907a-a27c68f64d8b.png)

PR #467, opened 2018-11-25
feat(ChordSymbols): Chord symbols are now updating their bounding box…

bneumann:
… in the skyline.
#464 

This should at least remedy most collisions. What I didn't include is the dynamic positioning of the chord labels. As @sschmidTU said they are only place at a fixed distance above the staffline. I downloaded some examples from musescore which looked quite OK with it. 
@bukaznik: Can you send me your files for further testing?


sschmidTU (approved):
Looks good, works.
Together with the new metronome mark feature, all christmas sample scores render well.

bneumann:
Coolio

PR #493, opened 2019-02-23
Fix/ajax timeout

bneumann:


sschmidTU:
LGTM.
What is IE? Internet Explorer?
I assume you tested that part.

bneumann:
No I didn't, am not mad and install IE on my pristine system. 😁 But I read the docs for http requests and in a footnote was mentioned that you should support timeouts as well.

sschmidTU:
You might have tested IE on a VM with Windows XP for "legacy" support ;)

bneumann:
True I still have one somewhere. But it's a little change anyways

Issue #482, opened 2019-01-05
[Feature Request] Ability to render parts of a score


turnerhayes:
This may be prohibitively difficult, but it would be really neat if this library could render individual segments of a score (I'm thinking single measures). A use case for this would be if you wanted to progressively and interactively show a piece as it was playing.

loganknecht:
This would be amazing! +1 on this request!

It would be fantastic if we could arbitrarily select measures to render. I know that ties, slurs, etc. might make it difficult, but it would be fantastic to select any possible combination.

For those instances, I'm of a mind where I'd say they wouldn't need too be rendered,.

It's usage might be like
'osmd.render()'

But then also something like
'osmd.render([1,2,3])'
or
'osmd.render([1,5,9,11])'

sschmidTU:
This would be awesome, yes, though it may well require huge changes in the code. I've looked at this before, the problem is that many different methods try to access the whole xml and score.

There is already an option to draw measures 1 to X, where you specify X. Have you seen that?
[See OSMDOptions](https://github.com/opensheetmusicdisplay/opensheetmusicdisplay/blob/develop/src/OpenSheetMusicDisplay/OSMDOptions.ts)

Of course drawing arbitrary measures would be nice, but it might be hard. Maybe i'll look at this again in the near future, because several people would like to have this.

In the meantime, you could try various programs to extract specific measures from a piece/XML and make it into a new XML.

loganknecht:
@sschmidTU do you have specific library/program suggestions?

Right now I'm using music21, a python library, to break a song into single measures and it works, but it's just ok. Do you know if there is a JavaScript specific library to do the same or have any guidance?

Also very cool feature from 1 to X! I didn't know that!

sschmidTU:
I don't have library suggestions for this, sorry. I would use Musescore for this to do it manually, but i haven't needed to extract bars from pieces automatically yet.
Good to know that music21 can do this though. I've heard of it before, the maker is an active contributor to Vexflow too.

turnerhayes:
Thanks for the suggestion and the pointer to `drawUpToMeasureNumber`! Maybe you could add a `drawFromMeasureNumber` option (or something). It looks like it would be pretty easy to implement; it seems like you can just change [this line](https://github.com/opensheetmusicdisplay/opensheetmusicdisplay/blob/develop/src/MusicalScore/Graphical/MusicSheetCalculator.ts#L674) to be something like

```
[...]

PR #528, opened 2019-05-30
Added feature `MinMeasureToDrawIndex`

praisethemoon:
Hello,

I have added a new feature I personally wanted, allowing to render a particular interval of measures. Now,  `MaxMeasureToDrawIndex` is already implemented. I simply wanted to add its opposite `MinMeasureToDrawIndex`.

There is one thing that I want to ask you, how should I handle the case where `MinMeasureToDrawIndex > MaxMeasureToDrawIndex` ? Currently, an exception `TypeError: this.graphicalMusicSheet.MusicPages[0] is undefined` is thrown automatically. Other than that, here is a screenshot of the feature in action.

Screenshot 2019-05-30 at 03 35 25

### DIFF
--- a/src/MusicalScore/Graphical/EngravingRules.ts
+++ b/src/MusicalScore/Graphical/EngravingRules.ts
@@ -190,6 +190,7 @@ export class EngravingRules {
     private defaultColorLabel: string;
     private defaultColorTitle: string;
     private maxMeasureToDrawIndex: number;
+    private minMeasureToDrawIndex: number;
     /** Whether to render a label for the composer of the piece at the top of the sheet. */
     private renderComposer: boolean;
     private renderTitle: boolean;
@@ -406,6 +407,7 @@ export class EngravingRules {
         this.defaultColorLabel = this.defaultColorNotehead;
         this.defaultColorTitle = this.defaultColorNotehead;
         this.maxMeasureToDrawIndex = Number.MAX_VALUE;
+        this.minMeasureToDrawIndex = 0;
         this.renderComposer = true;
         this.renderTitle = true;
         this.renderSubtitle = true;
@@ -1403,6 +1405,12 @@ export class EngravingRules {
     }
     public set MaxMeasureToDrawIndex(value: number) {
         this.maxMeasureToDrawIndex = value;
+    }
+    public get MinMeasureToDrawIndex(): number {
+        return this.minMeasureToDrawIndex;
+    }
+    public set MinMeasureToDrawIndex(value: number) {
+        this.minMeasureToDrawIndex = value;
     }
     public get RenderComposer(): boolean {
         return this.renderComposer;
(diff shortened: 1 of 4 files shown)


bneumann:
Awesome work! I personally have no preference on how to handle the error. In the skybottom calculator I simple switched min and max if min > max but now you could argue if that error tolerant or confusing 😂

Maybe @matt-uib has any preferences?

sschmidTU:
Awesome! I had tried to implement selecting a first and last measure to render, but i always had errors in functions that tried to access measures out of the range. You fixed two of them.

During testing i actually found a few more of these issues and fixed them. So i'll merge this PR to a feature branch and fix it up for merging to develop.

By the way, minMeasureToDrawIndex should be drawFromMeasureNumber - 1, because measure 1 has index 0.

@praisethemoon btw, is your nickname a Dark Souls/Bloodborne reference? Love it.
And thanks for submitting the PR, without it this wouldn't have been done for a while!

sschmidTU:
Oh and when the minimum measure to draw is greater than the max measure length, i just set it to max length - 1 and give a debug warning on console. otherwise we have 0 measures to draw and get the error you mentioned.
of course we could change that. drawing nothing would be the logical choice, but could be confusing.

sschmidTU:
I'll make a new release as well after this is merged, because we've collected a few nice improvements again.

praisethemoon:
@bneumann @sschmidTU I am glad you've found this PR useful!

@sschmidTU You got me  :sweat_smile: I am a huge fan of Miyazaki's works

I have done a couple more tests and I have found two issues related to Cursors and Min/Max measure which I plan to fix although, i am very new to OSMD and TypeScript, so ideas are more than welcome ❤️ . Here are the issues I've found:

1. Changing the `MinMeasureToDrawIndex` from `0` to any other value, then re-rendering `osmd` works just fine. However, when rendering the cursor, an exception is raised within the Cursor class with message `TypeError: this.graphicalMusicSheet.MusicPages[0] is undefined`, and the cursor wont be able to be updated any longer. However, if we redo the same process but first incrementing the cursor until it reaches the `MinMeasureToDrawIndex` measure, then changing it within `osmd`, it will work just fine.

2. The same error applies to `MaxMeasureToDrawIndex`, when the cursor is in a measure that is higher than the specified maximum.

What is the best way to approach these two issues?

sschmidTU:
Ah, yes, i haven't tested the cursor with this yet, makes sense that it needs adjustments.
I'm just about to push my fixes so far for this PR to the new branch.
I'll look at the cursor after that.

ps: I'm a huge fan of Miyazaki's works as well 😄

praisethemoon:
I will keep trying to fix it as well, it would help me learn the architecture and internal design of the library. Meanwhile, you should try Sekiro  :video_game: 😆

sschmidTU:
this will close #482 by the way, we'll make a lot of people happy by implementing this feature ;)

and yes, i'm dying to play Sekiro, just didn't get to it yet :)

sschmidTU:
my fixes are pushed to the branch feat/drawFromMeasureNumber, take a look if you're interested.
it's even possible to draw any single measure from a piece now.
let's see if we can fix the Cursor as well. though for today this was it for me.

sschmidTU:
for testing, you can decommentate the lines in index.js.selectSampleOnChange(), which will draw a random range of measures.

sschmidTU:
@praisethemoon by the way, for quick (and/or random) chatting, there's also gitter ;)
https://gitter.im/opensheetmusicdisplay/opensheetmusicdisplay

praisethemoon:
Thanks for tips and the invitation! Will try your suggestions :)

sschmidTU:
@praisethemoon I finally got around to merging in your work to develop with my fixes, sorry for the delay! release will be out soon as well. see #482

Issue #576, opened 2019-10-11
Add support to use OSMD embedded into iframes


FloHin:
OSMD should be able to be run embedded into an iframe (of another website).

In future, also loading XMLs from other origins should be allowed, for now the following is needed:
- show in a compact embedded mode and hide unnecessary widgets and selects
-  handle window parameters for showingControls, showing zoom controls and zoom
- handle window parameter to load the specified xml file

FloHin:
Done and merged.

PR #577, opened 2019-10-11
Feature #576: Added basic support for iframe embedding

FloHin:
- add parameter handler for embedded, controls, header and zoomControls
- add option to set zoom via parameter
- add feature to display compact zoomControl widget

### DIFF
--- a/demo/embedded_demo.html
+++ b/demo/embedded_demo.html
@@ -1,0 +1,93 @@
+<!doctype html>
+<html lang="en">
+<head>
+    <meta charset="utf-8">
+    <meta name="viewport" content="width=device-width, initial-scale=1">
+
+    <title>OpenSheetMusicDisplay as an embedded iframe</title>
+    <meta name="description" content="A showcase for OpenSheetMusicDisplay.">
+    <meta name="author" content="OpenSheetMusicDisplay contributors">
+
+    <script src="https://cdnjs.cloudflare.com/ajax/libs/jquery/3.3.1/jquery.min.js" type="text/javascript"></script>
+    <script src="https://cdnjs.cloudflare.com/ajax/libs/semantic-ui/2.3.0/semantic.min.js"
+            type="text/javascript"></script>
+    <link href="https://cdnjs.cloudflare.com/ajax/libs/semantic-ui/2.3.0/semantic.css" media="all" rel="stylesheet"/>
+    <link rel="icon" href="./favicon.ico?" type="image/x-icon"/>
+</head>
+<body>
+<br>
+<div style="text-align: center;width: 60%;margin:0 auto">
+    <h1 class="ui centered header" id="header">
+        <img src="./favicon.ico?" class="ui image">
+
+    </h1>
+    <h2>
+        OpenSheetMusicDisplay as an embedded iframe
+    </h2>
+</div>
+<div style="padding: 40px;margin:0 auto">
+    <h3>
+        Use parameters:
+    </h3>
+    <p>
+    <table class="ui table celled ">
+        <tr>
+            <th>Parameter</th>
+            <th>Values</th>
+            <th>Default</th>
+            <th>Description</th>
+        </tr>
+        <tr>
+            <td><b>embedded</b></td>
+            <td>"0" and "1"</td>
+            <td>"0" = OFF</td>
+            <td>Activates the embedded mode: controls are hidden (unless other params are given) and sheet will be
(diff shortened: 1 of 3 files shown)


demo/embedded_demo.html
```diff
@@ -0,0 +1,93 @@
...
+            <td>"0" = OFF</td>
+            <td>Activates the embedded mode: controls are hidden (unless other params are given) and sheet will be
+                rendered more compactly
+            </td>
+        </tr>
+        <tr>
+            <td>showControls</td>
+            <td>"0" and "1"</td>
+            <td>"0" = OFF</td>
+            <td>If active, the controls of OSMD are shown like usually.</td>
```
sschmidTU: some controls like zoom level are shown, not all OSMD controls like options, re-render and so on.

demo/embedded_demo.html
```diff
@@ -0,0 +1,93 @@
...
+        </tr>
+        <tr>
+            <td>showZoomControl</td>
+            <td>"0" and "1"</td>
+            <td>"0" = OFF</td>
+            <td>If showControls == 0 and showZoomControl == 1, a small zoomControl widget will be display to safe
```
sschmidTU: displayed

demo/embedded_demo.html
```diff
@@ -0,0 +1,93 @@
...
+            <td>showHeader</td>
+            <td>"0" and "1"</td>
+            <td>"0" = OFF</td>
+            <td>If activated, the OSMD title will be rendered on top of the sheet</td>
+        </tr>
+        <tr>
+            <td>zoom</td>
+            <td>floating number between 0.1 and 5.0</td>
+            <td>1.0</td>
+            <td>If given, the zoom will be set at loading time</td>
```
sschmidTU: zoom -> OSMD zoom level. no biggie, but it could be unclear what kind of zoom is used

demo/index.js
```diff
@@ -48,7 +48,9 @@ import { OpenSheetMusicDisplay } from '../src/OpenSheetMusicDisplay/OpenSheetMus
     },
 
     zoom = 1.0,
-    // HTML Elements in the page
```
sschmidTU: why remove the comment?

demo/index.js
```diff
@@ -256,14 +329,15 @@ import { OpenSheetMusicDisplay } from '../src/OpenSheetMusicDisplay/OpenSheetMus
         if (!isCustom) {
             str = sampleFolder + selectSample.value;
         }
-        zoom = 1.0;
+        // zoom = 1.0;
```
sschmidTU: awesome. not resetting the zoom level between sample changes is more correct anyways. if anything, we need a reset zoom button in the (main) demo.

sschmidTU:
Looks great! I just found a few minor points, see comments.
When the user hasn't started the local server for OSMD, should we show some kind of notice? Maybe in the div or console?
image
We could say something like `OSMD was expected to run at localhost:[port], could not connect`.
Also, maybe we could put that information somewhere in the embedded demo.html.

sschmidTU:
this closes #576

demo/embedded_demo.html
matt-uib: we need to define now what controls and other stuff we want to allow the embedder to customize/hide/show.
I would say only developers would need something like re-render which should not be part of the embedding options.

matt-uib: But on this demo page should change the text "the controls of OSMD are shown like usually" to be more correct.
Also the "Default" values given don't seem to be correct, right? At least the showZoomControl seems to be on by default, but off is written there.

demo/embedded_demo.html
```diff
@@ -0,0 +1,93 @@
...
+            <td>"auto"</td>
+            <td>If given, this attribute will refresh the state of the css attribute overflow to control the
+                scrollbar.
+            </td>
+        </tr>
+    </table>
+</div>
+<div>
+    <center>
+        <iframe src="http://localhost:8000/?embedded&showZoomControl=1&zoom=0.5" width="800" height="1500"
```
matt-uib: Frame-border should be possible to be switched off.
Actually we might need: thickness and color of border.
Nice would be also sheet background color.

FloHin: This has to be defined by the window which displays the iframe! Not inside, but from the outside

demo/index.js
FloHin: Added it again

PR #622, opened 2019-12-04
Allow updating the graphical sheet from the music sheet

simonspoerri:
This would allow updating the model in the sheet and rerendering the graphical sheet thereof, thus allowing mutations on the OSMD model instead of the graphical model directly. However, this still requires a full rerender.


sschmidTU:
Looks great, thanks for the PR! merging.

Issue #625, opened 2019-12-09
OSMD render extra whole note without musicxml instruction


DavidQinqianWu:
OSMD: 
<img width="1682" alt="截屏2019-12-09下午5 48 46" src="https://user-images.githubusercontent.com/23088138/70426018-45be8680-1aad-11ea-9075-50dbf57fef2e.png">
museScore:  which is original score should look like
![截屏2019-12-09下午5 57 04](https://user-images.githubusercontent.com/23088138/70426060-5969ed00-1aad-11ea-9363-f87b5e0a69c0.png)
I realize that osmd Seems will automatically put the " whole rest" note on rendering. if there are no notes information on staff 1. I just check the musicxml file which does not include the "whole rest note" information on the first 4 measures. here is the musicxml file 
[autoAddWholeRestNote.musicxml.zip](https://github.com/opensheetmusicdisplay/opensheetmusicdisplay/files/3939025/autoAddWholeRestNote.musicxml.zip)

sschmidTU:
Ah, yes, OSMD automatically places a whole rest when there are no notes in a measure.
We should probably not do that because of samples like yours.
The reason we do it is that in Vexflow a voice needs to be filled with notes that fill the measure.

We should see if we can add a ghost note instead (which is invisible).
There could be an OSMDOption to show a whole rest or not, but i think MusicXML pieces pretty much always contain whole rests when they are supposed to be rendered anyways.

DavidQinqianWu:
Thanks!  that would be perfect

sschmidTU:
Fixed.

The default option is now not to fill empty measures with rest notes.
![image](https://user-images.githubusercontent.com/33069673/70812908-d4424900-1dc8-11ea-9f39-cb374a31a9da.png)

there's also a new option fillEmptyMeasuresWithWholeRest:
```js
osmd.setOptions({fillEmptyMeasuresWithWholeRest: 0}); // new default: don't fill, leave measure empty
osmd.setOptions({fillEmptyMeasuresWithWholeRest: 1}); // as before
osmd.setOptions({fillEmptyMeasuresWithWholeRest: 2}); // fill with rests, but make them invisible
// or of course in TypeScript:
osmd.setOptions({fillEmptyMeasuresWithWholeRest: FillEmptyMeasuresWithWholeRest.YesVisible}); // typescript, = 1
```

PR #626, opened 2019-12-10
Add new render options (place ghost whole rest not or visible whole rest note)

DavidQinqianWu:
Hi @sschmidTU  According to your suggestion on  #625 
I have added osmd render option, which could choose to print visible auto place whole rest note, or invisible


sschmidTU:
Your pull request was fine, thanks for helping!
however i actually decided to add further options, see #625.
so unfortunately this PR is now not needed anymore. Sorry! Still appreciate the input!

PR #639, opened 2019-12-26
UI Fix of Follow Cursor feature

praisethemoon:
When I introduced the PR that added the follow cursor mechanism, I have set the scroll behavior to "smooth". When changing the cursor too often while the view is so far away, it will be sluggish and possibly wont even scroll to the element because of the frequent successive calls of `scrollIntoView` which makes it stop for a while on every call (at least on chrome). On my lastest commit (https://github.com/opensheetmusicdisplay/opensheetmusicdisplay/commit/691b4bb8f4c2dc4a4af72aa96cff596577136336), I added a test of how far the cursor element is from the view. If it is too far, we do an "auto' scroll which jumps straight into the cursor.


sschmidTU:
i've merged it onto a fix branch for now to add some refactors, thanks @praisethemoon !

sschmidTU:
Somehow the merge to the branch was incomplete, only one file of 6 were changed. Sorry, will have to fix this

sschmidTU:
@praisethemoon I'm confused, the merge of this PR effectively only modified one file, though it says 6 files were changed. Was part of this PR merged earlier already? I tried two more PRs to fix this on a branch, but it seems like that was unnecessary. Sorry for the confusion ^^

praisethemoon:
@sschmidTU The other files are previous PRs, I AM sorry for the confusion, I should have created a new branch! I just kept pushing stuff into the same old one sorry 😅

sschmidTU:
No worries! Still strange that Github shows more changes in the PR than the merge will effect. Anyways, not so important (=

Issue #559, opened 2019-08-14
Multiple OSMD instances inevitably share their settings


dnlfrst:
# Issue
***Note:*** *this issue is illustrated in a [minimal working example](https://codesandbox.io/s/osmd-issue-558-t4503).*

## Assumption
- I want to display a full-blown music sheet (e.g. _Beethoven's_ "_[An die ferne Geliebte](https://github.com/opensheetmusicdisplay/opensheetmusicdisplay/blob/develop/test/data/Beethoven_AnDieFerneGeliebte.xml)_") and a fragment of the very same (e.g. the very first measure) on the same page.
- As the fragment consists of a single measure only, I want to have it displayed compactly and not show the title, the composer, or the part names.

## Implementation Detail
Both, the full-blown music sheet and the fragment, receive a separate [`OpenSheetMusicDisplay`](https://opensheetmusicdisplay.github.io/classdoc/classes/opensheetmusicdisplay.html) instance with their own container.  
Now, since the fragment should be compact and not receive a title, a composer, or part name, I remove them using the appropriate options for the respective `OpenSheetMusicDisplay` instance. The full-blown music sheet does not receive any options (i.e. the default options):

```javascript
const musicSheetFragmentOptions = {
  drawComposer: false,
  drawPartAbbreviations: false,
  drawPartNames: false,
  drawSubtitle: false,
  drawTitle: false,
  darwingParameters: DrawingParametersEnum.compact
};

const musicSheetFragment = new OpenSheetMusicDisplay(
  "music-sheet-fragment",
  musicSheetFragmentOptions
);
const fullMusicSheet = new OpenSheetMusicDisplay("full-music-sheet");
```

The [result of the above](https://codesandbox.io/s/osmd-issue-558-t4503) is that both `OpenSheetMusicDisplay` instances receive the default options while the given ones are ignored.

# Failed Attempt at a Solution
- Setting the default options on the second instance explicitly,
- lazily setting the options after rendering, or
- changing the rendering order

does not do the trick. All of which does not come to my surprise since the [`engraving rules`](https://github.com/opensheetmusicdisplay/opensheetmusicdisplay/blob/develop/src/MusicalScore/Graphical/EngravingRules.ts) as the underlying driver of the settings are implemented as a [Singleton](https://github.com/opensheetmusicdisplay/opensheetmusicdisplay/blob/develop/src/MusicalScore/Graphical/EngravingRules.ts#L437).

# Questions
[...]

PR #647, opened 2020-01-09
Allow for multiple OSMD instances on the same page

dnlfrst:
# What does this PR contribute?
The PR allows for multiple OSMD instances with different `OSMDOptions` on the same page. This was not possible until now as  described in #559.

# How does this PR achieve its contribution?
The PR adds a reference to an instance of `EngravingRules` to the following classes:

- `OpenSheetMusicDisplay`: contains an instance of `EngravingRules` which was previously only provided statically
- `SourceMeasure` as [suggested](https://github.com/opensheetmusicdisplay/opensheetmusicdisplay/issues/559#issuecomment-544885999) by @matt-uib to cover most of the uses of `EngravingRules`
- `MusicSheetReader` to set `rules` of `MusicSheet`
- `DrawingParameters` to a use of `EngravingRules` where no other class instance with a reference to `EngravingRules` is available
- `SvgVexFlowBackend` to a use of `EngravingRules` where no other class instance with a reference to `EngravingRules` is available

The PR also adds a test case to verify that multiple instance of OSMD on the same page can feature different `OSMDOptions`.

# What problems still persist with this PR?

There are three occasions ([here](https://github.com/dnlfrst/opensheetmusicdisplay/blob/fix/entangled-engraving-rules/src/MusicalScore/Graphical/DrawingParameters.ts#L23), [here](https://github.com/dnlfrst/opensheetmusicdisplay/blob/fix/entangled-engraving-rules/src/MusicalScore/Graphical/GraphicalVoiceEntry.ts#L24), and [here](https://github.com/dnlfrst/opensheetmusicdisplay/blob/fix/entangled-engraving-rules/src/MusicalScore/ScoreIO/MusicSheetReader.ts#L28)) where I conditionally instantiate `EngravingRules`. Could anyone verify that this does not pose any issues?

# What is still left to do?

- [x]  Remove the [use](https://github.com/opensheetmusicdisplay/opensheetmusicdisplay/blob/develop/src/MusicalScore/Graphical/GraphicalLabel.ts#L44) of `EngravingRules` in `GraphicalLabel`. The class does not contain another class instance with a reference to the `EngravingRules` instance. I could not come up with a practical solution. Any ideas?
- [x]  Remove the [use](https://github.com/opensheetmusicdisplay/opensheetmusicdisplay/blob/develop/src/MusicalScore/Graphical/VexFlow/VexFlowConverter.ts#L656) of `EngravingRules` in `VexFlowConverter`. The enclosing static method is used only two times but I could not come up with a practical solution. Any ideas?
- [x] Remove the static instance of `EngravingRules`.

# When is this PR ready to be merged?
As soon as the above problems are mitigated and someone reviewed this PR. @matt-uib @sschmidTU

### DIFF
--- a/src/MusicalScore/Graphical/AbstractGraphicalExpression.ts
+++ b/src/MusicalScore/Graphical/AbstractGraphicalExpression.ts
@@ -11,7 +11,7 @@ export abstract class AbstractGraphicalExpression extends GraphicalObject {
     /** Internal cache of read expression */
     protected expression: AbstractExpression;
     /** EngravingRules for positioning */
-    protected rules: EngravingRules = EngravingRules.Rules;
+    protected rules: EngravingRules;
 
     constructor(parentStaffline: StaffLine, expression: AbstractExpression) {
         super();
@@ -19,6 +19,7 @@ export abstract class AbstractGraphicalExpression extends GraphicalObject {
         this.boundingBox = new BoundingBox(this, parentStaffline.PositionAndShape);
         this.parentStaffLine = parentStaffline;
         this.parentStaffLine.AbstractExpressions.push(this);
+        this.rules = parentStaffline.ParentMusicSystem.rules;
     }
 
     /** Graphical label of the expression if available */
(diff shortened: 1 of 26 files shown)


sschmidTU:
Thank you for the PR! I will review the changes and try to solve the remaining TODOs and conflicts. This will be a very useful fix.
And sorry for the late response, I did not get to work on OSMD recently.

sschmidTU:
@dnlfrst I concluded the merge in PR #721, it will be merged into develop and released in OSMD 0.7.5 soon. Thanks again for your initial work! Even though i still had to do quite a few adjustments, this was very helpful.

Issue #549, opened 2019-07-18
How to get the HTML element id of a specific note?


Abefocuson:
for example 

`<g class="vf-stavenote" id="vf-auto37980"> ... </g>`

how to get "vf-auto37980" ?

sschmidTU:
parse it from the SVG using a regular expression.
What do you want to do with this HTML tag or id string though?
I don't see how that would be useful.

sschmidTU:
Or use an XML Parser of course

Abefocuson:
in this way i can get the specific note in OSMD:

osmd.sheet.sourceMeasures[0].verticalSourceStaffEntryContainers[0].staffEntries[0].voiceEntries[0].notes

and I want to get the  HTML element id of this note. Is there a method for me to call to get this id?

sschmidTU:
No. Why would you want the html element for this? do you want to remove the child element from the SVG? That's the only use i can see for this.
This id is probably generated by the Vexflow SVG Renderer and seems to have an automatically generated id.

If you search through the HTML Elements in the Google Developer Tools, you can see the note highlighted for the vf-stavenote element.
![image](https://user-images.githubusercontent.com/33069673/61460537-3492a300-a96f-11e9-8dd0-2e1e1e267ecb.png)

But if you want to find this out programmatically, you probably need to do some bounding box calculations.
You can get the OSMD bounding box with note.PositionAndShape, though it will be different from the Vexflow bounding box. One OSMD distance unit is 10 pixels in Vexflow.

sschmidTU:
If you just want to remove a note, call `note.noteheadColor = "#00000000"` (transparent) and `osmd.render()`.
In theory, you should be able to set `note.PrintObject = false` to make it invisible, but re-rendering for that doesn't work yet, which i can hopefully fix with the transparent color method soon.

misterallx:
> If you just want to make a note invisible, call `note.PrintObject = false` and `osmd.render()` (included in next release).
> Or set `note.noteheadColor = "#00000000"` (transparent color)

When will the next release be available?
Thank you!

sschmidTU:
> When will the next release be available?

hopefully in the next few days.
You should already be able to use the noteheadColor method in the current release though.

dnlfrst:
# Use Case

I do not agree with @sschmidTU's standpoint:

> No. Why would you want the html element for this? do you want to remove the child element from the SVG? That's the only use i can see for this.

[...]

PR #659, opened 2020-01-22
Added a reference to `GraphicalNote` from `Note`

praisethemoon:
Hello!

I believe that this tiny change would really prove usefull as it is strongly related to https://github.com/opensheetmusicdisplay/opensheetmusicdisplay/issues/549 in the sense that it would be easier to get the graphical note object from, for example cursor using`osmd.cursor.iterator.currentVoiceEntries[0].notes[0].graphicalNote.getSVGGElement().id` which is more convinient than fetching it from `osmd.GraphicSheet.MeasureList[0][0].staffEntries[0].graphicalVoiceEntries[0].notes[0]`. If there is something that already exists and solves this problem, please let me know!

For me this is the best way to hack into the low-level SVG generated objects to change their style, and I do hope you find it useful!

Cheers!


sschmidTU:
Hey, i basically answered this in a new issue in #660.
The thing is, we wanted to avoid creating references to graphical classes in sheet model classes, to modularize OSMD and make it possible to use the sheet model without the graphics classes.
It should be almost as easy with the solution in #660 though. We can also add a static method to Cursor or GraphicalNote that gets you the GraphicalNote from a Note via the new dictionary.

Jack-Works:
How does this PR going? 👀

sschmidTU:
@Jack-Works As i said, we do not want a direct reference to graphical classes in our sheet model classes, but hopefully #659 will get resolved and then you will have a way to find a GraphicalNote from a Note that is almost as easy as a reference in Note.

ice6:
@praisethemoon this is a great feature, when you design an interactive sheet music app.

@sschmidTU this library is a bridge layer between `musicxml` and `vexflow`, it is inevitable that you have to interact with the ui layer(vexflow). 

Maybe we can add a `GraphicalNote ` property to `Note.ts` with datatype `any`, if you don't want to make a circular dependency. 

With @praisethemoon 's change, we can access HTML element from the note data happily like this.

`osmd.cursor.iterator.currentVoiceEntries[0].notes[0].graphicalNote.vfnote[0].attrs.el`

If we dig the code deeper, I think there is no reason for `AJAX.ts` existing in this library. 😄

ice6:
@sschmidTU 

After rethinking, maybe we just need to add a `currentGraphicalNotes` attribute to `osmd.cursor.iterator`.

This can solve many ui interaction problems.

matt-uib:
@ice6 thanks for your comments here also.
Our idea is to keep the general objectmodel built up from musicxml and also the cursor clean from the final graphical implementations. An Idea is for example: a musicxml audio player without even rendering any notes.
Our solution would be the dict described in #660 . Would you see any problems with this approach?

ice6:
After more digging, I found that `osmd.cursor.iterator.currentMeasure.verticalMeasureList[0].staffEntries[0].graphicalVoiceEntries[0].notes[0].vfnote[0].attrs.el`, this already can solve many ui interaction requirements.

@matt-uib it is ok to use dictionary to solve the problem. and the final api is clean. If it is already implemented, I would not discuss here. :P

sschmidTU:
> If we dig the code deeper, I think there is no reason for `AJAX.ts` existing in this library. 😄

@ice6  AJAX is used in OpenSheetMusicDisplay.ts to retrieve a MusicXML file asynchronously from a website.

ice6:
> > If we dig the code deeper, I think there is no reason for `AJAX.ts` existing in this library. 😄
> 
> @ice6 AJAX is used in OpenSheetMusicDisplay.ts:load() to retrieve a MusicXML file asynchronously from a website.

@sschmidTU sure, I know. just from the architecture perspective.

below is the code I use to retrieve file by url, it provides more flexibility some time.

```
let res = await fetch(url)
let content = await res.text()
```

sschmidTU:
There is now the static `GraphicalNote.FromNote(note: Note, rules: EngravingRules)`,
so you can easily get the GraphicalNote like this:
`const gNote = GraphicalNote.FromNote(note, osmd.rules);`

See #660.

sschmidTU:
There is an even easier method now:
`const gNote = osmd.rules.GNote(note)`
will give you the GraphicalNote from a Note reference.

PR #690, opened 2020-03-11
Simplify Blob polyfilling

Richienb:


sschmidTU:
Thanks for the simplification!
I'd like to not remove the explanatory comment, but i'll merge this and readd it, so you get the contribution.

PR #754, opened 2020-05-27
Optimize staff lines distance using SkyBottomline intersections

sschmidTU:
Optimize vertical distance between staff lines.

This lets staff lines be closer in distance by checking every point in a staffline's skyline for collisions with the skyline of the next staffline. This way, even if the lowest bottomline value of a staffline overlaps the highest skyline value of the next staffline, they can still be moved closer beyond this previous limitation, if the skylines don't cross at a specific location. [sschmidTU]

Added new parameter for customizing vertical layout.

implementation by @buenaflor , thank you very much!

I will now check these changes with [visual regression tests etc.](https://github.com/opensheetmusicdisplay/opensheetmusicdisplay/wiki/Testing), looks good so far!

### DIFF
--- a/src/MusicalScore/Graphical/DrawingParameters.ts
+++ b/src/MusicalScore/Graphical/DrawingParameters.ts
@@ -122,10 +122,21 @@ export class DrawingParameters {
     }
 
     public setForCompactTightMode(): void {
-        this.rules.CompactMode = true;
-        this.DrawCredits = false;
+        this.setForCompactMode(); // also sets CompactMode = true
         this.DrawPartNames = false;
-        this.drawHiddenNotes = false;
+
+        // tight rendering mode, lower margins and safety distances between systems, staffs etc. may cause overlap.
+        // these options can afterwards be finetuned by setting osmd.rules.BetweenStaffDistance for example
+        this.rules.MinimumSkyBottomLineDistance = 1.0; // default 1.0. this can cause collisions with slurs and dynamics sometimes
+        // note that this.rules === osmd.rules, since it's passed as a reference
+
+        this.rules.BetweenStaffDistance = 2.5;
+        this.rules.StaffDistance = 3.5;
+        this.rules.MinimumDistanceBetweenSystems = 1;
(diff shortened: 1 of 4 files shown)


sschmidTU:
@buenaflor i'm not getting any visual changes in the visual regression test, even at threshold 0.0000001 (10^-7). That's a bit unexpected, shouldn't this affect some of our samples? Do you have a good musicxml sample to illustrate this change?

matt-uib:
@sschmidTU that can really not be - we saw very nicely how it gives a different y-layout. Are you sure you had this changes active while testing? Maybe you tested not with the right repo?
You can also play with the parameters in rules to see how it behaves.

sschmidTU:
yes, i also had that suspicion, i'll run the tests again.

sschmidTU:
ok, tests show lots of changes now seems like i accidentally ran generate:blessed on the wrong branch, sorry ^^

sschmidTU:
Looks great! Love the change, scores overall look so much more compact and well layouted now!

One slight downside, slurs can overlap other elements a bit more commonly now:
Mozart - An Chloe, bar 10:
image

However, this is rather rare (doesn't happen in any other of our demo samples), and not too bad visually.

-----
But how about we set the default minimumSkyBottomLine distance to 2.0 instead of 1.0? Then it doesn't happen in this sample, and the scores look a bit less compact overall.
We can set it to 1.0 in compact mode instead.
Here's 2.0:
image

With the other samples as well, i think this looks slightly better, and still saves quite some space and makes the padding between staff lines look much less stretched.

----

There is one other issue this change does make more visible:
dynamics expressions like dim. or crescendo overlap the next staff line even more now.

Clementi sonatina op. 63 No 1 Part 2:
image

An die Geliebte, bar 10:
image

same, visual diff:
image

Of course, that's a bug that was already there and needs its own fix, but it's more visible after these changes.

sschmidTU:
MinimumSkyBottomLineDistance 2.0:
image

MinimumSkyBottomLineDistance 1.0:
image

sschmidTU:
Merging once builds are green. (already passed on Travis)

PR #756, opened 2020-05-28
Different Staffline Count/Percussion on Single Line

fredmeister77:
This adds an option to automatically render percussion staff's on a single percussion line, and contains the basic note positioning for that, as well as basic reading and handling off different staff line values specified via XML.

As might be expected, the visual regression tests were affected. Seems to be regarding spacing only, but might be good to double-check:
[visual_test_diff.zip](https://github.com/opensheetmusicdisplay/opensheetmusicdisplay/files/4692585/visual_test_diff.zip)

Karma tests pass:
[test_results.txt](https://github.com/opensheetmusicdisplay/opensheetmusicdisplay/files/4692586/test_results.txt)


sschmidTU:
Ah, did you do the visual regression tests with the latest develop? Then the diffs are probably because of the skyline changes I merged today, which make the score more compact (less vertical padding).
I'll merge develop into this feature branch locally and do the regression tests again compared to not merging develop, that should make the visual changes of this addition clear.

sschmidTU:
Other than that, this looks great! Thanks a lot!

fredmeister77:
Oh! That would explain it then. Thanks!

sschmidTU:
new diff with merged develop, only 4 changes, most of them very good,
just one little problem with OSMD Function Test Drumset, where it changes the placement of a few notes, which it shouldn't do:
image

sschmidTU:
But there are some nice rest positioning improvements,
like here in Mozart - An Chloe:
image

though i wonder why it's not the same for the lower staffline, which still has rest collisions.

fredmeister77:
@sschmidTU, Just committed corrections. Let me know if there are any issues/concerns. Thanks!

sschmidTU:
Looks great now, thanks! Merging.

PR #761, opened 2020-06-02
Fix skyline not being calculated for slurs properly

fredmeister77:
Test results:
[test_results.txt](https://github.com/opensheetmusicdisplay/opensheetmusicdisplay/files/4717967/test_results.txt)

Visual Results:
[results.txt](https://github.com/opensheetmusicdisplay/opensheetmusicdisplay/files/4717954/results.txt)
Differences seem to be related to spacing based on incorrect skyline, with current being more compact.

### DIFF
--- a/src/MusicalScore/Graphical/EngravingRules.ts
+++ b/src/MusicalScore/Graphical/EngravingRules.ts
@@ -409,7 +409,7 @@ export class EngravingRules {
 
         // Line Widths
         this.minimumStaffLineDistance = 4.0;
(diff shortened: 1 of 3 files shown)


sschmidTU:
Looks great, thanks! Will merge.

I'm just taking another look at the default `EngravingRules.minimumSkyBottomLineDistance` value, i think we can set it back to 1.0 now.
In #754, 1.0 was too compact because of slur collisions, so i increased it to 2.0.

But since we avoid slur collisions now, 2.0 is too expansive now i think, e.g. in Mozart - An Chloe:
image

1.0 looks fine:
image

PR #763, opened 2020-06-03
Fix for diminuendo wedges

fredmeister77:
diminuendo lines direction needed reversed for skyline/bottom line calculation. 
These are now calculated correctly in OSMD test files:
wedge-fix
wedge-fix2


sschmidTU:
Great!
But wouldn't it be even nicer if we supply a direction argument (maybe we call it WedgeDirection) to the `updateSkylineWithWedge()` function instead of reversing start and end point?
It's a minor refactor and we probably won't need to touch this again soon, but it both makes calling the function slimmer and the function itself more consistent, because it's more logical to have the start point of a wedge be the left-most point, which we confuse by reversing the arguments.

PR #765, opened 2020-06-03
Fix: Implemented placing the slurs on the side of the noteheads 

matt-uib:
Implemented placing the slurs on the side of the noteheads if start- and end-note have the same stemdirection.
Please check for slur changes in the visuals.


sschmidTU:
Running visual regression tests now.

sschmidTU:
Looks good! Will merge.

The regression tests show many slur changes of course. Most of them look better now.
There is a minority of slur placements which looked a bit better before, some slight collisions now, but overall the improvements far outweigh, and it would be difficult to find a general rule that also covers these cases anyways.

PR #766, opened 2020-06-03
Fix: added rendering uncategorized/unknown expressions

matt-uib:
added  missing implementation for uncategorized/unknown (no tempo or dynamic, ..) expressions to get rendered.
Please check the visuals - now there might be some expressions becoming visible..

### DIFF
--- a/src/MusicalScore/Graphical/AbstractGraphicalExpression.ts
+++ b/src/MusicalScore/Graphical/AbstractGraphicalExpression.ts
@@ -13,7 +13,7 @@ export abstract class AbstractGraphicalExpression extends GraphicalObject {
     /** EngravingRules for positioning */
     protected rules: EngravingRules;
 
-    constructor(parentStaffline: StaffLine, expression: AbstractExpression) {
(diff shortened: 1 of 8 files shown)


sschmidTU:
Great addition!

Ran the visual regression tests on this branch against this branch + merged develop.
positives:
* newly shows "espressivo" and "Ausdrucksvoll" in An die Ferne Geliebte measure 9, expressions in other samples
* fixes expression y placement in clementi 36 no 1 pt 2, not sure why. maybe because the new expression now got a proper y placement, then it's properly aligned.
(see branch fix/expression-collisions-skyline-y-shift)

newly unveiled issues (need to be fixed separately):
* expression overlap with metronome mark in Debussy Mandoline, clementi 36 no 3 pt 1 - need to avoid overlap between metronome mark and expressions
* two unnecessary expressions "Ornaments" in OSMD Function Test - Ornaments
(extra Ornaments label in OSMD Function Test - All, but doesn't look bad)
* default text label in OSMD Function Test - Expressions, should be removed from the XML

PR #767, opened 2020-06-03
fix: removed doublettes check for dynamic expressions

matt-uib:
...as the dynamic expression might make sense for e.g. a repetition start, where another dynamic was used at the end of the repetition.

We could implement this check as a afterSheetReadingModule when we know all Repetitions and Expressions..


sschmidTU:
Looks good! Merging.

Interestingly, while there were no regressions, there are a few dynamics that weren't rendered before but are now added, like a piano (in 3 voices each) in Mozart Clarinet Quintet bar 13.

PR #772, opened 2020-06-04
Feat: optimized system distance using sky- and bottomline values 

matt-uib:
same idea as in feat/OptimizeStafflineDistance.
Added comparing always with a certain x-range (-/+ 2units) to avoid too close objects in x-direction.
Please again check the visuals.


sschmidTU:
looks great! Merging.

Lots of diffs in the visual regression tests, of course, in most cases it gives the stafflines more space ("Luft"), but the page as a whole in most cases doesn't get longer, which is nice. In some cases the pages are very slightly longer now.

Of course, you can lower the `EngravingRules.MinSkyBottomDistBetweenSystems` value to make the score more compact, default is 5.0, in the CompactTight mode it's 2.0:
`osmd.setOptions({drawingParameters: "compacttight"})`

example:
image

sschmidTU:
This is basically the analogue of #754 for system distance. (#754 optimized distance between stafflines)

PR #774, opened 2020-06-05
fix: fixed breaking the system if the last measure has an endline 

matt-uib:
fixed breaking the system if the last measure has an endline (thin-bold line)

System stretching seems to not work still for these ending systems...

### DIFF
--- a/src/MusicalScore/Graphical/MusicSystemBuilder.ts
+++ b/src/MusicalScore/Graphical/MusicSystemBuilder.ts
@@ -54,7 +54,6 @@ export class MusicSystemBuilder {
     }
 
     public buildMusicSystems(): MusicSystem[] {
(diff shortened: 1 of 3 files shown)


sschmidTU:
Visual regression tests are fine, new line breaks in these samples:

OSMD_function_test_auto-custom-coloring-entchen.musicxml_1 36.1018
OSMD_function_test_bar_lines.musicxml_1 1.15003
OSMD_function_test_all.xml_1 0.901627
JosephHaydn_ConcertanteCello.xml_1 0.000824108 (11 systems before end)

the line break in the function test bar lines is a bit weird, maybe we don't want to line break there (MuseScore doesn't), but this is a non-sensical test sample anyways, so no big deal.
Maybe we can add an option to not automatically line break in the future.

PR #775, opened 2020-06-05
Calculate sky and bottom lines for octave marking

fredmeister77:
Adds sky and bottom line calculation for ottava lines. Additionally corrects issues with the positioning of them (avoiding collisions).

Need to correct issue with them not being rendered for middle stafflines (only renders for start and end currently)


sschmidTU:
Looks great!
Visual regression tests show 2 changes in Debussy Mandoline (negligible) and in OSMD Function Test Expressions, where the tempo expression now respects the skyline and is placed above the octave shift, which is nice.
image

PR #779, opened 2020-06-08
Fix measure number skyline offset

fredmeister77:
The staffline offset was only being taken into account for the first staffline. Removed this check.

The visual diff seems to be only related to the measure numbers and some spacing as a result of that:
[visual_diff.zip](https://github.com/opensheetmusicdisplay/opensheetmusicdisplay/files/4745793/visual_diff.zip)

Karma results:
[test_results.txt](https://github.com/opensheetmusicdisplay/opensheetmusicdisplay/files/4745794/test_results.txt)


sschmidTU:
Looks great! Merging.

A lot of measure number labels are better placed now (see tests).

PR #801, opened 2020-06-19
feat(label): Add support for multiline labels

fredmeister77:
detect line feed/CR, create lines
render lines with necessary spacing (composer, lyricist, title).

Visual diff results:
https://drive.google.com/file/d/1NOGbC5TZx28XJKwn5gNO_7ZKr9eEBbGF/view?usp=sharing

All seems related to the composer/lyricist spacing at the time. Seems expected?

Test Results:
[test_results.txt](https://github.com/opensheetmusicdisplay/opensheetmusicdisplay/files/4804544/test_results.txt)


src/MusicalScore/Graphical/MusicSystemBuilder.ts
```diff
@@ -1024,27 +1024,50 @@ export class MusicSystemBuilder {
...
+                    const totalLyricist: number = lyricistLineCount * lyricistFontHeight;
+
+                    if (totalLyricist > maxHeight) {
+                        maxLineCount = lyricistLineCount;
+                        maxFontHeight = lyricistFontHeight;
+                        maxHeight = totalLyricist;
+                    }
+
+                    currentYPosition += this.rules.TitleTopDistance + this.rules.SheetTitleHeight +
+                                            this.rules.TitleBottomDistance + maxHeight;
```
matt-uib, line 1048: here maxHeight shall not be added.
This actually increases the distance of the title to the notes which can be seen in the visual diffs in all cases.

matt-uib (commented):
There should not be any difference in the visual diffs if there is no multi label present.
Found the reason -> comment.
Also i saw that the text is not properly aligned in the multiLine test sheet (e.g. composer shall be right aligned) The text-lines have not the same right edge as i t should be.

fredmeister77:
@matt-uib, Thanks. Seems the x offset problem is due to whitespace in the XML when adding the newlines. 

Is there any scenario where we would want to render the whitespace on labels? Would it be fine to just trim() the text lines?

matt-uib:
@fredmeister77 yes good idea: trim the text of every line in a multiline label. Actually also on single labels -> all labels. Only problem might be some people use this to position their text by adding spaces - should not been done, but is done... let's see if it makes it worse somewhere.

fredmeister77:
Updated visual regression test:
[visual_results.zip](https://github.com/opensheetmusicdisplay/opensheetmusicdisplay/files/4816120/visual_results.zip)

Interestingly the break tests seem to be affected, not sure why.

matt-uib:
@fredmeister77 did you try the visuals again here or found the reason why there are other changes happening? Also in the Schubert piece the abbreviations are wrong before and after it seems...

fredmeister77:
I haven't rerun fresh visual tests on this branch yet. Doing that this morning

fredmeister77:
Corrected issue with system breaks, still trying to determine why instrument label position is affected on Schubert:
[visual_results.zip](https://github.com/opensheetmusicdisplay/opensheetmusicdisplay/files/4828743/visual_results.zip)

sschmidTU:
@fredmeister77 the Schubert label position doesn't matter, i guess, because in that sample the instrument label position is bugged in any case, for some reason. That's why we disable partnames in the index.js for the Schubert sample.
I forgot about this myself and wondered why the generateImages script had these weird labels while the demo didn't.

sschmidTU:
@fredmeister77 but i do wonder about the other visual diff, in the Function Test Labels, the top row of the title label is cut off in the png. Seems like it doesn't fit the container. Maybe we have to update the bounding box or extend the page height or something.

image

sschmidTU:
Also, the title labels are unnecessarily high, contributing to pushing them out of the top frame.
Maybe we can do local skyline checks there, instead of having the composer and lyricist labels and the title labels pushing each other in y-direction in any case.

fredmeister77:
@sschmidTU ,
Ah that's right. I was thinking the bounding box was being calculated based on the new lines, but maybe not. Will investigate. Thanks

fredmeister77:
Still needs some tweaks I think but close. Will finish out tomorrow

fredmeister77:
I think this is okay now.

1. I've commented out the y position offset code. It would only be necessary if the composer or lyricist lines were underneath the title, which seems pretty rare. Think we would want to use the skyline to discover this anyway, and not do it in MusicSystemBuilder.
2. Bounding box is fixing. Reverted to constructing the text blocks from top down vs. bottom up. This resolved the cutoff issue.
3. Positioning of lyricist + composer uses skyline now instead (if more than one line). it also updates the skyline. This could maybe use more tweaking but seems to work.

Now the only diff left is the labels test (as expected) since develop is merged in:
[diff.zip](https://github.com/opensheetmusicdisplay/opensheetmusicdisplay/files/4836995/diff.zip)

Here's the Karma tests:
[test_results.txt](https://github.com/opensheetmusicdisplay/opensheetmusicdisplay/files/4836996/test_results.txt)


Let me know if there are any issues or concerns on this. I can update further as needed.

sschmidTU:
Looks great, thanks! Merging.

matt-uib:
I have just seen in #826 that the alignment of right aligned labels (like the Composer) did not work so far.
I implemented this now and pushed it on develop.

sschmidTU:
great!

PR #807, opened 2020-06-23
Add option for untracked local env config file

fredmeister77:
@sschmidTU,

I think this is a good option to have, let me know if you think there are any issues with having this included in the codebase.

Specifically, I needed to add the 'allowedHosts' option with a custom entry for my local environment. I can imagine other local-specific environment options could exist for other developers as well.


sschmidTU:
Sure, fine with me! :)

PR #808, opened 2020-06-24
Enable print-object attribute for part names

fredmeister77:
Enables print-object attribute for part names. If set to no, the part name (and abbrev.) will not be rendered for that part.

Visual Test Results:
[visual_results.zip](https://github.com/opensheetmusicdisplay/opensheetmusicdisplay/files/4822507/visual_results.zip)

The differences here are due to the attribute being present/set in the XML sheets in question.

Karma test results:
[test_results.txt](https://github.com/opensheetmusicdisplay/opensheetmusicdisplay/files/4822508/test_results.txt)


matt-uib:
@fredmeister77 here again the Schubert piece seems to make problems:
Is there really only on the second piano staff a "print-object=no"?
And the abbreviated names are not taken into account by the system layout - they are inside the system start measures. -> this seems to be a bug that already was present and most likely has nothing to do with your changes.

fredmeister77:
Yeah, I had to check last night to see why, but that is how the XML is. print-object="no" only on second Piano:
schubert-xml

Also noticed the issue with the system layout, but that it was already present. Figured I would create a separate issue for that

sschmidTU:
I just merged the feat/label/print branch again, no code changes, because in SourceTree it didn't show the branch as merged into develop, which was confusing, because a specific commit was merged in this PR, instead of the branch.

sschmidTU:
By the way, the Actor Prelude sample is 1.2MB, so it's not ideal that we copied it as a test sample, but maybe it's useful.

fredmeister77:
@sschmidTU, Yeah that's true. I can come up with a better test file if preferred: The only reason I used that was it being readily available and the large number of part names + part abbreviations.

Also this: 

> I just merged the feat/label/print branch again, no code changes, because in SourceTree it didn't show the branch as merged into develop, which was confusing, because a specific commit was merged in this PR, instead of the branch.

Did I do something incorrectly with the commit here? I'm not sure how I would have accidentally committed to PR instead of branch, very strange.

sschmidTU:
No problem with the commit, it was just how the PR was merged, it seems like a specific commit was merged instead of the branch, so SourceTree showed the branch as not merged.

PR #846, opened 2020-08-02
fix(tie messup): Ties are now rendered correctly again

### LINKED ISSUE #844: Bug about the display of ties of chords
I found that the ties of the chords with multiple sustains has wrong displays , just like the following pic shows:
image
[tie_test.musicxml.zip](https://github.com/opensheetmusicdisplay/opensheetmusicdisplay/files/4999277/tie_test.musicxml.zip)

bneumann:
This should fix #844


bneumann:
Wow! Sorry for the merge madness. Out of a habit I based my fix on master instead of develop. I noticed it AFTER the PR. So that's why there are so many merges. The fix itself is rather small

sschmidTU:
Looks good! Merging.

Visual regression passes. Test sample see #844.

Also, no problem about the merges, I'll just squash and merge.

PR #847, opened 2020-08-02
feat(stretch last line): option to stretch last system line

### LINKED ISSUE #842: Music sheet doesn't end at the right end of the line
image
↑ OSMD Demo

Music sheet must end at the right end of the line (in common)
but in OSMD, if there's not enough notes or lyrics in the system, 
sheet doesn't end at the right end of the line

using `newSystemFromXML: true` doesn't work when there's no measure anymore in the next line(system)

If making it work even when it's last, will save me a lot!
Thanks :)

bneumann:
This should fix #842


sschmidTU:
Looks good! Merging.

Visual regression tests pass.

Note that this can also be fixed by a JustifiedMusicSystemBuilder. (justification = blocksatz)

PR #848, opened 2020-08-02
fix(metronome marks): Metronome marks were only drawn in first line

### LINKED ISSUE #804: Handling tempo marking changes
Sheet music with multiple tempo markings lose their coordinates on the score and end up rendering overlapped in the upper-left corner. Musescore handles it fine so I don't think there is an error in the musicxml.

Here is an example score I tested on the OSMD demo.
[A_New_Beginning.musicxml.zip](https://github.com/opensheetmusicdisplay/opensheetmusicdisplay/files/4814162/A_New_Beginning.musicxml.zip)

bneumann:
Fix #804 
- Getting the staffline/measure now from the expression and setting the mark to the correct position
- Switch first line and other lines x-shift


sschmidTU:
Fixed a little bug in pieces starting with pickup measures (Auftakt), where the metronome mark was given in measure number 0, and the measure index wasn't math.maxed with 0, like in the Mozart string quartet sample.

sschmidTU:
also added OSMD Function Test - Metronome Marks to samples.

PR #852, opened 2020-08-02
fix(rest note y positions): Rest notes respect the display-step and display-octave tags now

### LINKED ISSUE #759: Rest positioning: take display-step from XML?
I noticed that there is a rest / beam overlap in the bass clef of measure 13 of An die ferne Geliebte in the OSMD Demo.

image

This musicxml for this has a display-step setting to try to position the rest in the same place as the notes before and after it.

```
<note default-x="76.94" default-y="-236.96">
        <pitch>
          <step>C</step>
          <octave>4</octave>
          </pitch>
        <duration>2</duration>
        <voice>5</voice>
        <type>eighth</type>
        <stem>down</stem>
        <staff>2</staff>
        <beam number="1">begin</beam>
        </note>
      <note>
        <rest>
          <display-step>C</display-step>
          <display-octave>4</display-octave>
          </rest>
        <duration>1</duration>
        <voice>5</voice>
        <type>16th</type>
        <staff>2</staff>
        </note>
```

We seem to be ignoring display-step.

When I load the musicxml into musescore, the rest is displayed inline with the C,

image

However, interestingly, when I load it into Sibelius, it ignores "display-step" but alters the stem length to avoid the overlap.

image

Have we looked at this issue before?
(I searched and could not find a discussion of it)

bneumann:
Fix #759 
I added the xml node the the Note.ts class. Now it get's the information if there is a repositioning of the rest in y direction. @sschmidTU I stumbled upon an old FIXME where you noticed, that we shift the octave by +3 in case of notes. It seems to depend on the clefs in case of rests also. Any idea what that might be? Some vexflow magic?

### DIFF
--- a/src/MusicalScore/Graphical/VexFlow/VexFlowConverter.ts
+++ b/src/MusicalScore/Graphical/VexFlow/VexFlowConverter.ts
@@ -125,6 +125,9 @@ export class VexFlowConverter {
             if (note.Clef().ClefType === ClefEnum.F) {
                 fixmeOffset = 2;
             }
+            if (note.Clef().ClefType === ClefEnum.C) {
+                fixmeOffset = 2;
+            }
         }
         const octave: number = pitch.Octave - note.Clef().OctaveOffset + fixmeOffset;
         const notehead: Notehead = note.sourceNote.Notehead;
--- a/src/MusicalScore/Graphical/VexFlow/VexFlowConverter.ts
+++ b/src/MusicalScore/Graphical/VexFlow/VexFlowConverter.ts
@@ -117,8 +117,16 @@ export class VexFlowConverter {
     public static pitch(note: VexFlowGraphicalNote, pitch: Pitch): [string, string, ClefInstruction] {
         const fund: string = NoteEnum[pitch.FundamentalNote].toLowerCase();
         const acc: string = Pitch.accidentalVexflow(pitch.Accidental);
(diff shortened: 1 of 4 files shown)


sschmidTU:
@bneumann yeah, i'm not sure why this fixme offset is necessary, but it gets the right results.
Added a fix for c clef octave.

For the positioning of rest notes, the display step will be the bottom of the rest note.
Maybe we want to center the position instead?

This rest is given as display-step F3 (one line above the top line on g-clef):
image

The MusicXML spec doesn't tell what the display-step means exactly, whether it's the bottom or center of the rest note.
https://usermanuals.musicxml.com/MusicXML/Content/EL-MusicXML-display-step.htm

sschmidTU:
Visual regression:

Cornelius_P_Christbaum_Opus_8_1_1865.mxl_1 1.25178
Beethoven_AnDieFerneGeliebte.xml_skybottomline_1 0.604729
Mozart_DasVeilchen.xml_1 0.0110626
JosephHaydn_ConcertanteCello.xml_1 0.00147506
Beethoven_AnDieFerneGeliebte.xml_1 0.000602635
Dichterliebe01.xml_1 0.000568139
TelemannWV40.102_Sonate-Nr.1.2-Allegro-F-Dur.xml_1 0.000433111
Debussy_Mandoline.xml_1 0.00033919

For almost all these cases, rest positioning is much improved.

Great work @bneumann ! Merging.

PR #859, opened 2020-08-08
Adding Samples for Guitar Tablature Effects

momolarson:
I've added samples to the demo app that demonstrate Slides, Pulls, Hammer Ons, Pull Offs, Vibratos, as well as more complex and complete cases.


sschmidTU:
Merged, with a small follow-up polish commit.

These samples are great, thanks a lot! Even more for the implementation of the effects of course!

momolarson:
thanks Simon, that means a lot to me

PR #861, opened 2020-08-11
Auto-generate multirest from subsequent rest-only measures

fredmeister77:
Test results:
[test_results.txt](https://github.com/opensheetmusicdisplay/opensheetmusicdisplay/files/5056980/test_results.txt)

Visual diff results:
[visual_results.txt](https://github.com/opensheetmusicdisplay/opensheetmusicdisplay/files/5056981/visual_results.txt)

Added two test files.

One question:
This setting is set to on by default; Perhaps we should default to off? Not sure.

### DIFF
--- a/demo/index.js
+++ b/demo/index.js
@@ -46,6 +46,8 @@ import * as svg2pdf from '../node_modules/svg2pdf.js/dist/svg2pdf.min';
             "OSMD Function Test - Tremolo": "OSMD_Function_Test_Tremolo_2bars.musicxml",
             "OSMD Function Test - Labels": "OSMD_Function_Test_Labels.musicxml",
             "OSMD Function Test - High Slur Test": "Slurtest_highNotes.musicxml",
+            "OSMD Function Test - Auto Multirest Measures Single Staff": "Test_Auto_Multirest_1.musicxml",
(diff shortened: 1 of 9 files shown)


sschmidTU:
Great work! Merged.

Yes, default on or off is debatable, I guess score creators should be deliberate about putting multirest measures into the score (which will be exported to XML).
On the other hand, there are very few scores with multiple rest measures in a row that are deliberate and shouldn't be contracted, most of those will be orchestral or ensemble (marching band) pieces where multiple rest measures are great to have.
I guess the user can always disable this setting by the option if they want to, but having it on by default will benefit a lot of people who haven't seen the option yet.

sschmidTU:
There was an issue with multirests being applied to all measures that have no rests, but have expressions or clef instructions.
Here in the Haydn sample bar 128 it removed the "Solo":
image

without autoMultiRest, and with the latest fix:
image

it's also strange that the rest measure numbers don't add up (126 -> 2 rest measures -> 133),
maybe that's another issue that needs looking into at one point,
maybe it's just because these multirests were written into the XML.

sschmidTU:
it also checks for a ClefInstruction at the end now, so that they aren't swallowed by a MultiRestMeasure.
image

fredmeister77:
Nice, thanks! Didn't think of those

Issue #541, opened 2019-07-05
Measure number regions that don't start at 1, and drawMeasureNumbers option


finneganwl:
I have run into 2 issues with measure numbers.

**Issue 1**
I tried to load a musicXml file where the first measure number is not 1, e.g. 

`<part id="P1">
    <measure number="100" ...`

and the measure numbers displayed still assume that the first measure is numbered 1. The rendering is ignoring the `number` property.

**Issue 2**
When loading a small score of only 1-2 measures, measure numbers are not displayed at all. It looks like by default, every other measure number is displayed starting with the 3rd measure. Should there exist a drawMeasureNumbers option in IOSMDOptions or another way to get finer grained control over how often to display measure numbers?

bneumann:
Thanks for submitting this. I think I stumbled upon this when implementing the drawing method.
My proposal is to use more options as you say. Something like ```UseAutomaticMeasureNumbers``` (default: true) and ```DrawEveryNthMeasureNumber``` (default: 5).
We need the earlier because we now have the possibility to draw subsets of pieces where you don't want the old measure numbers to show.

matt-uib:
@finneganwl yes currently we show the number on every second measure within every system, starting at each system start measure (without numbering the first measure of the first system) - if i'm correct.
Yes, makes total sense to have options for that for being able to customize this.
As this is rather a "nice to have" feature from what currently is pending, we would be very happy if someone of the community could take this over.
Would you like to maybe contribute on that?

sschmidTU:
FYI: there are `drawMeasureNumbers` and `measureNumberInterval` options now in OSMDOptions.
But still, further `UseAutomaticMeasureNumbers` or something like `useXMLNumbers` options would be nice to have.

sschmidTU:
Scores with measure numbers not starting at 1 should print the measure numbers given in XML now:
![image](https://user-images.githubusercontent.com/33069673/92279618-f9975a00-eef7-11ea-88fa-619de4d47141.png)
[measure_numbers_xml_starting_at_3_with_multirest.musicxml.zip](https://github.com/opensheetmusicdisplay/opensheetmusicdisplay/files/5176697/measure_numbers_xml_starting_at_3_with_multirest.musicxml.zip)

This is configured by `osmd.EngravingRules.UseXMLMeasureNumbers`, which defaults to true.

You can have basically arbitrary measure numbers as given in the XML now, even nonlinear and repeating:
[...]

PR #879, opened 2020-08-29
Add option to read measure numbers from musicxml measure attribute

jmorag:
Implements https://github.com/opensheetmusicdisplay/opensheetmusicdisplay/issues/541

### DIFF
--- a/demo/index.js
+++ b/demo/index.js
@@ -438,6 +438,7 @@ import * as svg2pdf from '../node_modules/svg2pdf.js/dist/svg2pdf.min';
 
             //drawMeasureNumbers: false, // disable drawing measure numbers
             //measureNumberInterval: 4, // draw measure numbers only every 4 bars (and at the beginning of a new system)
+            useXMLMeasureNumbers: true, // read measure numbers from xml
 
             // coloring options
             coloringEnabled: true,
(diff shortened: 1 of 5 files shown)


sschmidTU:
Good idea, thanks for the implementation!

However, we currently use the measure number for some array indexes and comparisons. The easiest option is probably to save the printed measure number as an extra variable.
It makes sense to save both anyways, because you may want to start your printed measure numbers at something other than 1, but logically a sheet always has a linear sequence of measures, where we can use 0 as an array index for the first measure.

Adding the printed measure number as an extra variable in SourceMeasure is easy, i'll do that, after merging this to a feature branch.

sschmidTU:
@jmorag by the way, do you have a nice small MusicXML sample with modified measure numbers?

sschmidTU:
This is now merged to develop, see #541.

jmorag:
> @jmorag by the way, do you have a nice small MusicXML sample with modified measure numbers you can share?

Sure, here it is. 
[Prokofiev_violin_concerto_No_2_excerpt.txt](https://github.com/opensheetmusicdisplay/opensheetmusicdisplay/files/5177006/Prokofiev_violin_concerto_No_2_excerpt.txt)

Thanks for improving my implementation. The non-linear numbering is pretty cool. I just hope I never encounter any music that needs it :-D

sschmidTU:
Thanks for the sample! Looks good to me.
image

Yes, i'm not sure there is a valid use case for nonlinear measure numbering, but supporting it is easier than not to ;)

PR #880, opened 2020-09-03
Certain xml docs cause hang when loading

DLiblik:
The `greatestCommonDenominator` function has no guard against non-convergence due to floating-point errors and this causes loading of some musicxml docs to spin forever in the for loop.

Basically, in some cases `b` becomes very small (< `2.0e-14`) and never converges to 0.

### DIFF
--- a/src/Common/DataObjects/Fraction.ts
+++ b/src/Common/DataObjects/Fraction.ts
@@ -64,7 +64,7 @@ export class Fraction {
       return 1;
     }
 
-    while (b !== 0) {
+    while (Math.abs(b) > 0.0000001) {
       if (a > b) {
         a -= b;
       } else {


sschmidTU:
Great idea, thanks!
@DLiblik do you have a sample (musicxml) where this caused a hang? would be nice for testing.
Also, personally i'd write this as `1.0e-8` instead of `0.00000001`, for brevity and modification purposes. (but i can edit that in myself)

DLiblik:
Let me see if I can track it down.  It was a batch (I was doing some testing as we are evaluating using the tech in a project we are working on) so give me a few days to track it down.

All good on scientific notation (`1.0e-8`) - counting zeroes is problematic at best :bowtie:

sschmidTU:
We should probably also use `return Math.round(a)` instead of `return a`, so that we don't return e.g. 4.0000001 as the greatest common denominator. I'll add that now.

Issue #911, opened 2020-10-30
if the guitar tablature has a chord with a note on each string, OSMD won't display tab 


momolarson:
I created a simple tab in Muse Score and have attached it.  If you drop this file into the demo app, you get a lot of errors talking about svg heights being NaN and Infinity.  

I have included two examples in the zip:  

1> Full_Chord_Error.musicXML has the error
2> OLTB.xml - same chord, nothing on 6th string, works fine.

[chordError.zip](https://github.com/opensheetmusicdisplay/opensheetmusicdisplay/files/5462504/chordError.zip)

momolarson:
@sschmidTU - I have been looking into this and I am willing to fix, I just wanted to get this documented!  Also, maybe some sage guidance!  The error happens here:

https://github.com/opensheetmusicdisplay/opensheetmusicdisplay/blob/6b10b01c0a5cdd73087b09541b1aa89b2a9d5b6e/src/MusicalScore/Graphical/VexFlow/VexFlowStaffEntry.ts#L58

but I am having issues determining where those values are calculated.

momolarson:
After a little more digging, It looks like the error is occurring in BoundingBox: https://github.com/opensheetmusicdisplay/opensheetmusicdisplay/blob/6b10b01c0a5cdd73087b09541b1aa89b2a9d5b6e/src/MusicalScore/Graphical/BoundingBox.ts#L309

In the "final" call to calculateBoundingBox, the parent Bounding Box is getting the default values and then when looping through the children, the max function is called creating numbers like "5e-324" (Number.MAX_VALUE)

should these default to max width of the page? Thanks for any insight,

Gregg

sschmidTU:
Ah, interesting! Yes, we should probably initialize with max width(/height?) of the page instead of Number.MAX_VALUE.
But why is Number.MAX_VALUE `5e-324`? In my Firefox and Chrome it's `1.7976931348623157e+308`. The exponent should never be negative.
I'll see if initializing with max page width fixes it (though a bit later, if you could test and submit a PR it would be great too!)

momolarson:
Thanks @sschmidTU  - I appreciate that tip, and I will investigate why I am seeing such an odd MAX_VALUE, I am using Microsoft Edge (chrome), so perhaps it is something there.  I can test out the fix you mentioned too!  Hope all is well!

Gregg

sschmidTU:
[...]

PR #912, opened 2020-10-31
initial fix

momolarson:
First pass at fixing issue 911

### DIFF
--- a/src/MusicalScore/Graphical/SkyBottomLineCalculator.ts
+++ b/src/MusicalScore/Graphical/SkyBottomLineCalculator.ts
@@ -173,12 +173,13 @@ export class SkyBottomLineCalculator {
      */
     private findPreviousValidNumber(start: number, tSkyLine: number[]): number {
         for (let idx: number = start; idx >= 0; idx--) {
-            if (!isNaN(tSkyLine[idx])) {
+            if (!isNaN(tSkyLine[idx]) && tSkyLine[idx] !== undefined) {
                 return tSkyLine[idx];
             } else {
                 return 0;
             }
         }
+        return 0;
     }
 
     /**
--- a/src/MusicalScore/Graphical/SkyBottomLineCalculator.ts
+++ b/src/MusicalScore/Graphical/SkyBottomLineCalculator.ts
@@ -101,7 +101,12 @@ export class SkyBottomLineCalculator {
                         break;
                     }
                 }
+
+                if (tmpSkyLine[x] === undefined) {
+                    tmpSkyLine[x] = 0;
+                }
             }
+
             this.mSkyLine.push(...tmpSkyLine);
             this.mBottomLine.push(...tmpBottomLine);
 
--- a/src/MusicalScore/Graphical/SkyBottomLineCalculator.ts
+++ b/src/MusicalScore/Graphical/SkyBottomLineCalculator.ts
@@ -197,6 +197,7 @@ export class SkyBottomLineCalculator {
                 return tSkyLine[idx];
             }
         }
+        return 0;
     }
 
     /**
--- a/src/MusicalScore/Graphical/SkyBottomLineCalculator.ts
+++ b/src/MusicalScore/Graphical/SkyBottomLineCalculator.ts
@@ -102,8 +102,15 @@ export class SkyBottomLineCalculator {
                     }
                 }
 
-                if (tmpSkyLine[x] === undefined) {
-                    tmpSkyLine[x] = 0;
+               // if (tmpSkyLine[x] === undefined) {
+                 //   tmpSkyLine[x] = tmpSkyLine[x-1];
+               // }
+
+            }
+
+            for (let idx: number = 0; idx < tmpSkyLine.length; idx++) {
+                if (tmpSkyLine[idx] === undefined) {
+                    tmpSkyLine[idx] = (this.findPreviousValidNumber(idx, tmpSkyLine) + this.findNextValidNumber(idx, tmpSkyLine)) / 2; // tmpSkyLine[x-1];
                 }
             }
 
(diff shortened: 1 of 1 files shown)


sschmidTU:
did you forget to push a commit? because it seemed like you wanted to average the undefined skyline values between the previous and next value, as we said in #911.
But you don't necessarily need to write this yourself, i can also do it a bit later today.

sschmidTU:
Or maybe we just copy the previous value, because it would be a bit complicated to look ahead in the values, and this may be the better solution anyways

momolarson:
yeah, I was thinking the same.  I have both solutions here, but I agree with the copy previous value.  They are both in current PR.

sschmidTU:
Or rather, let's take the max value, because we could have something like [undefined, 95, 95], or [0,0,undefined,95].

src/MusicalScore/Graphical/SkyBottomLineCalculator.ts
```diff
@@ -152,6 +164,41 @@ export class SkyBottomLineCalculator {
...
+     */
+    private findPreviousValidNumber(start: number, tSkyLine: []): number {
+        for (let idx: number = start; idx >= 0; idx--) {
+            if (!isNaN(tSkyLine[idx])) {
+                return tSkyLine[idx];
+            } else {
+                return 0;
+            }
+        }
+    }
```
sschmidTU, line 182: we need to return 0 after the for loop as well, in case no value was not undefined

src/MusicalScore/Graphical/SkyBottomLineCalculator.ts
```diff
@@ -152,6 +164,41 @@ export class SkyBottomLineCalculator {
...
+    /**
+     * go backwards through the skyline array and find a number so that
+     * we can properly calculate the average
+     * @param start
+     * @param backend
+     * @param color
+     */
+    private findPreviousValidNumber(start: number, tSkyLine: []): number {
+        for (let idx: number = start; idx >= 0; idx--) {
+            if (!isNaN(tSkyLine[idx])) {
```
sschmidTU, line 176: this should be !isNan && !== undefined

src/MusicalScore/Graphical/SkyBottomLineCalculator.ts
```diff
@@ -152,6 +164,41 @@ export class SkyBottomLineCalculator {
...
+    private findNextValidNumber(start: number, tSkyLine: Array<number>): number {
+        if (start === tSkyLine.length) {
+            return tSkyLine[start];
+        }
+        for (let idx: number = start; idx < tSkyLine.length; idx++) {
+            if (!isNaN(tSkyLine[idx])) {
+                return tSkyLine[idx];
+            }
+        }
+    }
```
sschmidTU, line 200: here return 0 after the for loop as well

src/MusicalScore/Graphical/SkyBottomLineCalculator.ts
```diff
@@ -152,6 +164,41 @@ export class SkyBottomLineCalculator {
...
+    /**
+     * go forward through the skyline array and find a number so that
+     * we can properly calculate the average
+     * @param start
+     * @param backend
+     * @param color
+     */
+    private findNextValidNumber(start: number, tSkyLine: Array<number>): number {
+        if (start === tSkyLine.length) {
+            return tSkyLine[start];
```
sschmidTU, line 193: this won't work, tSkyLine[tSkyLine.length] is an index out of bounds error.

sschmidTU:
i made a few review comments, but i can simply merge this on a branch and polish the last few things myself if you want.

src/MusicalScore/Graphical/SkyBottomLineCalculator.ts
sschmidTU: oh, nevermind, isNaN(undefined) is true, so this works.

src/MusicalScore/Graphical/SkyBottomLineCalculator.ts
```diff
@@ -190,7 +190,7 @@ export class SkyBottomLineCalculator {
      */
     private findNextValidNumber(start: number, tSkyLine: Array<number>): number {
         if (start === tSkyLine.length) {
```
sschmidTU, line 192: we could also do >= to be safe on values > tSkyLine.length.

src/MusicalScore/Graphical/SkyBottomLineCalculator.ts
```diff
@@ -190,7 +190,7 @@ export class SkyBottomLineCalculator {
...
-            return tSkyLine[start];
+            return tSkyLine[start - 1];
```
sschmidTU, line 193: good!

src/MusicalScore/Graphical/SkyBottomLineCalculator.ts
```diff
@@ -101,7 +101,19 @@ export class SkyBottomLineCalculator {
...
+
+               // if (tmpSkyLine[x] === undefined) {
+                 //   tmpSkyLine[x] = tmpSkyLine[x-1];
+               // }
+
+            }
+
+            for (let idx: number = 0; idx < tmpSkyLine.length; idx++) {
+                if (tmpSkyLine[idx] === undefined) {
+                    tmpSkyLine[idx] = (this.findPreviousValidNumber(idx, tmpSkyLine) + this.findNextValidNumber(idx, tmpSkyLine)) / 2; // tmpSkyLine[x-1];
```
sschmidTU, line 113: let's do Math.max(previous, next) here instead of the average

src/MusicalScore/Graphical/SkyBottomLineCalculator.ts
```diff
@@ -152,6 +164,43 @@ export class SkyBottomLineCalculator {
...
+    /**
+     * go backwards through the skyline array and find a number so that
+     * we can properly calculate the average
+     * @param start
+     * @param backend
+     * @param color
+     */
+    private findPreviousValidNumber(start: number, tSkyLine: number[]): number {
+        for (let idx: number = start; idx >= 0; idx--) {
+            if (!isNaN(tSkyLine[idx]) && tSkyLine[idx] !== undefined) {
```
sschmidTU, line 176: we don't need the `&& tSkyLine[idx] !== undefined`, sorry, that was my mistake. isNaN() covers undefined.

src/MusicalScore/Graphical/SkyBottomLineCalculator.ts
```diff
@@ -152,6 +164,43 @@ export class SkyBottomLineCalculator {
...
+                return tSkyLine[idx];
+            } else {
```
sschmidTU, line 178: remove the else branch. this would always return 0 if the first previous value is undefined. but we want to keep looking until we find a better previous value.
we now return 0 after the for loop, that's why we don't need the else anymore.

sschmidTU: This is already fixed in findNextValidNumber, it looks like.

sschmidTU:
@momolarson great, this looks good to me now! Are you done with your changes? I'll just run the tests and merge then.

src/MusicalScore/Graphical/SkyBottomLineCalculator.ts
momolarson: all good!  I appreciate your review!

sschmidTU (approved):

momolarson:
> @momolarson great, this looks good to me now! Are you done with your changes? I'll just run the tests and merge then.

yeah, I believe I have addressed all of your issues.  thanks for the help.

sschmidTU:
Alright, thanks for the fix! Merging.

Issue #597, opened 2019-10-23
Chord Spacing


AlbertHart:
In this score, the Chord letters overlap.

Apparently the measure spacing does not take into account chord letter.

The Rendering scale does not seem to have any effect.

I suppose I could add some fake lyrics to make the measure larger.

Does anyone else have any ideas?

![image](https://user-images.githubusercontent.com/8965352/67360800-17e4a900-f524-11e9-9a80-fe7d25473756.png)

AlbertHart:
Adding a dummy lyric did work for now...

![image](https://user-images.githubusercontent.com/8965352/67398911-a3802900-f568-11e9-8634-9fe406e33e32.png)

sschmidTU:
Yes, the problem is that we don't have a chord symbol spacing method yet, as we have for lyrics. We rarely see chord symbol overlaps and so many long chord symbols in one measure, though it's of course a valid use case.
It would be great to port the system for lyrics spacing, which also enlargens the measure if there isn't enough space, and apply it to chord symbols.
Actually, as @matt-uib said, it would be great to apply this x-positioning system to all symbols that are on the same y-plane (need x layout), expressions too.

AlbertHart:
How do I get this labelled as a feature request (or bug)?

sschmidTU:
Best way is asking :) i applied the feature label.
probably not a bug, because chord spacing doesn't exist yet.

sschmidTU:
In #599 i wrote a few starting ideas about chord spacing.
We should space chords equally when there's a lot of space available,
and extend the measure to avoid collisions, like the existing lyrics spacing.

![image](https://user-images.githubusercontent.com/33069673/70249820-3e317180-177d-11ea-9fb5-8aee8931a443.png)
[chord_spacing_with_and_without_extending_measure.musicxml.zip](https://github.com/opensheetmusicdisplay/opensheetmusicdisplay/files/3927684/chord_spacing_with_and_without_extending_measure.musicxml.zip)

A simple method to space the symbols using the available space, without extending measures:
Check the remaining space for chords, which is the space until the next StaffEntry if there is one in the measure, or the end of the measure if there's none.
Then, space the chord symbols equidistant by the remaining space, minus the width of the chordsymbols, divided by number of chordsymbols.
This way, two chord symbols over a whole note are half the measure apart.
If there's not enough space (result << 1), use minimal spacing.

[...]

PR #928, opened 2021-01-08
Elongate measures for chords

hartman42:
Used the calculateMeasureWidthFromLyrics function in src/MusicalScore/Graphical/VexFlow/VexFlowMusicSheetCalculator.ts as a template for calculateMeasureWidthFromChords function.  

#597


demo/index.js
```diff
@@ -10,6 +10,7 @@ import * as svg2pdf from '../node_modules/svg2pdf.js/dist/svg2pdf.min';
     var openSheetMusicDisplay;
     var sampleFolder = process.env.STATIC_FILES_SUBFOLDER ? process.env.STATIC_FILES_SUBFOLDER + "/" : "",
         samples = {
+            "Chord Spacing Test":"ChordSpacingTest.musicxml",
```
bneumann, line 13: maybe move this one down to the test files so we have the Beethoven one as default

src/MusicalScore/Graphical/VexFlow/VexFlowMusicSheetCalculator.ts
```diff
@@ -409,6 +409,151 @@ export class VexFlowMusicSheetCalculator extends MusicSheetCalculator {
     return oldMinimumStaffEntriesWidth * elongationFactorForMeasureWidth;
   }
 
+  public calculateMeasureWidthFromChords(
+    measuresVertical: GraphicalMeasure[],
+    oldMinimumStaffEntriesWidth: number
+  ): number {
+    let elongationFactorForMeasureWidth: number = 1;
+
+    interface ChordEntryInfo {
```
bneumann, line 418: Was the definition of interfaces inside the class on purpose? I'd assume they should be either parallel or in a separate file.

src/MusicalScore/Graphical/VexFlow/VexFlowMusicSheetCalculator.ts
```diff
@@ -409,6 +409,151 @@ export class VexFlowMusicSheetCalculator extends MusicSheetCalculator {
...
+            // extend: graphicalChordContainer.LyricsEntry.extend,
+            chordXPosition: chordXPosition,
+            extend: false,
+            labelWidth: chordLabelWidth,
+            measureNumber: measure.MeasureNumber,
+            // sourceNoteDuration: graphicalChordContainer.LyricsEntry.Parent.Notes[0].Length,
+            text: graphicalChordContainer.GetGraphicalLabel.Label.text
+          };
+
+          // currentSpacingToMeasureEnd = maxXInMeasure - chordXPosition;
```
bneumann, line 539: Commented code

bneumann (approved):
Looks good. See my review notes as recommendation

matt-uib (commented):
Sorry i wanted to write already earlier..
Thanks very much for your solution to this.
There is a little problem i see when taking the "lyrics" method here:
For the lyrics we always measure the needed space up to the next note - that's typically ok, as every note has lyrics underneath.
This is not the case for chords, which are given not very often. And also the chord lasts musically typically for some notes to come.
So my point is, that we shall not use the collision with the next GraphicalStaffEntry, but only with the next GraphicalChordSymbolContainer.
The problem is actually well visible in the image you sent:
The chord C#m7(b5) leads to a rather big stretch of the whole measure, which would be less if we would do the collision only up to the next chord F#7(#9).
This would actually be generally the best way of colliding things of same kind.
So we should also apply this for the lyrics as well! (i had this already in mind for some time ...)
Same holds for dynamic expressions, ...
So it would be great to have here a more or less generic solution to collide a list of elements of same kind and this will put out a final stretch factor for the measure.
This can be done then for the chord containers, for the lyrics (but each verse needs to be done separately), for the dynamics,...
it needs some kind of structure like this:
scaleFactor = calcScaleFactorForEntries(List of Entries with x-position and required x-width)
From all calculated scalefactors wethen finally take the maximum.
So that's the master plan i had in mind for this and maybe you can change the code somehow into this direction.
We don't need to have immediately the final goal and e.g. also change the lyrics in this way, but maybe we can get closer to that goal.
We can also take the current version, as it will be already nicer, but it would be nice to make this improvement.
Also i would suggest to render the chord symbols left-aligned (so the start of the chord description should be above the note(timestamp) it belongs to.
Another idea to make the current version already look better is to use superscript for our chord text which will make the width smaller (->reduces the stretching) and looks nicer.
So that was a lot of text, but i hope things make sense :-)

demo/index.js
hartman42: Yeah, I wasn't sure where to put that.  I had it at the top so it would load first for me in the demo and just left it there.  I'm thinking I'll just move it down just below the first chord test.

src/MusicalScore/Graphical/VexFlow/VexFlowMusicSheetCalculator.ts
hartman42: Good question.  I'm not sure really...  I used the function that does the lyrics as a template. I imagine they put the interfaces there to keep it simple?

hartman42:
@matt-uib Thank you for your reply!  I really appreciate it.  I was going to bring up the chord alignment.  That didn't make sense to me either.  I was also thinking of some way to reuse some of the code there for this.  I am stuck on why there's an elongation factor used and not just a width value returned?

hartman42:
Refactored the code a bit here.  I see now why the elongation factor was used.  It seems to be a heuristic method of figuring out what the minimum width of the whole measure ought to be.  Correct me if I'm wrong but it doesn't seem possible to specify exactly the new left position for each staff entry (and subsequent container).

I did discover that the containers (lyrics and chords) are tied to each staff entry however, the elongation factor doesn't consider the previous staff entry unless it has a corresponding lyric or chord associated with it (on the same line).  I believe my example just happened to show otherwise.  You can see that here when the alignment of the chords is left aligned.
image

I did add a current overlap field in the dictionary entry there to track what the total overlap width is for each line of lyrics/chords.  Perhaps we could incorporate this into the heuristic somehow to get rid of some of the overstretching.

sschmidTU:
@hartman42 Feel free to tell me when this is ready for review, even just a first review.
The changes in #933 are hard to see in the samples because of the rendered chord text (symbol) overlaps.

src/MusicalScore/Graphical/VexFlow/VexFlowMusicSheetCalculator.ts
```diff
@@ -295,152 +295,216 @@ export class VexFlowMusicSheetCalculator extends MusicSheetCalculator {
     return minStaffEntriesWidth;
   }
 
-  public calculateMeasureWidthFromLyrics(measuresVertical: GraphicalMeasure[], oldMinimumStaffEntriesWidth: number): number {
-    let elongationFactorForMeasureWidth: number = 1;
+  private calculateElongationFactor(containers: any[], staffEntry: GraphicalStaffEntry, lastEntryDict: any,
+                                    oldMinimumStaffEntriesWidth: number, elongationFactorForMeasureWidth: number,
+                                    measureNumber: number, oldMinSpacing: number, nextMeasureOverlap: number): number {
```
sschmidTU, line 301: I really like that there's one method for both lyrics and chords measure elongation now!

sschmidTU (approved):
Looks great to me now with the changes!
Visual regression tests also only show one diff, a positive change as expected, from OSMD Function Test - Chord Symbols (there were two chords overlapping there before).

sschmidTU:
Merging.

src/MusicalScore/Graphical/GraphicalChordSymbolContainer.ts
```diff
@@ -29,7 +29,7 @@ export class GraphicalChordSymbolContainer extends GraphicalObject {
     }
     private calculateLabel(textHeight: number, transposeHalftones: number, keyInstruction: KeyInstruction): void {
         const text: string = ChordSymbolContainer.calculateChordText(this.chordSymbolContainer, transposeHalftones, keyInstruction);
-        this.graphicalLabel = new GraphicalLabel(new Label(text), textHeight, TextAlignmentEnum.CenterBottom, this.rules, this.boundingBox);
-        this.graphicalLabel.PositionAndShape.RelativePosition = new PointF2D(0.0, 0.0);
+        this.graphicalLabel = new GraphicalLabel(new Label(text), textHeight, TextAlignmentEnum.LeftBottom, this.rules, this.boundingBox);
+        this.graphicalLabel.PositionAndShape.RelativePosition = new PointF2D(-1.0, 0.0);
```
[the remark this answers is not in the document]
hartman42: Yeah, that's something I wrestled with a bit... I always liked chord symbols left-aligned over the notes they start on... that way there's no ambiguity on when the chord change happens - particularly when the chord gets a little long in the tooth.

The `-1` in the position seems to line it up a little better.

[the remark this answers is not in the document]
sschmidTU: makes sense, thanks!

src/MusicalScore/Graphical/VexFlow/VexFlowMusicSheetCalculator.ts
hartman42: Yeah, that's something that was bugging me too!  That whole DRY thing, right?

hartman42:
Hey @sschmidTU,  Thanks for doing this!  I really appreciate it as this is my first experience working with others on here!

sschmidTU:
@hartman42 my pleasure! Great first work, and a superb addition to OSMD! The new chord spacing tests look really ugly on the OLD version before the changes ;)

PR #932, opened 2021-01-13
export SVG via browser and script, integrate feat/exporting, fixes #670

### LINKED ISSUE #670: Export PNG/SVG
Feature: use OSMD to export a PNG (given width, height), or SVG,
to a file or on a REST-Server.

This is an often-requested feature (#523, #491, #437).
It's also useful for building a script or server that uses OSMD to generate score image files.

There is already the [feat/exporting](/opensheetmusicdisplay/opensheetmusicdisplay/tree/feat/exporting)-branch ([this commit](https://github.com/opensheetmusicdisplay/opensheetmusicdisplay/commit/c0e8947d4b42d2a7b810de34f6472e93fc047cd0)), which allows you to download the SVG in the browser.

However, it would be useful to be able to save the file locally via node.js or to have a local server that gives you the files via GET request.
The node.js script is now on the feature branch.

The first step for PNG export, already done, was being able to set the background color of the canvas, because the canvas image always exported with a transparent background color (the white background in the demo comes from the browser/div/style).

So there's now an OSMDOption `pageBackgroundColor: string`, e.g. `"#FFFFFF"` is white, or `#FF000088` is a half-transparent red.
It can also be used as a parameter for the demo: `localhost:8000/?pageBackgroundColor=E1E1E1`
(the `#` has to be omitted for the url parameter)
image
`http://localhost:8000/?showHeader=0&debugControls=0&backendType=canvas&pageBackgroundColor=E1E1E1`

hallvors:
- old branch brought up to date with current `develop`
- added test/Util/generateSvg_browserless.js to batch generate SVG


sschmidTU:
@hallvors i just checked the generateSVG script, and i think we can integrate it into generateImages, the diff between the two is very small. Otherwise we have mostly copied code between the two, and need to maintain both separately.
Here's the diff:
https://www.diffchecker.com/URtZ562z

We can just create a command line argument to select PNG or SVG.

Otherwise it's looking very good!
I'll merge this to a feature branch, merge the scripts, and merge to develop.
Thank you very much for implementing browserless SVG export! :)

test/Util/generateSvg_browserless.js
```diff
@@ -0,0 +1,339 @@
...
+
+    try {
+        await osmdInstance.load(loadParameter) // if using load.then() without await, memory will not be freed up between renders
+    } catch (ex) {
+        console.error('load error: ' + ex)
+    }
+
+    debug('xml loaded', DEBUG)
+    try {
+        await osmdInstance.render()
```
sschmidTU, line 281: just a note: render does not return a promise, it's always synchronous, so no need for await :)

hallvors: That's interesting, because adding await actually seemed to fix things and ensure the markup was ready?! Magic :)

Issue #949, opened 2021-01-27
Grace Note Fingerings, String Numbers (Violin)


jmorag:
First of all, thank you very much for fixing #878. It's been very helpful to be able to insert fingerings on grace notes. Before, I was re-writing pieces without grace notes in order to be able to do that. It got ugly.... Anyway, my application also needs to display string numbers. So far, I've been doing that by hacking in lyrics, but that gets complicated with grace notes and multiple voices. It would be preferable if string numbers were natively supported by OSMD. Keeping with the ysaye example from #878,  Musescore renders 
![image](https://user-images.githubusercontent.com/24941170/105945704-cb93ed00-6033-11eb-9f5b-877cbd72898b.png)
whereas OSMD renders
![image](https://user-images.githubusercontent.com/24941170/105945856-17df2d00-6034-11eb-8221-c26f37a275b7.png)
(ignore the blue note at the end, I manually colored that in the svg for unrelated reasons).

Here's the updated file [test_grace_note_fingerings_and_strings_Ysaye_excerpt.txt](https://github.com/opensheetmusicdisplay/opensheetmusicdisplay/files/5877839/test_grace_note_fingerings_and_strings_Ysaye_excerpt.txt)

This came up a while ago in https://github.com/opensheetmusicdisplay/opensheetmusicdisplay/issues/711. Since then, I've made some progress on the live-editing front. You can see the results at https://open-strings.com. It allows users to upload and edit fingerings and even infers fingerings if you ask it to. Live editing of fingerings in OSMD is accomplished by this [pile of hacks](https://github.com/jmorag/open-strings/blob/master/static/js/fingeringeditor.js) that operate directly on the SVG in lieu of proper re-rendering at the Vexflow level. I doubt any of that code would be useful to upstream but if you think it could be helpful I can work on integrating it into OSMD proper.

sschmidTU:
Done.

![image](https://user-images.githubusercontent.com/33069673/106046664-dbb6d580-60e2-11eb-8aa8-ec6d9a3b2a9c.png)

New EngravingRule:
```js
osmd.rules.StringNumberOffsetY = 0.0; // default, try 5.0 (vexflow units / pixels)
```

jmorag:
Thanks for getting to this so quickly @sschmidTU! It doesn't seem to work for chords, though.
OSMD
[...]

PR #957, opened 2021-02-02
String number chord hotfix

jmorag:
Resolves remaining issues with https://github.com/opensheetmusicdisplay/opensheetmusicdisplay/issues/949


jmorag:
The test example looks like this:
image
Alternating left and right positions for the string numbers was the only way I could figure out avoiding the circles colliding, but I'm happy to use a different strategy if you have a better idea.

sschmidTU:
Sounds good to me. You'd probably have to reduce the size of the string numbers otherwise, and then it could get hard to read (even though fingerings are smaller, but adding a small circle adds noise)

jmorag:
Simply alternating left and right has some bad edge cases. Any note with preceding grace notes and a string annotation comes out weirdly. 
image
My other idea is to, for single notes, put the string always on the right, and for chords, put the bottom string on the bottom, top string on the top, and alternate right and left for the middle notes.
image
image
image

My other idea was to put everything on the right except in cases when the notes are too close together. Is there an easy way to figure out which line of the staff the notes inside a `GraphicalVoiceEntry` lie on? I was looking through the `Note` class but didn't find what I was looking for.

sschmidTU (commented):
> for single notes, put the string always on the right, and for chords, put the bottom string on the bottom, top string on the top, and alternate right and left for the middle notes.

This already sounds like a very decent solution that needs no extra line checking.

> My other idea was to put everything on the right except in cases when the notes are too close together. Is there an easy way to figure out which line of the staff the notes inside a `GraphicalVoiceEntry` lie on?

Yes, this would probably be an even better solution.
There's a way to get a note's line from Vexflow (hopefully stable):
```typescript
for (let i: number = 0; i < vfnote.getKeyProps().length; i++) { // for all notes of a (potential) chord
    const line: number = vfnote.getKeyProps()[i].line;
    console.log("note: " + vfnote.getKeyProps()[i].key + ", line: " + line);
}
```
        
image
[line_check_vexflow_simple_one_measure.musicxml.zip](https://github.com/opensheetmusicdisplay/opensheetmusicdisplay/files/5928195/line_check_vexflow_simple_one_measure.musicxml.zip)

As you can see, in Vexflow, the top treble line (F5) is `5`, the bottom treble line (E4) is `1`, and the violin open g string (G3) is `-1.5`.
We could add this as a helper function to the VexFlowGraphicalNote. (not possible in Note because it doesn't have access to the vfnote)

So, if you could implement this second option, that would be great.
You can also add the helper function to VexFlowGraphicalNote.ts, or let me do it.
Actually, it might be best to save this line in GraphicalNote once the note is created in VexFlowConverter, then we don't need to get the keyprops again in the helper function.

sschmidTU:
> for single notes, put the string always on the right, and for chords, put the bottom string on the bottom, top string on the top, and alternate right and left for the middle notes.

This already sounds like a very decent solution that needs no extra line checking.

> My other idea was to put everything on the right except in cases when the notes are too close together. Is there an easy way to figure out which line of the staff the notes inside a `GraphicalVoiceEntry` lie on?

Yes, this would probably be an even better solution.
There's a way to get a note's line from Vexflow (hopefully stable):
```typescript
for (let i: number = 0; i < vfnote.getKeyProps().length; i++) { // for all notes of a (potential) chord
    const line: number = vfnote.getKeyProps()[i].line;
    console.log("note: " + vfnote.getKeyProps()[i].key + ", line: " + line);
}
```
        
image
[line_check_vexflow_simple_one_measure.musicxml.zip](https://github.com/opensheetmusicdisplay/opensheetmusicdisplay/files/5928195/line_check_vexflow_simple_one_measure.musicxml.zip)

As you can see, in Vexflow, the top treble line (F5) is `5`, the bottom treble line (E4) is `1`, and the violin open g string (G3) is `-1.5`.
We could add this as a helper function to the VexFlowGraphicalNote. (not possible in Note because it doesn't have access to the vfnote)

So, if you could implement this second option, that would be great.
You can also add the helper function to VexFlowGraphicalNote.ts, or let me do it.
Actually, it might be best to save this line in GraphicalNote once the note is created in VexFlowConverter, then we don't need to get the keyprops again in the helper function.

sschmidTU:
> for single notes, put the string always on the right, and for chords, put the bottom string on the bottom, top string on the top, and alternate right and left for the middle notes.

This already sounds like a very decent solution that needs no extra line checking.

> My other idea was to put everything on the right except in cases when the notes are too close together. Is there an easy way to figure out which line of the staff the notes inside a `GraphicalVoiceEntry` lie on?

Yes, this would probably be an even better solution.
There's a way to get a note's line from Vexflow (hopefully stable):
```typescript
for (let i: number = 0; i < vfnote.getKeyProps().length; i++) { // for all notes of a (potential) chord
    const line: number = vfnote.getKeyProps()[i].line;
    console.log("note: " + vfnote.getKeyProps()[i].key + ", line: " + line);
}
```
        
image
[line_check_vexflow_simple_one_measure.musicxml.zip](https://github.com/opensheetmusicdisplay/opensheetmusicdisplay/files/5928195/line_check_vexflow_simple_one_measure.musicxml.zip)

As you can see, in Vexflow, the top treble line (F5) is `5`, the bottom treble line (E4) is `1`, and the violin open g string (G3) is `-1.5`.
We could add this as a helper function to the VexFlowGraphicalNote. (not possible in Note because it doesn't have access to the vfnote)

So, if you could implement this second option, that would be great.
You can also add the helper function to VexFlowGraphicalNote.ts, or let me do it.
Actually, it might be best to save this line in GraphicalNote once the note is created in VexFlowConverter, then we don't need to get the keyprops again in the helper function.

jmorag:
Final idea: assuming the string numbers are for violin family instruments only, we can display them _all_ on the right of notes as roman numerals and patch vexflow not to generate the circles at all. 
Pros: 
- fingers on the left and strings on the right is aesthetically pleasing (to me)
- without circles we don't need to maintain any collision logic, or at least no collision logic that doesn't also apply to finger numbers (unisons are a bit funky but that's true for fingerings as well currently)
Cons:
- maintenance burden of another vexflow patch
Some examples:
image
image
image

jmorag:
If you're not a fan of the roman numerals, I'll implement the second option and maintain the roman numerals myself for my own needs.

sschmidTU:
Oh yes, roman numerals for string numbers sounds great! I'm used to that from practice (i play viola/violin).
I had also briefly thought about removing the string number circles, but then they would be hard to differentiate from fingerings. With roman numerals, that's not a problem.

I think we can adapt roman numerals as the default solution, it seems to be standard in music engraving, at least for the violin family, as you say.

It seems like classical guitar uses encircled string numbers, it would be nice to still have the option to have them circled here.
image

image

sschmidTU:
Another slight con of the "always roman numerals on the right" option may be that it's not the easiest to read in general, and specifically when the notes are close, e.g. the D F# chord in your sample. You could also have two notes only one line apart with different string numbers, e.g. a G on the D string (III) and an A on the open A string (II).

sschmidTU:
Btw this is an interesting guitar example, incidentally the notes are never too close for the circles to be a problem:
image

sschmidTU:
Regarding Vexflow patch burden, yes, that's a slight consideration, but we already have quite a few vexflow patches and are regularly if not frequently adding more, so it probably can't be helped.
(i assume you're familiar with our `src/VexFlowPatch` system)

jmorag:
> Another slight con of the "always roman numerals on the right" option may be that it's not the easiest to read in general, and specifically when the notes are close, e.g. the D F# chord in your sample. You could also have two notes only one line apart with different string numbers, e.g. a G on the D string (III) and an A on the open A string (II).

That seems true for finger numbers as well with the default left placement. Maybe a combination of roman numerals for strings and some staff logic is the answer then.

With the guitar example, all the thirds would have colliding circles if we rendered them without manually adjusting the y-offsets, which we might want to do anyway with the roman numerals.

I think we agree that for string numbers, we want to go ahead with roman numerals by default and add the option to revert to arabic numerals inside circles for guitar music. Annotating GraphicalNote with its staff line is a good idea in general, and we can use it to space out anything too close together. I can work on this in the next few days.

In beat 3 of measure 50 of the sibelius 2nd movement example, there's a unison G. Right now both fingerings and string numbers get rendered on top of each other. What would be the easiest way to go about detecting voice collisions in cases like that, where they're between voices, not notes in the same chord?

sschmidTU:
> > You could also have two notes only one line apart with different string numbers, e.g. a G on the D string (III) and an A on the open A string (II).
> 
> That seems true for finger numbers as well with the default left placement.

Actually, two notes only one line apart would have a horizontal offset, so that shouldn't be a problem.

> I think we agree that for string numbers, we want to go ahead with roman numerals by default and add the option to revert to arabic numerals inside circles for guitar music.

Yes, great, let's do that!

> Annotating GraphicalNote with its staff line is a good idea in general, and we can use it to space out anything too close together. I can work on this in the next few days.

Yes, please, that would be nice!

> In beat 3 of measure 50 of the sibelius 2nd movement example, there's a unison G. Right now both fingerings and string numbers get rendered on top of each other. What would be the easiest way to go about detecting voice collisions in cases like that, where they're between voices, not notes in the same chord?
Can you share that extract? In #949 it only goes from measure 32 to 36. So i'm not sure what that looks like right now.
Collisions between voices are a bit more tricky, since the voice entries are processed separately and independently, but when you have a `GraphicalVoiceEntry` `gve`, you should be able to see `gve.parentStaffEntry.graphicalVoiceEntries`, where you should find the other voice entries and notes from the second voice, and their (staff)lines, assuming we do save them in `GraphicalNote`.

jmorag:
This is what it looks like right now with all the strings to the right

image

[Sibelius_violin_concerto_excerpt.zip](https://github.com/opensheetmusicdisplay/opensheetmusicdisplay/files/5935689/Sibelius_violin_concerto_excerpt.zip)

sschmidTU:
Yes, it seems like for the violin roman numerals are preferable for now, especially with that double unison G.

jmorag:
Good news! Getting rid of circles doesn't require a vexflow patch. We can just set the radius of the circles to 0. Fingerings and string numbers in chords now don't overlap. Voices are still WIP. 

Here's what the file I addded looks like.
image

jmorag:
It kind of works for multiple voices now
image

Still not totally robust to accidentals, but it definitely wasn't before either, so this isn't a regression
image

I want to say this is ready to merge. @sschmidTU what do you think?

sschmidTU:
@jmorag great, good news indeed! I'll review asap.

sschmidTU:
There's an unfortunate side effect in one of our samples - `OSMD Function Test - Tabulature`.
image
(bright red is previous version, black with red border PR version)

With the PR, some of the fingerings disappear into preceding accidentals, etc.

There should probably be an EngravingRule to disable this behavior.
Also, it looks like we don't need to x-shift the fingerings in these cases (a lot of the notes are 2+ lines apart).
Judging from the code in `calculateModifierXOffsets()`, shouldn't it only offset for fingerings if the notes are less than 2 lines apart?

sschmidTU:
Found the fix, there was a `Math.abs()` missing.

jmorag:
You mean in https://github.com/opensheetmusicdisplay/opensheetmusicdisplay/blob/string-number-chord-hotfix/src/MusicalScore/Graphical/VexFlow/VexFlowStaffEntry.ts#L106?

sschmidTU:
One more thing, the roman numerals (+ fingerings) take up a lot of space compared to string numbers.
Wouldn't it be better to take encircled numbers instead of roman numerals when there's only one note per VoiceEntry?

sschmidTU:
I implemented my review changes in the branch `pr/957-review`.

jmorag:
We could do roman numerals on top for the single note case.

sschmidTU:
Right, even better!

src/MusicalScore/Graphical/VexFlow/VexFlowMeasure.ts
```diff
@@ -1394,19 +1404,46 @@ export class VexFlowMeasure extends GraphicalMeasure {
...
+                        stringNumber = "I";
+                        break;
+                    case "2":
+                        stringNumber = "II";
+                        break;
+                    case "3":
+                        stringNumber = "III";
+                        break;
+                    case "4":
+                        stringNumber = "IV";
```
sschmidTU, line 1423: by the way, we should probably support string numbers up to at least 6 / "VI",
e.g. for the guitar.
i'll just add that after merging i think.

src/MusicalScore/Graphical/VexFlow/VexFlowMeasure.ts
```diff
@@ -1394,19 +1404,46 @@ export class VexFlowMeasure extends GraphicalMeasure {
...
+                // disambiguation from fingerings since we use Roman
+                // Numerals for RenderStringNumbersClassical
+                (<any>vfStringNumber).radius = 0;
+                const offsetY: number = -this.rules.StringNumberOffsetY;
+                // if (note.sourceNote.halfTone < 50) { // place string number a little higher for notes with ledger lines below staff
+                //     // TODO also check for treble clef (adjust for viola, cello, etc)
+                //     offsetY += 10;
+                // }
+                if (voiceEntry.notes.length > 1 || voiceEntry.parentStaffEntry.graphicalVoiceEntries.length > 1) {
+                    vfStringNumber.setOffsetX(note.baseStringNumberXOffset * 15);
```
sschmidTU, line 1437: i think 15 is a bit much, uses too much space, 13 looks good to me. i'll also change this after merging.

sschmidTU: a nice enhancement would be to make the font slightly smaller for III, IV and up, compared to I and II.

sschmidTU (approved):
looks great now!
a few minor adjustments will be made after merging.

sschmidTU:
Merging.
Thanks @jmorag for a great addition to OSMD!

jmorag:
You're very welcome, glad I could contribute!

Issue #951, opened 2021-01-28
Hidden stem tag is ignored


massimobio:
In lead-sheets it is common to add slashes for improvised sections.
OSMD seems to ignore the `<stem>none</stem>` tag.
Sibelius:
<img width="653" alt="Screen Shot 2021-01-28 at 5 49 15 PM" src="https://user-images.githubusercontent.com/240599/106208890-16c41200-6192-11eb-80f6-275b479eeead.png">

OSMD 0.9.2:
<img width="710" alt="Screen Shot 2021-01-28 at 5 49 50 PM" src="https://user-images.githubusercontent.com/240599/106208909-20e61080-6192-11eb-8f1b-b9d0fec5bc19.png">

Sample score:
[stems.musicxml.zip](https://github.com/opensheetmusicdisplay/opensheetmusicdisplay/files/5890466/stems.musicxml.zip)

sschmidTU:
This should be pretty easy to fix, we just need to read this and set the `currentVoiceEntry.StemColor` to transparent in `VoiceGenerator.ts`. (ideally just setting `note.StemColorXml` would be enough)

Thanks for the report!

sschmidTU:
The stems are now not displayed anymore, thanks to PR #964.
As written in discord, ideally we should not just make these stems transparent, but not render them at all, using e.g.:
```js
(vfStaveNote as any).shouldRenderStem = false
```

that would be an improvement on the implementation.

sschmidTU:
Example screenshot:
![image](https://user-images.githubusercontent.com/33069673/107416583-efb3fb80-6b14-11eb-80aa-f01077ddc36a.png)
`test/data/OSMD_function_test_chord_tests_various.musicxml`

(first 4 notes have `<stem>none</stem>`)

massimobio:
Fantastic, thank you. Looking forward to 0.9.3

PR #964, opened 2021-02-09
No stems

hartman42:
I started this about a week ago but still can't wrap my head around just removing the stems.  This PR is just making the stems transparent.  This is related to issue #951


sschmidTU:
Good workaround solution for now.
Merging.

PR #968, opened 2021-02-12
Interfaces needs exported for plugins

fredmeister77:
The interfaces need exported as well to enable existing plugins (Transpose) and potentially future plugins.

I think this should be harmless, but let me know if there are any concerns with this.


sschmidTU:
right, good catch, thanks!

PR #985, opened 2021-03-03
Fix Rehearsal mark position after pickup measure

### LINKED ISSUE #983: Rehearsal marks shift when there is a pickup measure.
### MuseScore
スクリーンショット 2021-03-03 14 42 28

### OSM
スクリーンショット 2021-03-03 14 43 11

### MusicXML
[0012-Autumn Leaves.xml.zip](https://github.com/opensheetmusicdisplay/opensheetmusicdisplay/files/6073827/0012-Autumn.Leaves.xml.zip)

I've looked some songs that have pickup bars and this bug is present.

ryohorie:
Closes #983

### DIFF
--- a/src/MusicalScore/ScoreIO/MusicSheetReader.ts
+++ b/src/MusicalScore/ScoreIO/MusicSheetReader.ts
@@ -423,7 +423,7 @@ export class MusicSheetReader /*implements IMusicSheetReader*/ {
      * @returns {boolean}
      */
     private checkIfMeasureIsImplicit(maxInstrumentDuration: Fraction, activeRhythm: Fraction): boolean {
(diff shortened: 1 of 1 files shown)


sschmidTU:
Please note that this change modifies the code at the wrong place and would make the `checkIfMeasureIsImplicit()` function incorrect, and probably say that the measure is not implicit (= pickup emasure). The solution to fixing the rehearsal mark position in scores with pickup measures is not to say that they don't have pickup measures.
I will commit a different fix for #983 which modifies the code at the correct location.
(of course i have more experience with the code base, so as a first try and hint at what the issue may be, a pull request like this is still very welcome, thanks!)

Issue #215, opened 2018-03-06
Microtonal Support


muratcorlu:
Vexflow has support for microtonal accidentals https://github.com/0xfe/vexflow/wiki/Microtonal-Support But as I see, opensheetmusicdisplay is not supporting these accidentals. Is it hard to add these?

bneumann:
No it actually is only putting more cases inside the VexFlowConverter.ts at the accidental method. Actually a quite nice contribution topic

sschmidTU:
@muratcorlu I added support for quarter tones in the commit referenced above, should be in develop/the demo soon. Were you looking for other microtonal accidentals like the Turkish ones?
Unfortunately microtonal accidentals are indeed not trivial to support, and MusicXML and many sheet music programs aren't ideal for them, MuseScore needs an extension for Turkish ones.
One more issue is that we have a transposition engine in OSMD, and transposing microtones is not so easy.
But if you can send me an example MusicXML file with some more microtonal accidentals, preferably those supported by VexFlow, i'll see what i can do.
Closed for now, can be reopened.

PR #998, opened 2021-03-24
Turkish accidentals

stefdoerr:
Hi all! I added support for all turkish accidentals as well as 3-quarter-flat/sharps just in case someone else wants them. 
This PR is related to https://github.com/opensheetmusicdisplay/opensheetmusicdisplay/issues/215

I based my example file on the microtonal file found here: https://github.com/cuthbertLab/musicxmlTestSuite/blob/master/xmlFiles/01d-Pitches-Microtones.xml
And added four more missing accidentals (5 commas flat/sharp, 8 commas flat/sharp)

[01d-Pitches-Microtones-Modified.zip](https://github.com/opensheetmusicdisplay/opensheetmusicdisplay/files/6200625/01d-Pitches-Microtones.zip)
It renders correctly like this now
image

The only thing I changed in the existing code is your workaround for the triple sharp because you were using the "++" symbol which corresponds to three quartertones sharp as per here: https://github.com/0xfe/vexflow/wiki/Microtonal-Support
So I converted it to "###" so that it doesn't collide with the quartertone symbol.

### DIFF
--- a/src/Common/DataObjects/Pitch.ts
+++ b/src/Common/DataObjects/Pitch.ts
@@ -25,6 +25,12 @@ export enum AccidentalEnum {
     TRIPLEFLAT,
     QUARTERTONESHARP,
     QUARTERTONEFLAT,
+    THREEQUARTERSSHARP,
+    THREEQUARTERSFLAT,
+    FIVECOMMASSHARP,
+    FIVECOMMASFLAT,
+    EIGHTCOMMASSHARP,
+    EIGHTCOMMASFLAT,
 }
 
 // This class represents a musical note. The middle A (440 Hz) lies in the octave with the value 1.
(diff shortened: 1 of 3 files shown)


stefdoerr:
Actually now that I think about it, the values I used for the commas are probably wrong.
Since MusicXML defines 1 as the semitone it gets a bit hard to define turkish music which doesn't have the semitone in the middle of the tone.
image

I'll need to look around to see what the correct values are for the commas in MusicXML

sschmidTU:
Nice! Thanks for the contribution!
Code looks good so far. Giving you time to check on the correct comma values in MusicXML.

stefdoerr:
I looked at a discussion of the MusicXML developers with the Mus2 developers
https://forums.makemusic.com/viewtopic.php?f=12&t=2191
It seems like I should use the accidentals tag to define the symbols.

The issue with the alter tag is that if Do - Re = 2 alters then if the turkish has 9 commas then 2/9 = 0.222222222 so you run into rounding issues. And as far as I can tell it does not support fractions, only floats and from a quick search could not find any fixed values for them.

So that would mean that we cannot (or should not) read the accidentals from the alter but from the accidentals tag instead.
We could keep though the 3 quartertones in this PR which are not related to turkish music commas.

I'll close this PR and maybe make a new one on the 3-quartertones while I investigate the turkish conventions better.

sschmidTU:
I mean, the rounding issue doesn't seem too important to me, you could define a maximum accuracy after which you assume it is repeating (e.g. 0.222222222 = 2/9). And even if you don't, you probably won't be able to hear the difference, if the playback software is even that precise.
But if there's a better way to do this, feel free to adapt.

sschmidTU:
Yeah, there is some good code in this PR that could be merged in any case, the e.g. quartertone changes, and naming triple sharp ### instead of ++.

stefdoerr:
I made a new PR now with the triple-quarters. As for the rounding, my issue is mostly that I have the impression that the `<alter>` tags should not be used for guessing the accidentals. Instead the `<accidentals>` tag needs to be defined. That was at least my personal impression after a quick google search but I really just started yesterday looking at musicXML.

If it's a matter of what frequency will be played once there is playback available, then yes the approximation of 0.22222 ought to be close enough to have something workable, although the tuning of the quartertones varies wildly between eastern traditions. 

But we should probably not use this value to guess the accidental.

sschmidTU:
Good, those uncontroversial changes are merged in the new PR #999.
Let me know when you find a solution for the turkish accidentals.

PR #999, opened 2021-03-25
Added support for three-quarter flats and sharps

stefdoerr:
I made a new PR with just the three-quarter flats and shaps and without the turkish accidentals until I investigate further how they are meant to be represented in musicXML.
I also renamed the triple-sharp from `++` to `###` so that it doesn't collide with the naming of the three-quarter sharp.

image


sschmidTU:
Looks good, as discussed in #998. (Visual) tests pass.
Merging.

Issue #886, opened 2020-09-17
How re-render the display based on key


dwarka-code:
Current if i load XML it start with default key D but i want to switch it to E and render the display with this key

ifredom:
change options, then  call render function

sschmidTU:
This is called Transposing. Duplicate of #733 .

dwarka-code:
guys can you give simple code how I can change the key and render based on key

sschmidTU:
The OSMD Transposition plugin is now integrated into OSMD, see https://github.com/opensheetmusicdisplay/opensheetmusicdisplay/issues/733#issuecomment-823530818.
(currently on the develop branch, will be released in the next OSMD version after 0.9.5)

Issue #769, opened 2020-06-03
New Transpose Function for OSMD


AlbertHart:
**Transposing Scores in OSMD**

ice6 and I have been working on a transpose function for OSMD which we would like to include as part of the OSMD code.

We have created a demonstration of the new feature here:

[](https://ice6.github.io/opensheetmusicdisplay/build/)

You can load or paste a score, select a transpose selection and see it transposed.

Transposing is controlled by two new options on the Demo Page

![image](https://user-images.githubusercontent.com/8965352/83698977-14d48f80-a5c0-11ea-83b7-8dff44eb9b2d.png)

The DEMO code calls a single function - transpose_xml() after loading the xml file and before passing it to the OSMD rendering engine. This could easily be called by any OSMD developers wanting to offer an option to transpose the Music XML files.

[sschmidTU: [code here](https://github.com/ice6/opensheetmusicdisplay/blob/master/demo/transpose_functions.js) ]

We propose to add transpose_xml() to the core OSMD code, and also to add the call to it to the demonstration page, so it can be easily tested.

This would be available, of course, to all OSMD participants,

If this seems like a good idea to, we can create a fork, add the code, then let you test it and incorporate it.

**Bugs, etc.**

I am sure you will find some bugs as soon as you try to test it. We can fix most of them easily,

And ice6 and I will both be willing to make changes to the code as bugs are found or new features are requested.

sschmidTU:
The OSMD Transposition plugin is now integrated into OSMD, see https://github.com/opensheetmusicdisplay/opensheetmusicdisplay/issues/733#issuecomment-823530818.
(currently on the develop branch, will be released in the next OSMD version after 0.9.5)

Still, the transposition plugin you've been working on here looks very nice, and the transposition by key and direction instead of semitones is a feature we could add as well.

Issue #613, opened 2019-11-04
Transposing XML before loading it into OSMD


AlbertHart:
i wrote a Javascript function to transpose the XML before it is loaded into OSMD.

Basically, it reads the XML - line by line - and then rewrites it, while transposing just 4 entries.

`<fifths>-4</fifths>`

```
<pitch>
     <step>D</step>
     <octave>4</octave>
    </pitch>
```

```
<root>
      <root-step>B</root-step>
     <root-alter>-1</root-alter>
</root>
```

```
<bass>
     <bass-step>F</bass-step>
     <bass-alter>1</bass-alter>
    </bass>
```

There is a little work involved to get #'s vs b's to match key and chords, and to change the octave properly. And I still need to run it through a little more testing.

But I anyone would like a copy of it, let me know. I can post it here if you like.

Or if OSMD wants to look an including it in the full version.

This is going to be great for my project because I can easily transpose and print my scores without having to return to Sibelius, transopose, resave as MusicXML and reload.

#580

sschmidTU:
If you can share your code, that would be great. Do you have a fork of the repository?
Also, if you want to transpose pieces with explicitly given key signatures (at the beginning of the staffline), you'll need to transpose the key instruction too.

ice6:
If I want to implement this feature, where should I start the code?

change the musicxml part or change the omsd object ?

AlbertHart:
The routine I wrote takes the loaded, but not processed, ascii xml text and reads things like <pitch>, and <fifth> and transposes them. If you have a musicxml file you want to send me, I'll dust the off and transpose it. It also attempts to reverse stems, and accidentals, etc.

One use would be as a function people can call after loading the XML ascii file to tarnspose the before processing it. We could add this to the OSMD demo. WHich would then load a musicxml file, but display it transposed. This would be the function we would like to expose to other developers.

Another use would be a stand-alone musicxml transposer. Which would be useful for people using musicxml for other purposes. It is also valuable as a debugging tool for us. For example, I can output musicxml from Sibelius, run it through my routine, and compare it to a second musicxml file which was transposed in sibelius to look for items I may have missed.

If you want to write a similar routine, I'll be glad to send my routine as a starting point. It is in Javascript and not in typescript - which I haven't mastered (or done anything useful with yet)

![image](https://user-images.githubusercontent.com/8965352/81614256-028a7b80-939d-11ea-8ece-2d9af4be8f7d.png)

[...]

Issue #580, opened 2019-10-14
How to do transposing


naranbaat:
Some notes happen at really low place out of range of staff lines.
This is the screenshot from demo
![rendering low](https://user-images.githubusercontent.com/48118372/66746110-8934b400-eeb3-11e9-8dce-4ee5e3374026.png)

------------------------------------------------------------------------------------------------------------
That is the screenshot from the same music xml file using [another tool](https://www.soundslice.com/musicxml-viewer).
![rendering using transposing](https://user-images.githubusercontent.com/48118372/66746321-0ceea080-eeb4-11e9-8e06-c245407aab93.png)

I want them to display above using transposing like above.
Is there any option to do it easily?

matt-uib:
@naranbaat that is strange and should work correctly for a standard bass clef.
I suspect the problem within the xml file. With which program was the file created? did you try to load it in another program (not the one where it has been created)? It seems to me that the octave of the pitch is not correctly set in the xml.

naranbaat:
It's created by Sibelius.
let me attach file.
[allforyou.xml.txt](https://github.com/opensheetmusicdisplay/opensheetmusicdisplay/files/3724708/allforyou.xml.txt)

matt-uib:
![grafik](https://user-images.githubusercontent.com/17180665/66757288-2257d080-ee9c-11e9-9270-c1888b07c1e4.png)

I tried it with musescore - same problem. So really the octave is wrong in the xml.

matt-uib:
Soundslice seems to either render it wrong or has some logic which tries to "fix" this...

naranbaat:
yeah,  i want to know how to do transposing in that case like Soundslice.

matt-uib:
The right way to me would be: Open some editor like musescore (free for download), transpose there and save.
We have nothing built in for that purpose, but you could also iterate all measures, within all staffentries, within all voiceentries and within all notes and change the octave in the pitch.
But i would rather remove the cause and not the syptom...

naranbaat:
That make sense.

naranbaat:
I think I can write this feature for OSMD.
Let me know how can i request to commit to repository for OSMD project.

sschmidTU:
fork the repository, submit a pull request

ice6:
is there any progress about the transposing function?

sschmidTU:
[...]

PR #1012, opened 2021-04-20
Release of Transpose plugin

### LINKED ISSUE #733: Transposing

fredmeister77:
Fixes #733 (makes it open source).

Also see #886, #769, #613, #886, #580, and #132, all obsolete (all issues about transposing!).

### DIFF
--- a/demo/index.html
+++ b/demo/index.html
@@ -63,7 +63,7 @@
             </div>
             <div class="item">
                 <div class="ui toggle checkbox">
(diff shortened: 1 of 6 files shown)


sschmidTU (approved):
Looks great! Also tested, works fine in the demo with the new UI. Merging (soon).

There is still an occasional problem when you transpose a second time after load and render to a new key like F# or B (+3 or -4 for Beethoven Geliebte),
but that can be solved by cleanly loading a sample (again) and setting the Transpose before the first render.

PR #1014, opened 2021-04-26
Adapt the cursor to single line and multiple lines

BppleMan:
In single-line rendering, the cursor is always set to block:center, which will cause the vertical axis to be centered, but users may have their own needs, so block should not be set during single-line rendering, but inline needs to be set to follow cursor


sschmidTU (approved):
Looks good, i'll trust that this should be better for most people (i almost never use single staffline rendering). Otherwise we'll add another option.
Thanks for submitting the change!

PR #1023, opened 2021-05-08
change cursor width when its  type is '1'

wuxq:
as cursor width is 3 when in 'default' type,  style of left need take away half width（1.5）， so in type of '1' need take away 0.25


sschmidTU:
The intention of cursor type 1 is to be on the left of the current notes, as declared in the `CursorsOptions` interface (`OSMDOptions.ts`).

With this change, the type 1 cursor would be a thin line through the center of the current notes:



That's also possible, but this would be a new cursor type.

PR #1067, opened 2021-10-04
docs: docstring correction

joshstovall:
Fixed error in the docs (cursor follow/alpha descriptions were switched)


sschmidTU:
Good catch, thanks! :)

PR #1083, opened 2021-11-07
Vexflow 4.0 migration (includes webpack 5.0) 

rvilarl:
No default Vex, classes exported in vexflow.

PR #1059, opened 2021-09-02
Migrate to webpack 5

jmorag:
I've been playing around with building this as an esm module and a necessary prerequisite is to upgrade to webpack 5 which supports that. Haven't gotten the esm stuff working yet, but this is nice to have on its own.

PR #1085, opened 2021-11-10
Vexflow migration 4.0.0 (old packages)

rvilarl:


rvilarl:
Hi Simon, I have reverted to the old packages as there were some issues with the new ones and as you said, let us keep the issues separated.

rvilarl:
This integration is complete with all checks passed!

rvilarl:
@sschmidTU I would say that there is some problem with ties, can you have a look at the demo? Thanks

sschmidTU:
As said in #1083, please remove all the refactors that change the return type from Node to void.

rvilarl:
> As said in #1083, please remove all the refactors that change the return type from Node to void. 

I made that because I saw that svg was returning Node and canvas undefined and I thought that if it works with undefined it was better to unify.

> Maybe we do want the new webpack to improve the build process, though it seems to compile for now. But yeah, a few packages will be problematic to update, so i appreciate not updating all of them at once.

I took out the packages changes based on your comment and also based on the fact that I awas getting the same error as in #1059:

✖ multiple instances

rvilarl:
Please confirm if you still want the Node change reverted and if this is the case, I will do it.

sschmidTU:
> I made that because I saw that svg was returning Node and canvas undefined and I thought that if it works with undefined it was better to unify.

Canvas can't return a node because it doesn't have a DOM.

> Please confirm if you still want the Node change reverted and if this is the case, I will do it.

Yes, please revert it.

rvilarl:
Done

src/MusicalScore/Graphical/VexFlow/CanvasVexFlowBackend.ts
```diff
@@ -106,28 +107,29 @@ export class CanvasVexFlowBackend extends VexFlowBackend {
...
-            this.CanvasRenderingCtx.fillStyle = VexFlowConverter.style(styleId);
+            const fillStyle: string | undefined = VexFlowConverter.style(styleId);
+            if (fillStyle) {this.CanvasRenderingCtx.fillStyle = fillStyle;}
         }
         this.CanvasRenderingCtx.globalAlpha = alpha;
         this.ctx.fillRect(rectangle.x, rectangle.y, rectangle.width, rectangle.height);
         this.CanvasRenderingCtx.fillStyle = old;
         this.CanvasRenderingCtx.globalAlpha = 1;
-        return undefined; // can't return dom node like with SVG
+        return undefined; // can't return svg dom node
```
sschmidTU, line 129: please revert this comment change. the previous one is clearer. of course you can't return a svg node in canvas.

rvilarl: Only this instance or make the same comment on the other instances within CanvasVexFlowBackend? They should all be the same within this file, should not them?

rvilarl: Ok I will unify to the clearer one

sschmidTU:
It looks like Github is having some issues with comments on PRs, i can't reply to some of them anymore at some point, and sometimes it shows the comment belonging to the wrong code line.

Responding to:
image
You only need to revert this one comment change. I don't see where you could make a similar change either.

src/MusicalScore/Graphical/VexFlow/CanvasVexFlowBackend.ts
rvilarl: Done

Issue #1081, opened 2021-11-04
Fingering can overlap notes especially with chords, with FingeringPosition Above/Below


sschmidTU:
![image](https://user-images.githubusercontent.com/33069673/140360709-ced2c042-7458-423e-b5d0-259aee469bd5.png)
[Fingering_Overlap_Test_Chord_Bass.zip](https://github.com/opensheetmusicdisplay/opensheetmusicdisplay/files/7476079/Fingering_Overlap_Test_Chord_Bass.zip)

`osmd.rules.FingeringPosition = 0 // above`
or:
`osmd.rules.FingeringPosition = 1 // below`

Also see #350 for initial implementation.

With the default fingering position *Left* (`2`), this is not a problem:
![image](https://user-images.githubusercontent.com/33069673/140362197-01f99774-5b11-4e43-913d-fcba6781f006.png)

sschmidTU:
Done, see #1086 for comments and visual diffs.

Here are three images of a sample with skybottomline and label bounding boxes:
![image](https://user-images.githubusercontent.com/33069673/141202590-6de5bb8c-2ea3-480c-a63c-3198ccb733d7.png)

Here are the relevant osmd.EngravingRules (options) with the default settings:
```typescript
        this.FingeringPosition = PlacementEnum.AboveOrBelow; // AboveOrBelow = correct bounding boxes
        this.FingeringPositionFromXML = true;
        this.FingeringPositionGrace = PlacementEnum.Left;
        this.FingeringInsideStafflines = false;
        this.FingeringLabelFontHeight = 1.7;
        this.FingeringOffsetX = 0.0;
        this.FingeringOffsetY = 0.0;
        this.FingeringPaddingY = -0.2;
        this.FingeringTextSize = 1.5;
```

PR #1086, opened 2021-11-10
Fix Fingering collisions (#1081), make Above/Below default position

sschmidTU:
before: (left / FingeringPositionFromXML)
image

after: (AboveOrBelow, FingeringPositionFromXML)
image
[test_Fingerings_Simple_Chords_Treble_Bass.zip](https://github.com/opensheetmusicdisplay/opensheetmusicdisplay/files/7516089/test_Fingerings_Simple_Chords_Treble_Bass.zip)

visual diffs:
[diffs_fingering_above_below_collision_fix2.zip](https://github.com/opensheetmusicdisplay/opensheetmusicdisplay/files/7516223/diffs_fingering_above_below_collision_fix2.zip)

a valid alternative to the new above/below default, also to save space, is left (or right) positioning:
```js
osmd.EngravingRules.FingeringPosition = 2; // PlacementEnum.Left
osmd.EngravingRules.FingeringPositionFromXML = false; // otherwise an XML placement 'below' will put fingering for those notes below
```

image

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -37,7 +37,7 @@ module.exports = {
           }
         }
     ],
-    "ignorePatterns": ["webpack*.js"],
+    "ignorePatterns": ["webpack*.js","karma.conf.js"],
     "rules": {
         "@typescript-eslint/dot-notation": "error",
         "@typescript-eslint/explicit-function-return-type": "error",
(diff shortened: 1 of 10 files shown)


sschmidTU:
Grace note fingerings are still an edge case, currently still created with the Vexflow method, by default left:
image

`osmd.EngravingRules.FingeringPositionGrace = 5; // PlacementEnum.AboveOrBelow`:
image

Left seems to be the better default for now, because above/below, the spacing is different, and the grace fingerings are bold. Also, you wouldn't be able to set their size in the same way.

Ideally, we would implement the grace note fingerings as labels as well, but it could become tricky with the grace note position and relation to the staffentry position etc.

PR #1088, opened 2021-11-11
Vexflow 4.0 migration

rvilarl:


rvilarl:
@sschmidTU Can you look please at the build error? I have reverted to a state that it was working and now it is not working any more. The compilation works perfectly fine locally in my computee

sschmidTU:
@rvilarl it looks like the build fails because of the Node return type refactorings. did you revert them and forget to push them? did you forget to push your last commit in general? in any case, it should help to revert the Node refactorings.

PR #1089, opened 2021-11-11
migration vexflow 4.0

rvilarl:
I decide to re-start the migration.

PR #1091, opened 2021-11-12
Migration/vexflow4.0webpack5.0

rvilarl:
Webpack 5.0

Trying to resolve Vex.Flow. removal issue with webpack 5.0


sschmidTU:
Just making a note: Will be reviewed/merged after the vexflow 4.0 without package changes is merged.

rvilarl:
> Just making a note: Will be reviewed/merged after PR #1089 (with vexflow 4.0 but without package changes) is merged.

Hi Simon, it is a trial to resolve the Vex.Flow. prefix. I have tried everything that @ronyeh and I could think about and we had the idea that perhaps it was due to the old packages, we will see :)

sschmidTU:
@rvilarl not sure what the exact problem is right now, but have you tried looking at `.appveyor.yml`? It's still using node 12, so you could try node 16 (current LTS).

rvilarl:
Wow! Checks passed! 
I will also look into the node topic, thanks Simon!

rvilarl:
What I am trying to resolve is to remove the Vex.Flow. this has not been possible without updating the packages.🙂

rvilarl:
@sschmidTU the node 16 also does not solve the problem of process indefinition with webpack 5 :( This is an issue to be resolved. And something funny because it fails on the CI environment but not run locally.

rvilarl:
Ready to review, all Vex.Flow. removed.

sschmidTU:
@rvilarl just change the `drawableBoundingBoxElement` string to "None" instead of "false", as i mentioned in the review comment above.

rvilarl:
> @rvilarl just change the `drawableBoundingBoxElement` string to "None" instead of "false", as i mentioned in the review comment above. It should have the same effect, but "false" is not a correct value. We check for certain values, one of them being "None".

Done

PR #1111, opened 2021-12-12
Fix follow for multiple cursors

rvilarl:
@sschmidTU testing with the migration I have realised that the independent control of follow for multiple cursors was not working.

Do you agree with the approach?
- the general one, if set, forces to follow all the cursors.
- the independent one, if set, forces to follow this particular cursor

### DIFF
--- a/src/OpenSheetMusicDisplay/Cursor.ts
+++ b/src/OpenSheetMusicDisplay/Cursor.ts
@@ -169,7 +169,7 @@ export class Cursor {
     const measurePositionAndShape: BoundingBox = this.graphic.findGraphicalMeasure(iterator.CurrentMeasureIndex, 0).PositionAndShape;
     this.updateWidthAndStyle(measurePositionAndShape, x, y, height);
 
-    if (this.openSheetMusicDisplay.FollowCursor) {
+    if (this.openSheetMusicDisplay.FollowCursor || this.cursorOptions.follow) {
       if (!this.openSheetMusicDisplay.EngravingRules.RenderSingleHorizontalStaffline) {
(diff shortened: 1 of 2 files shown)


sschmidTU:
Why would you ever want to follow multiple cursors?
Don't you just want to always follow the first cursor?
As far as I saw, multiple cursor were used as a utility to show additional things, not to be followed.
But I have to admit I haven't really seen a use case for multiple cursors in action from other people.

rvilarl:
I do not know. In my app I just want to follow one of them but we did allow to configure several each with its follow config and I guess we should be consistent. Another possibility is to remove the independent configuration and state that we just follow the first one.

rvilarl:
The current implementation follows all of them. Setting the focus to the next one that moves.

rvilarl:
There are posible solutions but the current one makes no sense. All cursors are followed when the global option is set.
The first question is: **do we want to allow more than one cursor to be follow?** In my app the current play cursor is follow and the tempo cursor is not follow.

sschmidTU:
Yeah, if it's possible to follow any cursor other than the main cursor, it makes sense to let the developer choose which cursor to follow. So the .follow boolean looks good to me, simple and hopefully working without issues.

rvilarl:
> Yeah, if it's possible to follow any cursor other than the main cursor, it makes sense to let the developer choose which cursor to follow. So the .follow boolean looks good to me, simple and hopefully working without issues. Maybe make it && .follow instead of `or`? Otherwise it would always follow every cursor if `openSheetMusicDisplay.FollowCursor` is true, right?

You are right, and makes more sense. But in order to make does work I had to change the default to true when options are not provided (case of one cursor). Done and ready to merge!!!!

PR #1105, opened 2021-11-27
npm-upgrade

rvilarl:
Npm packages updated

PR #1119, opened 2022-01-07
dependencies: update to webpack 5, patch karma-webpack

sschmidTU:
Includes Npm updates from #1105, which was first merged to feat/webpack5.
Karma-webpack 5.0 needs some patches to work correctly with Windows and MacOS file paths, which are now complete.


sschmidTU:
Turns out the npm test error in Windows comes from webpack, not directly from karma-webpack,
and it looks like it's not simple to fix:
webpack/webpack#12759

So we'll have to give up windows multiple drive support for now and just merge this,
this is the easiest option,
as it works on MacOS (with patch) and Linux now.

Issue #1084, opened 2021-11-09
MusicXML accidental support 


infojunkie:
This is a documentation of the currrent state of MusicXML accidental support in OSMD, based on the currently used and latest VexFlow versions, respectively. The end goal is to support all MusicXML accidentals.

| [MusicXML](https://w3c.github.io/musicxml/musicxml-reference/data-types/accidental-value/) | [OSMD](https://github.com/opensheetmusicdisplay/opensheetmusicdisplay/blob/87b681f79ac2838be1f79ef406b31553664c3f2d/src/Common/DataObjects/Pitch.ts#L12-L31) | [VexFlow 1.2.93](https://github.com/0xfe/vexflow/blob/430f607053f7caa24cb96feac722c1b5fb821b06/src/tables.js#L303-L324) | [Vexflow master](https://github.com/0xfe/vexflow/blob/f79be12f85f2e82c69d95afffe8ab9683f60fa9f/src/tables.ts#L55-L319) | Noes |
| --- | --- | --- | --- | --- |
| sharp | SHARP | # | | | 
| natural | NATURAL | n | | | 
| flat | FLAT | b | | |
| double-sharp | DOUBLESHARP | ## | | |
| sharp-sharp | | | | TBD if that concatenation of accidentals is feasible | 
| flat-flat | DOUBLEFLAT | bb | | |
| natural-sharp | | | | TBD Verify if concatenation of accidentals is feasible |
| natural-flat | | | | TBD Verify if concatenation of accidentals is feasible |
| quarter-flat | QUARTERTONEFLAT | d | | |
| quarter-sharp | QUARTERTONESHARP | + | | |
| three-quarters-flat | THREEQUARTERSFLAT | db | | |
| three-quarters-sharp | THREEQUARTERSSHARP | ++ | | |
| sharp-down | | | | |
| sharp-up | | | | |
| natural-down | | | | | 
| natural-up | | | | |
| flat-down | | | | |
| flat-up | | | | |
| double-sharp-down | | | | |
| double-sharp-up | | | | |
| flat-flat-down | | | | |
| flat-flat-up | | | | |
| arrow-down | | | | |
| arrow-up | | | | |
| triple-sharp | TRIPLESHARP | ### | | |
| triple-flat | | | | TBD Verify if concatenation of accidentals is feasible |
| slash-quarter-sharp | - | +- | | Not encoded in OSMD |
[...]

PR #1140, opened 2022-02-08
Full MusicXML accidental support #1084

infojunkie:
This PR is still a work in progress. The aim is to support all MusicXML accidentals as per #1084. My approach is to create a data structure that will contain the info needed to translate between MusicXML, OSMD and VexFlow accidentals, instead of maintaining giant `switch` statements that are scattered over multiple functions and will become unmanageable with the full set of accidentals.

### DIFF
--- a/src/Common/DataObjects/Pitch.ts
+++ b/src/Common/DataObjects/Pitch.ts
@@ -35,6 +35,139 @@ export enum AccidentalEnum {
     KORON
 }
 
+/** Accidentals database.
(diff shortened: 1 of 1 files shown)


rvilarl:
@infojunkie why did you close?

infojunkie:
Oops I renamed the branch and it closed the PR :sob: I will open another one with the renamed branch.

infojunkie:
https://github.com/opensheetmusicdisplay/opensheetmusicdisplay/pull/1155

PR #1154, opened 2022-03-06
Chord accidentals

infojunkie:
This PR replaces "b", "#" and "x" chord accidentals with their more accurate counterparts "♭", "♯" and "𝄪". This helps clarify the chord display, as per attached screenshot.

screenshot-blog karimratib me-2022 03 06-13_09_56

### DIFF
--- a/src/MusicalScore/Graphical/EngravingRules.ts
+++ b/src/MusicalScore/Graphical/EngravingRules.ts
@@ -763,8 +763,8 @@ export class EngravingRules {
 
     public resetChordNames(): void {
         // addChordName(alternateName, chordKindText, adds, alters, subtracts)
-        this.addChordName("alt", "major", ["#5", "b9", "#9"], ["b5"], []);
-        this.addChordName("7alt", "dominant", ["#5", "b9", "#9"], ["b5"], []);
+        this.addChordName("alt", "major", ["♯5", "♭9", "♯9"], ["♭5"], []);
+        this.addChordName("7alt", "dominant", ["♯5", "♭9", "♯9"], ["♭5"], []);
         this.addChordName("7sus4", "dominant", ["4"], [], ["3"]);
         this.addChordName("7sus4", "suspendedfourth", ["7"], [], []);
         this.addChordName("9sus4", "dominantninth", ["4"], [], ["3"]);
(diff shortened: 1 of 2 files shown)


sschmidTU:
I like it, though I can also imagine some people would prefer to use ascii symbols. Maybe we should make it an option?

`♭` is a smaller/narrower character than `b` and can be harder to read, e.g. here:

image

full visual diffs:
[visual_diffs_chord_accidentals_symbols.zip](https://github.com/opensheetmusicdisplay/opensheetmusicdisplay/files/8225980/visual_diffs_chord_accidentals_symbols.zip)

-----

Also, why does this PR change how chordtexts / aliases are rendered?
before (chord spacing test):
image

after:
image

I'm not sure if it's good or bad, but I don't know why this would change from the changes in this PR.

infojunkie:
> I'm not sure if it's good or bad, but I don't know why this would change from the changes in this PR.

Yes, that's strange. I'll look more deeply into it.

> I like it, though I can also imagine some people would prefer to use ascii symbols. Maybe we should make it an option?

That's a subjective assessement that can only be answered by actually asking people, so maybe adding the option is the best answer.

infojunkie:
Found the regression: The code reads the generated chord name and rewrites it in some cases. 

Added the updated visual diff: [visual_regression_diff.zip](https://github.com/opensheetmusicdisplay/opensheetmusicdisplay/files/8238877/visual_regression_diff.zip)

Should I proceed with a new `EngravingRules` option? I would use `ChordAccidentalsUnicode` (`true` / `false` defaults to `false`) or please suggest another name :-)

infojunkie:
I added `EngravingRules.ChordAccidentalTexts` as a dictionary from `AccidentalEnum` to `string`, which is built in the constructor and defaults to regular characters. It can be modified later via `EngavingRules.resetChordAccidentalTexts()` with a boolean flag to use Unicode characters instead.

sschmidTU (approved):
Great, good solution, thank you! Visual tests confirmed. Merging.

I appreciate that you can set each value customly, e.g. `osmd.rules.ChordAccidentalTexts.setValue(1, "♭")`. (in typescript, use AccidentalEnum instead of number), or set all of them via `osmd.rules.resetChordAccidentalTexts(osmd.rules.ChordAccidentalTexts, false)` for non-unicode characters.
(`ChordAccidentalTexts` probably didn't need to be a method parameter, but it's not important)

infojunkie:
Will this be merged into `osmd-extended` ?

sschmidTU:
Yes, I haven't gotten around to merging osmd to osmd-extended recently, will do that soon!

sschmidTU:
@infojunkie merged.

PR #1158, opened 2022-03-23
Compute skylines and bottom lines in batches (+ WebGL)

paxbun:
image

# Introduction

The current rendering logic requires calculation of skyline and bottom line information of each measure, which is the main bottleneck of OSMD. For example, when rendering [Clementi's Sonatina Op. 36 No. 3](https://github.com/opensheetmusicdisplay/opensheetmusicdisplay.github.io/blob/master/demo/MuzioClementi_SonatinaOpus36No3_Part1.xml) with OSMD on Microsoft Edge in a M1 Mac mini device which takes 765 milliseconds on average, skyline and bottom line computation takes about 450 milliseconds. Almost half of the computation is for drawing measures in a temporary canvas, and another half is for retrieving pixel data (`ctx.getImageData()`) from the canvas.

The current logic draws a single measure on a single canvas, so the number of calls to `ctx.getImageData()` is equal to the number of measures, which makes the computation very slow. This fix is to reduce the bottleneck of retrieving pixel data by drawing multiple measures in a single canvas, so OSMD can render the sheet music with fewer number of calls to functions accessing the pixel data. This fix also introduces WebGL-accelerated skyline and bottom line calculation, which computes the lines of measures on a single canvas simultaneously.

# Summary of change

* Removed `calculateLines()` from `SkyBottomLineCalculator`.
* Renamed `SkyBottomLineCalculator` to `SkyBottomLine.ts`, as it does not contain the calculation logic.

* Added classes for batch calculation
  * `SkyBottomLineBatchCalculator`
  * `SkyBottomLineBatchCalculatorBackend`
    * `PlainSkyBottomLineBatchCalculatorBackend`
    * `WebGLSkyBottomLineBatchCalculatorBackend`
  * `SkyBottomLineCalculationResult`

* Updated module rules in `webpack.common.js` to import GLSL files
  * Added `global.d.ts` for typing

* Moved calls to `SkyBottomLineCalculator` from `MusicSheetCalculator` to `VexFlowMusicSheetCalculator`
  * `SkyBottomLineCalculator` references `VexFlow*` classes which makes a circular dependency. For some reason, `MusicSheetCalculator` in
    ```typescript
    class VexFlowMusicSheetCalculator extends MusicSheetCalculator
    ```
    is undefined in runtime. I had to remove this circular dependency to fix this issue.

# Benchmarks

(The width of `body` is set to `900px` in all tests)

* Initial rendering time in milliseconds

| Browser        | Before | Plain  | WebGL  |
| -------------- | ------ | ------ | ------ |
| Edge (Avg)     | 765.61 | 586.26 | 494.49 |
| Edge (Stdev)   | 34.18  | 25.80  | 16.83  |
| Safari (Avg)   | 656.28 | 489.55 | 577.97 |
| Safari (Stdev) | 40.76  | 32.15  | 40.17  |

* Initial rendering time improvements

| Browser | Plain | WebGL |
| ------- | ----- | ----- |
| Edge    | x1.30 | x1.55 |
| Safari  | x1.34 | x1.14 |

* Rendering time taken in repetitive rendering
  (The improvement becomes x1.55 - x1.7 with WebGL on Edge in subsequent renderings)
  Microsoft Edge 99
  image
  Safari 15 4
  image

# What if the browser does not support WebGL?

`PlainSkyBottomLineBatchCalculatorBackend` operates as the fallback logic.

# What makes the WebGL version faster?

From the canvas where the measures are drawn (see the image on the top), the WebGL logic generates an image which has the same width with the canvas and whose height is equal to the number of rows of the canvas. That is, the number of pixels in the output are 300 times fewer than the input, which makes the time taken to `gl.readPixels` shorter. Also, the WebGL version computes the lines of all measures in the canvas at the same time.

# Why is the WebGL version is slower on Safari?

While `gl.texImage2D` takes about 12 milliseconds on Microsoft Edge (Chromium), `gl.texImage2D` takes about 90 milliseconds on Safari. `gl.texImage2D` converts the canvas where the measures are drawn into a texture, so there might be an implicit pixel data copy. There might be another browser which shows similar behavior. We have to test on as many browsers as we can, so we can select specific `SkyBottomLineBatchCalculatorBackend` according to the value of `navigation.userAgent`.

# Potential issues

* The canvas where the measures are drawn is big (about 2400 x 1500), so there might be some memory issue.

# Benchmark codes

(I don't have detailed banchmark with other `IOSMDOptions`, but the trend is similar)

```typescript
import { IOSMDOptions, OpenSheetMusicDisplay } from "opensheetmusicdisplay";

const url = "https://opensheetmusicdisplay.github.io/demo/MuzioClementi_SonatinaOpus36No3_Part1.xml";

function getOSMDOptions(): IOSMDOptions {
  return {
    autoResize: true,
    backend: "svg",
    drawLyricist: false,
    drawLyrics: false,
    drawFingerings: true,
    drawTitle: false,
    drawComposer: false,
    drawCredits: false,
    drawSubtitle: false,
    drawPartNames: false,
    drawPartAbbreviations: false,
    drawingParameters: "compact",
  };
}

async function main() {
  const container = document.getElementById("container")! as HTMLDivElement;
  const options = getOSMDOptions();
  const osmd = new OpenSheetMusicDisplay(container, options);
  await osmd.load(url);

  const samples: number[] = [];
  for (let i = 0; i < 10; ++i) {
    const start = window.performance.now();
    osmd.render();
    const end = window.performance.now();
    samples.push(end - start);
  }

  for (const sample of samples) {
    console.log("OSMD: ", sample);
  }

  const result = `[${navigator.userAgent}] ${samples.join(",")}`;
  await fetch("C# backend URL - code is below", {
    method: "POST",
    body: result
  });
}

window.addEventListener("load", main);
```

```csharp
using System.Text.RegularExpressions;

var builder = WebApplication.CreateBuilder(args);

builder.Services.AddCors(options =>
{
    options.AddDefaultPolicy(policyBuilder =>
    {
        policyBuilder.AllowAnyOrigin();
        policyBuilder.AllowAnyHeader();
        policyBuilder.AllowAnyMethod();
    });
});

var app = builder.Build();
var statistics = new Dictionary<string, List<List<double>>>();

app.UseCors();
app.MapPost("/", async context =>
{
    StreamReader reader = new(context.Request.Body);
    Regex bodyRegex = new(@"\[(.*)\] (.*)");

    string body = await reader.ReadToEndAsync();
    Match match = bodyRegex.Match(body);
    if (!match.Success)
        return;

    string userAgent = match.Groups[1].Value;
    string timeElapsed = match.Groups[2].Value;

    List<double> results = new();
    foreach (var element in timeElapsed.Split(","))
    {
        if (!double.TryParse(element, out double result))
            return;
        results.Add(result);
    }

    if (!statistics.TryGetValue(userAgent, out List<List<double>>? list))
    {
        list = new List<List<double>>();
        statistics.Add(userAgent, list);
    }

    list.Add(results);
});

app.MapGet("/", () =>
    statistics.Select(pair =>
    {
        (var userAgent, List<List<double>> samples) = pair;
        int maxLength = samples.Select(list => list.Count).Max();

        List<ResultEntry> entries = new();
        for (int i = 0; i < maxLength; ++i)
        {
            double[] ithSamples = samples.Where(list => i < list.Count).Select(list => list[i]).ToArray();
            double average = ithSamples.Sum() / samples.Count;
            double standardDeviation = 0.0;
            if (ithSamples.Length != 1)
            {
                double variation = ithSamples.Select(sample => Math.Pow(sample - average, 2)).Sum() / (samples.Count - 1);
                standardDeviation = Math.Sqrt(variation);
            }
            
            entries.Add(new ResultEntry(average, standardDeviation, ithSamples.Length));
        }
        
        return new Result(userAgent, entries);
    })
);

app.Run();

record Result(string UserAgent, IEnumerable<ResultEntry> Entries);

record ResultEntry(double Average, double StandardDeviation, int NumSamples);
```


paxbun:
Isn't there a possibility that WebGL logic is not tested correctly due to lack of GPU support in CI/CD test environment?

sschmidTU:
First of all, thank you very much for creating this huge change, the performance improvements look great, and you documented and explained the changes excellently!
However, running the [visual regression tests](https://github.com/opensheetmusicdisplay/opensheetmusicdisplay/wiki/Testing#visual-regression-tests-unixmacwsl2-only-for-now), there are unfortunately a few problems/functional changes/regressions introduced by this PR currently.

The link above tells you how to run the visual regression tests. In short, you need to checkout the base commit that you created your branch off of (develop), build it and run `npm run generate:blessed` to get the 'correct' (blessed) images. Then check out your branch and run `npm run test:visual:build`, or alternatively `npm run build; npm run generate:current; npm run test:visual`, which does the same thing.

Here is an example of our default piece:
image
I'm not sure where these dips in the bottomline come from, maybe they only appear in the last and next to last "line" (system)?

- [ ] The added space should also not be there.

It's also in the Brooke - West sample:
image

Also, Debussy - Mandoline shows that it's probably not due to lyrics:
image

Bach - Air:
image

Maybe OSMD Function Test - Autobeam is a good test, very strange that the added space should be exactly there:
image

Even simpler, OSMD Function Test Container Height:
image


- [ ] In OSMD Function Test - All, the ornaments are erroneously moved:
image

- [ ] Coda is moved in Test - StaveRepetitions:
image

These issues need to be resolved before we can merge. Ideally, the visual regression tests should show no changes on your branch, or only improvements.

paxbun:
I've fixed all problems you mentioned! All image differences are under the threshold in the regression test when `PlainSkyBottomLineBatchCalculatorBackend` is used.
I've added WebGL support in the regression test script. See commit https://github.com/opensheetmusicdisplay/opensheetmusicdisplay/pull/1158/commits/0fef5662cf1c85a2f8e2bbd10af1346447c925a3 for more details.
Unfortunately, `WebGLSKyBottomLineBatchCalculatorBackend` generates slightly incorrect images. All image differences are less than 1.3,
```
test_tabs_4_strings_bass_guitar.musicxml_1 1.28376
MuzioClementi_SonatinaOpus36No1_Part1.xml_1 0.520843
Beethoven_AnDieFerneGeliebte.xml_skybottomline_1 0.421994
OSMD_function_test_chord_spacing.mxl_1 0.149897
Cornelius_P_Christbaum_Opus_8_1_1865.mxl_1 0.13735
OSMD_function_test_chord_symbols.musicxml_1 0.116578
Debussy_Mandoline.xml_1 0.0519077
...
```
and these are all because there is a one- or two-pixel difference in the resulting skylines and bottom-lines.

* `test_tabs_4_strings_bass_guitar.musicxml_1 1.28376` (I can't understand why the test script reports a difference)
test_tabs_4_strings_bass_guitar musicxml_1
* `MuzioClementi_SonatinaOpus36No1_Part1.xml_1 0.520843`
MuzioClementi_SonatinaOpus36No1_Part1 xml_1
* `Beethoven_AnDieFerneGeliebte.xml_skybottomline_1 0.421994`
Beethoven_AnDieFerneGeliebte xml_skybottomline_1

paxbun:
I've just replaced the current logic with a more precise one, and I think the result is satisfying.
```
test_tabs_4_strings_bass_guitar.musicxml_1 1.28376
Beethoven_AnDieFerneGeliebte.xml_skybottomline_1 0.0753852
OSMD_function_test_all.xml_1 0.0179828
MuzioClementi_SonatinaOpus36No1_Part1.xml_1 0.0120329
Parlez-moi.mxl_1 0.0102471
...
```

* `test_tabs_4_strings_bass_guitar.musicxml_1 1.28376` (🤔 )
test_tabs_4_strings_bass_guitar musicxml_1
* `Beethoven_AnDieFerneGeliebte.xml_skybottomline_1 0.0753852`
Beethoven_AnDieFerneGeliebte xml_skybottomline_1
* `OSMD_function_test_all.xml_1 0.0179828`
OSMD_function_test_all xml_1
* `MuzioClementi_SonatinaOpus36No1_Part1.xml_1 0.0120329`
MuzioClementi_SonatinaOpus36No1_Part1 xml_1

paxbun:
These differences are due to the low pixel value resolution of the resulting WebGL image; while the current OSMD logic requires a resolution of 0 to 299, each pixel in the WebGL image has a value in scale of 0 to 255. These differences can be completely removed by splitting 9-bit values into 1-bit values and 8-bit values as below:

* `FragmentShader.glsl`
```glsl
precision mediump float;
uniform sampler2D u_image;
varying vec4 v_position;

#define NUM_ROWS 5
#define ELEMENT_HEIGHT 300

void main() {
    const float halfPixel = 1.0 / float(ELEMENT_HEIGHT * 2);

    vec2 absolutePosition = (v_position.xy + vec2(1.0)) / vec2(2.0);
    float absX = absolutePosition.x;
    float absY = absolutePosition.y;

    int skyLine = 0;
    for (int i = 0; i < ELEMENT_HEIGHT; ++i) {
        float ratioY = float(i) / float(ELEMENT_HEIGHT);
        float relY = (ratioY - 0.5 + halfPixel) / float(NUM_ROWS);
        float x = absX;
        float y = absY + relY;

        float currentAlpha = texture2D(u_image, vec2(x, y)).a;
        if (currentAlpha > 0.0) {
            skyLine = i;
            break;
        }
    }

    int bottomLine = ELEMENT_HEIGHT;
    for (int i = ELEMENT_HEIGHT - 1; i >= 0; --i) {
        float ratioY = float(i) / float(ELEMENT_HEIGHT);
        float relY = (ratioY - 0.5 + halfPixel) / float(NUM_ROWS);
        float x = absX;
        float y = absY + relY;

        float currentAlpha = texture2D(u_image, vec2(x, y)).a;
        if (currentAlpha > 0.0) {
            bottomLine = i;
            break;
        }
    }

    int r = skyLine;
    if (r > 256) {
        r -= 256;
    }
    int g = bottomLine;
    if (g > 256) {
        g -= 256;
    }
    int b = (skyLine / 256 * 2) + (g / 256);

    gl_FragColor = vec4(float(r) / 255.0, float(g) / 255.0, float(b) / 255.0, 1.0);
}

```

* `WebGLSkyBottomLineBatchCalculatorBackend.ts`
```typescript
const r: number = pixels[x * rgbaLength + xOffset + yOffset];
const g: number = pixels[x * rgbaLength + xOffset + yOffset + 1];
const b: number = pixels[x * rgbaLength + xOffset + yOffset + 2];
const skyLinePixel: number = r + (b / 2 * 256);
const bottomLinePixel: number = g + (b % 2 * 256);
skyLine[x] = skyLinePixel;
bottomLine[x] = bottomLinePixel;
```

but are these differences must be removed despite these divisions and multiplications?

paxbun:
The criteria for deciding which algorithm (batch or non-batch) to use is set to 30 measures, but there is no basis for this, so we must do some tests before merging. See https://github.com/opensheetmusicdisplay/opensheetmusicdisplay/pull/1158/commits/aabd00d877efd5694a11266a5f19fa625633beb2 for more details.

# To-do list
- [ ] Add `generate:*` NPM scripts for WebGL tests in `package.json`
- [ ] Select a reasonable criteria for batch algorithm
- [ ] Expose `OSMD_SKY_BOTTOM_BATCH_CALCULATION_THRESHOLD` by `IOSMDOptions`

paxbun:
I tested performance for the three algorithms with varying numbers of measures, and it seems like setting the default batch threshold to 5 is a reasonable choice. I changed the default value of `EngravingRules.SkyBottomLineBatchCriteria` to 5 (It was 25). I think this PR is ready to be merged.

# Test results

The number of measures (x-axis) in the graphs below is doubled since there are two measures (the left-hand part and the right-hand part) in a single vertical measure group in the sheet music.

Microsoft Edge 99

Apple Safari 15 4

```typescript
import { IOSMDOptions, OpenSheetMusicDisplay, SkyBottomLineBatchCalculatorBackendType } from "opensheetmusicdisplay";

const url = "https://opensheetmusicdisplay.github.io/demo/MuzioClementi_SonatinaOpus36No3_Part1.xml";

function getOSMDOptions(): IOSMDOptions {
  return {
    autoResize: true,
    backend: "svg",
    drawLyricist: false,
    drawLyrics: false,
    drawFingerings: true,
    drawTitle: false,
    drawComposer: false,
    drawCredits: false,
    drawSubtitle: false,
    drawPartNames: false,
    drawPartAbbreviations: false,
    drawingParameters: "compact",
    skyBottomLineBatchCriteria: 0,
    preferredSkyBottomLineBatchCalculatorBackend: SkyBottomLineBatchCalculatorBackendType.Plain
  };
}

async function load(osmd: OpenSheetMusicDisplay, url: string, numMeasures: number) {
  const response = await fetch(url);
  const text = await response.text();
  const parser = new DOMParser();
  const content = parser.parseFromString(text, "application/xml");

  const parts = content.querySelectorAll("part");
  for (const part of parts) {
    const measures = part.querySelectorAll("measure");
    for (let i = numMeasures; i < measures.length; ++i) {
      measures[i].parentElement!.removeChild(measures[i]);
    }
  }

  await osmd.load(content);
}

function sleep(ms: number) {
  return new Promise(resolve => setTimeout(resolve, ms));
}

async function main() {
  const container = document.getElementById("container")! as HTMLDivElement;
  const options = getOSMDOptions();
  const osmd = new OpenSheetMusicDisplay(container, options);

  const samples: number[] = [];
  for (let i = 1; i <= 50; ++i)
  {
    await load(osmd, url, i);
    const start = window.performance.now();
    osmd.render();
    const end = window.performance.now();
    samples.push(end - start);
    console.log(`${i} measure${i === 1 ? "" : "s"}: ${(end - start).toFixed(2)} ms`);
    await sleep(0);
  }

  const result = `[${navigator.userAgent}] ${samples.join(",")}`;
  await fetch("C# backend URL", {
    method: "POST",
    body: result
  });

  await sleep(1000);
  location.reload();
}

window.addEventListener("load", main);
```

src/MusicalScore/Graphical/SkyBottomLine.ts
```diff
@@ -0,0 +1,509 @@
...
+     * @param end End index of the range (excluding)
+     * @param value ??
+     */
+    public updateBottomLineInRange(startIndex: number, endIndex: number, value: number): void {
+        this.updateInRange(this.BottomLine, startIndex, endIndex, value);
+    }
+
+    /**
+     * Resets a SkyLine in a range to its original value
+     * @param  to reset the SkyLine in
```
sschmidTU, line 197: remove this line

src/MusicalScore/Graphical/SkyBottomLine.ts
```diff
@@ -0,0 +1,509 @@
...
+     * @param startIndex Start index of the range
+     * @param endIndex End index of the range (excluding)
+     */
+    public resetSkyLineInRange(startIndex: number, endIndex: number): void {
+        this.updateInRange(this.SkyLine, startIndex, endIndex);
+    }
+
+    /**
+     * Resets a bottom line in a range to its original value
+     * @param  to reset the bottomline in
```
sschmidTU, line 207: remove this line

src/MusicalScore/Graphical/SkyBottomLine.ts
```diff
@@ -0,0 +1,509 @@
...
+     * @param end End index of the range
+     * @param value ??
+     */
+    public updateSkyLineInRange(startIndex: number, endIndex: number, value: number): void {
+        this.updateInRange(this.mSkyLine, startIndex, endIndex, value);
+    }
+
+    /**
+     * This method updates the BottomLine for a given range with a given value
+     * @param  to update the BottomLine for
```
sschmidTU, line 186: remove this line

src/MusicalScore/Graphical/SkyBottomLine.ts
```diff
@@ -0,0 +1,509 @@
...
+    public updateStaffLineBorders(): void {
+        this.mStaffLineParent.PositionAndShape.BorderTop = this.getSkyLineMin();
+        this.mStaffLineParent.PositionAndShape.BorderMarginTop = this.getSkyLineMin();
+        this.mStaffLineParent.PositionAndShape.BorderBottom = this.getBottomLineMax();
+        this.mStaffLineParent.PositionAndShape.BorderMarginBottom = this.getBottomLineMax();
+    }
+
+    /**
+     * This method finds the minimum value of the SkyLine.
+     * @param staffLine StaffLine to apply to
```
sschmidTU, line 271: this parameter doesn't exist

src/MusicalScore/Graphical/SkyBottomLine.ts
```diff
@@ -0,0 +1,509 @@
...
+    }
+
+    public getSkyLineMinAtPoint(point: number): number {
+        const index: number = Math.round(point * this.SamplingUnit);
+        return this.mSkyLine[index];
+    }
+
+    /**
+     * This method finds the SkyLine's minimum value within a given range.
+     * @param staffLine Staffline to apply to
```
sschmidTU, line 284: this parameter doesn't exist

src/MusicalScore/Graphical/SkyBottomLine.ts
```diff
@@ -0,0 +1,509 @@
...
+     * @param startIndex Starting index
+     * @param endIndex End index (including)
+     */
+    public getSkyLineMinInRange(startIndex: number, endIndex: number): number {
+        return this.getMinInRange(this.SkyLine, startIndex, endIndex);
+    }
+
+    /**
+     * This method finds the maximum value of the BottomLine.
+     * @param staffLine Staffline to apply to
```
sschmidTU, line 294: this parameter doesn't exist

src/MusicalScore/Graphical/SkyBottomLine.ts
```diff
@@ -0,0 +1,509 @@
...
+    }
+
+    public getBottomLineMaxAtPoint(point: number): number {
+        const index: number = Math.round(point * this.SamplingUnit);
+        return this.mBottomLine[index];
+    }
+
+    /**
+     * This method finds the BottomLine's maximum value within a given range.
+     * @param staffLine Staffline to find the max value in
```
sschmidTU, line 307: this parameter doesn't exist

src/MusicalScore/Graphical/SkyBottomLine.ts
```diff
@@ -0,0 +1,509 @@
...
+     * @returns Maximum value inside bounding box boundaries or undefined if not possible
+     */
+    public getBottomLineMaxInBoundingBox(boundingBox: BoundingBox): number {
+        //TODO: Actually it should be the margin. But that one is not implemented
+        const startPoint: number = Math.floor(boundingBox.AbsolutePosition.x + boundingBox.BorderLeft);
+        const endPoint: number = Math.ceil(boundingBox.AbsolutePosition.x + boundingBox.BorderRight);
+        return this.getMaxInRange(this.mBottomLine, startPoint, endPoint);
+    }
+
+    //#region Private methods
```
sschmidTU, line 328: move to below the next method, uncomment maybe

src/MusicalScore/Graphical/EngravingRules.ts
```diff
@@ -336,6 +341,10 @@ export class EngravingRules {
 
     public static FixStafflineBoundingBox: boolean; // TODO temporary workaround
 
+    /** skyline and bottom-line calculation preferences */
+    public PreferredSkyBottomLineBatchCalculatorBackend: SkyBottomLineBatchCalculatorBackendType;
+    public SkyBottomLineBatchCriteria: number;
```
sschmidTU, line 346: rename to SkyBottomLineBatchSize, add JSDoc description (/**)

sschmidTU (changes requested):
Thanks for the updates! Visual regression tests now show no changes and performance is still good, so that's great!

See comments for most changes needed.

Also, was it really necessary to rename the SkyBottomLineCalculator class to SkyBottomLine and create an extra calculator class? This makes the diff ugly, because about 500 of 1500 lines changed are just changing this name (refactor), so it's hard to see the real (effective) changes.
Also, in our C# source code that OSMD is based on, SkyBottomLineCalculator is still used everywhere you now use SkyBottomLine.

So, in short, I would strongly prefer if you could merge SkyBottomLine into SkyBottomLineCalculator, and keep using SkyBottomLineCalculator instead of SkyBottomLine.

src/MusicalScore/Graphical/EngravingRules.ts
paxbun: I think renaming this property to `SkyBottomLineBatchSize` is not a good choice; this is not the number of measures in a batch, this is the minimum number of measures in the entire sheet music where the batch algorithms will be enabled. Is there a better one?

paxbun:
```typescript

    /** The skyline and bottom-line calculation algorithm to use. */
    public PreferredSkyBottomLineBatchCalculatorBackend: SkyBottomLineBatchCalculatorBackendType;

    /** The minimum number of measures in the entire sheet music where the skyline and bottom-line batch calculation is enabled. */
    public <the minimum number of measures where the batch algorithm is enabled>: number;
```

sschmidTU: Ah, ok. See, I didn't understand the variable from its name. So, how about SkyBottomLineBatchMinimumMeasures?

paxbun: I think `SkyBottomLineBatchMinMeasures` is okay. I will make a commit for this.

paxbun:
I merged SkyBottomLine and SkyBottomLineCalculator and reverted all changes from renaming SkyBottomLineCalculator to SkyBottomLine, so all ugly diffs you commented on are all removed! I also removed doc comments for function parameters that do not exist now in `SkyBottomLineCalculator.ts` (Sorry for the English grammar error in the commit message https://github.com/opensheetmusicdisplay/opensheetmusicdisplay/pull/1158/commits/07158ee680f06cccf70bf5985f714912b5f38122). I also renamed `SkyBottomLineBatchCriteria` to `SkyBottomLineBatchMinMeasures` in `IOSMDOptions` and `EngravingRules`. Are there other changes needed now?

sschmidTU (approved):
@paxbun Great, thank you so much! The diff looks quite clear now (even though very large), and tests are still passing.
This looks good now, great performance improvement, and fallback to previous method (non-WebGL) if not available or on Safari where WebGL is slower (for some reason, just Apple things. I suspect this will also become faster in the future)

sschmidTU:
@infojunkie Interesting. It looks like the C++ build is failing, and to me it looks like your compiler is out of date and can't handle a newer language convention. That can also happen with JS (e.g. `??` operator).
Have you tried deleting node_modules? What's your npm version? (`npm -v`)
It works for me on a Linux Mint (which is based on Ubuntu) virtual machine, and in Windows.
I'm using node v16.13.0, maybe try that as well?

infojunkie:
Happy to try different things, but if OSMD becomes difficult to install suddenly, this feels like a regression.

sschmidTU:
I think a 30-60% performance gain is worth getting a popular npm package to work on your linux system, when it works for a lot of other people, including other linux systems.

sschmidTU:
Trying node 16 could definitely be a good idea, because node 17 is not an LTS version, so you should expect unstableness.
`nvm use 16.13.0`
if not installed:
`nvm install 16.13.0`

Also, it might be that you actually need an older version of gcc instead of a newer one.
Try this:
`CXX=gcc-10 npm i`
(you can also try gcc-9)

There is some discussion of this here (and in linked and other issues):
https://github.com/stackgl/headless-gl/issues/214#issuecomment-871154871

infojunkie:
Downgrading (not upgrading) `gcc` and `g++` to 10 was the problem, because `gl` module has a dependency that is 7 years old. Thanks.

sschmidTU:
Yes, it's definitely not a pretty situation. I'll add it to the instructions.
It would be nice if using the module was optional or you didn't have to change the `npm install` command to make it work when you have a more recent version of gcc installed, but I can't think of a way to do this.

In any case, since this only effects people who want to build osmd, there is a workaround, and it works in the linux CI and on other linux systems even without a workaround, I'm afraid this will have to be a compromise we have to go for currently, in order to have a 30-60% performance boost.

infojunkie:
Yes, great improvement in rendering time :tada:

infojunkie:
I am probably misunderstanding a lot here, but `gl` seems to be a module for node.js, not the broswer. So:

a) why is it in the `dependencies` and not the `devDependencies` ?

b) what is lost if it cannot be installed (e.g. by moving to [`optionalDependencies`](https://docs.npmjs.com/cli/v8/configuring-npm/package-json#optionaldependencies)) ?

sschmidTU:
I thought `gl` provided the WebGLRenderingContext, but that seems to not be the case, because I could run a fresh clone of osmd after immediately deleting `gl` from the package.json entirely, and `npm install` and `npm start` worked without issues. Only `npm run generate:current` of course, which uses the library directly, didn't work.

@paxbun Is it correct that the `gl` npm package is only needed for the `generateImages_browserless.mjs` script? If so, it definitely needs to be moved to the `devDependencies` section, if not to the `optionalDependencies` section. It's fine to move to `optionalDependencies` as well, the developer just needs to be aware that they can't run visual regression tests then (or generate PNGs/SVGs).

sschmidTU:
For reference, one very small (and unimportant) issue I just noticed: I ran the visual regression tests with skylines and bottomlines enabled for all samples, and it found one diff, in `OSMD_Function_Test_Drums_one_line_snare_plus_piano.musicxml`:
before:
image
after:
image
diff:
image

This is probably not important enough to try fixing. It would only change a small part of the skyline that doesn't affect anything and is invisible unless you enable the skyline. Just wanted to mention it for reference.

paxbun:
@sschmidTU I am very very sorry about this. `gl` is definitely for unit testing, so it must be in devDependencies, not in dependencies. WebGL is supported natively in all modern web browsers (even in Internet Explorer!), so using 3rd-party library for gl stuff is not needed.

sschmidTU:
No problem, thanks for confirming!

paxbun:
`gl` is hard to install as @infojunkie mentioned as well as in macOS (which I'm using now) since the Python interpreters (2 and 3) which was installed by default in previous versions is now removed in macOS Monteray 12.3, I think it should be moved to optionalDependencies and we have to modify generateImages_browserless.mjs to load `gl` dynamically so `gl` is not used when using the plain algorithms or show some error if --webgl was given and `gl` is not installed.

sschmidTU:
Ah, yes, that sounds like a very good solution! I agree, I didn't know it was difficult to install on macOS as well, so putting it under `optionalDependencies` and trying to load it dynamically in the browserless script sounds like the best solution for everyone.

sschmidTU:
@paxbun I did some performance benchmarks, and WebGL was clearly faster for me in Chrome and Edge, but clearly slower in Firefox. So I've now set the preferred backend to Plain instead of WebGL for Firefox as well.
In https://github.com/opensheetmusicdisplay/opensheetmusicdisplay/pull/1160#issuecomment-1086160794 you mentioned that non-batch was faster than batch for you on a Windows machine. Which browser did you use? Non-batch was about 3 times as slow as batch for me in Chrome, and around 10% slower in Firefox.

My benchmarks:
```js
// render Actor prelude once:
console.time();
osmd.clear();
await osmd.load("ActorPreludeSample.xml");
osmd.render();
console.timeEnd();
```

```js
// render a piece 30 times in a row (unrealistic use case):
console.time();
for (let i=1; i <= 30; i++) {
    osmd.render();
}
console.timeEnd()
```

Results (Windows):

WebGL-Chrome (Actor 1 times):
2528.541015625 ms
2816.9599609375 ms
2306.72998046875 ms
2286.68896484375 ms
2324.35302734375 ms
2347.677001953125 ms

Plain-Chrome (Actor 1 times):
4152.33203125 ms
3906.240966796875 ms
3926.009765625 ms
3952.625 ms
3905.329833984375 ms

WebGL-Chrome-nonbatch (Actor 1 times):
6858.423095703125 ms
7329.387939453125 ms
6508.819091796875 ms
6274.072265625 ms
6364.10302734375 ms

WebGL-LinuxVM-Firefox (Actor 1 times):
7585ms
7859ms
7815ms
7960ms
7968ms

Plain-LinuxVM-Firefox (Actor 1 times):
7355ms
6150ms
6266ms
6415ms
6143ms

WebGL-Chrome (Beethoven 1 times):
459.828125 ms
300.888916015625 ms
471.35205078125 ms
378.969970703125 ms
327.675048828125 ms

Plain-Chrome (Beethoven 1 times):
358.01611328125 ms
377.591796875 ms
346.84619140625 ms
340.8759765625 ms
310.9580078125 ms

WebGL-Firefox (Actor 1 times):
7053ms
7326ms
6897ms
7197ms
6981ms

Plain-Firefox (Actor 1 times):
5411ms
5306ms
5284ms
5334ms
5282ms

Plain-Firefox-nonbatch (Actor 1 times):
5682ms
6352ms
5499ms
5392ms
5883ms

WebGL-Edge (Actor 1 times):
2465.232177734375 ms
2216.76416015625 ms
2071.905029296875 ms
2005.43212890625 ms
2031.97509765625 ms

Edge-plain (Actor 1 times):
3639.065185546875 ms
3543.73583984375 ms
3614.02197265625 ms
3571.411865234375 ms
3749.271728515625 ms

Rendering the same piece 30 times in a row is slightly slower in WebGL compared to Plain, but i think that's an unrealistic use case, and might be due to the startup performance cost of creating a WebGL context (sometimes the browser also complains about too many WebGL contexts created):
WebGL-Chrome (30 renders, Beethoven Geliebte):
7173.1201171875 ms
7107.591064453125 ms
7045.744140625 ms
6982.450927734375 ms
7134.962158203125 ms

Plain-Chrome (30 renders, Beethoven Geliebte):
6159.014892578125 ms
6113.953125 ms
5474.030029296875 ms
5697.508056640625 ms
5676.974853515625 ms

This was on a Windows machine. There were other background processes with around ~18% CPU usage.

The most important use case for performance is loading one big piece once, which can take several seconds on slow machines. And in Chrome and Edge, that's significantly faster with WebGL.

paxbun:
@sschmidTU It was tested on Windows Edge 99 as well, but with only 100 measures. I don't have detailed results now, but the trend was as follows:
image

That's why I said WebGL was slower. The slope of WebGL's graph was obviously more gradual, but I thought the x-intercept was too high.

sschmidTU:
Yes, it seems like WebGL has some startup/windup cost. I like your hand-drawn graph!
But as I said, the most important performance issue in OSMD is rendering large scores not too slowly. It's important whether a large score takes 4 seconds instead of 7 seconds, but it's not so important whether a small score takes 350ms or 400ms.
That's why I would set the default to WebGL and batch for Chrome and Edge.

paxbun:
@sschmidTU

> ... maybe we start using WebGL at 60 graphical measures minimum?

I think 60 is a reasonable choice. I set this to [5](https://github.com/opensheetmusicdisplay/opensheetmusicdisplay/blob/develop/src/MusicalScore/Graphical/EngravingRules.ts#L686), but considering the test results on my Windows machine and yours, 60 is much better than 5.

> To get the number of staves ...

I thought I was counting the number of graphical measures (as in [`VexFlowMusicSheetCalculator.ts`](https://github.com/opensheetmusicdisplay/opensheetmusicdisplay/blob/develop/src/MusicalScore/Graphical/VexFlow/VexFlowMusicSheetCalculator.ts#L1060)), but does this incorrectly calculates the total number of graphical measures? BTW, are measures filled with multi-measure rests also counted as graphical measures?

> Would you be interested in writing some tests or benchmarks for performance?

I would really like to, but since I'm working on another project now, I can't guarantee that I can write benchmarks for this at the moment. I will let you know before this weekend when I finish the task right now.

sschmidTU:
I did some performance tests on a Macbook (late 2012 Pro),
basically confirming that WebGL is slower in Safari and Firefox, but it is faster in Chrome on MacOS:

Webgl Safari Actor
19082.098ms
16388.795ms
15661.948ms
14636.030ms
14693.365ms

Plain Safari Actor
14736.309ms
13940.483ms
14715.820ms
14822.173ms
14957.695ms

Plain Firefox Actor
11600ms
10208ms
9771ms
10323ms
9860ms

WebGL Firefox Actor
13191ms
12929ms
13251ms
12298ms
13142ms

WebGL Chrome (MacOS) Actor
7257.363037109375 ms
6878.738037109375 ms 6146.2353515625 ms
6081.171142578125 ms
6288.977783203125 ms

Plain Chrome (MacOS) Actor
9005.740966796875 ms
9246.75390625 ms
9706.58203125 ms
9239.13720703125 ms
9062.65869140625 ms

(WebGL is enabled by default for Chrome on MacOS, because `navigator.vendor` doesn't include Apple)

PR #1160, opened 2022-03-31
Move `gl` to `optionalDependencies`

paxbun:
# Summary
The `gl` npm package added by PR #1158 is hard to install. This PR is to move `gl` to `optionalDependencies` so the developers can install OSMD easily as before. Before merging this PR, I want to find the exact circumstances under which `gl` can be installed. In addition, I think now is the time to add instructions about the batch and the WebGL visual regression tests to [the wiki](https://github.com/opensheetmusicdisplay/opensheetmusicdisplay/wiki/Testing#visual-regression-tests-unixmacwsl2-only-for-now).

I've tested Windows 10 using the Windows Sandbox, and it seems like `gl` can be installed without Python 2 or Visual C++ (which is contrary to [the docs](https://www.npmjs.com/package/gl#system-dependencies)). I will add test results for macOS or Ubuntu as well.

(+ Update) It seems like I removed `generate:blessed` NPM script in #1158... I've just restored it.

### DIFF
--- a/package.json
+++ b/package.json
@@ -92,7 +92,6 @@
     "eslint-plugin-no-null": "^1.0.2",
     "eslint-plugin-node": "^11.1.0",
     "eslint-plugin-standard": "^4.1.0",
-    "gl": "^5.0.0",
     "html-webpack-plugin": "^5.5.0",
     "jquery": "^3.6.0",
     "jsdom": "^19.0.0",
@@ -119,6 +118,9 @@
     "webpack-dev-server": "^4.7.0",
     "webpack-merge": "^5.8.0",
     "webpack-visualizer-plugin2": "^1.0.0"
+  },
+  "optionalDependencies": {
+    "gl": "^5.0.0"
   },
   "config": {
     "commitizen": {
--- a/package.json
+++ b/package.json
@@ -33,6 +33,7 @@
     "generate:current:webgl": "node ./test/Util/generateImages_browserless.mjs ../../build ./test/data ./visual_regression/current png 0 0 allSmall --osmdtesting 0 --webgl",
     "generate:current:debug": "node ./test/Util/generateImages_browserless.mjs ../../build ./test/data ./visual_regression/current png 0 0 allSmall --debugosmdtesting",
     "generate:current:singletest": "node test/Util/generateImages_browserless.mjs ../../build ./test/data ./visual_regression/current png 0 0 ^Beethoven --osmdtestingsingle",
+    "generate:blessed": "node ./test/Util/generateImages_browserless.mjs ../../build ./test/data ./visual_regression/blessed png 0 0 allSmall --osmdtesting",
     "test:visual": "bash ./test/Util/visual_regression.sh ./visual_regression",
     "test:visual:build": "npm-run-all prebuild build:webpack generate:current test:visual",
     "test:visual:build:batch": "npm-run-all prebuild build:webpack generate:current:batch test:visual",
--- a/test/Util/generateImages_browserless.mjs
+++ b/test/Util/generateImages_browserless.mjs
@@ -1,7 +1,6 @@
 import Blob from "cross-blob";
 import FS from "fs";
 import jsdom from "jsdom";
-import headless_gl from "gl";
 import OSMD from "../../build/opensheetmusicdisplay.min.js"; // window needs to be available before we can require OSMD
 /*
   Render each OSMD sample, grab the generated images, and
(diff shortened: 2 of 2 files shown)


paxbun:
I tried to make installation of `gl` fail as @infojunkie reported, but I couldn't make it fail in any circumstances. Here's my attempts:
* Windows 10 in Windows Sandbox 
  * Installed Node.js (with/without Chocolatey)
  * Installed git
  * Cloned `paxbun/opensheetmusicdisplay`
  * `npm i`
  * `gl` had been successfully installed
* Node.js official docker image
  ```dockerfile
  FROM node:latest

  WORKDIR /
  RUN git clone https://github.com/paxbun/opensheetmusicdisplay
  
  WORKDIR /opensheetmusicdisplay
  RUN npm i

  CMD ["/bin/bash"]
  ```
* Official Ubuntu docker image
  ```dockerfile
  FROM ubuntu:latest

  RUN apt-get update
  RUN apt-get install -y git curl
  RUN curl -sL https://deb.nodesource.com/setup_16.x -o /tmp/nodesource_setup.sh
  RUN bash /tmp/nodesource_setup.sh
  RUN apt-get install -y nodejs
  
  WORKDIR /
  RUN git clone https://github.com/paxbun/opensheetmusicdisplay
  
  WORKDIR /opensheetmusicdisplay
  RUN npm i
  
  CMD ["/bin/bash"]
  ```
The only configuration where the `gl` installation fails is macOS Monterey 12.3 or newer without Python 2.

sschmidTU:
@paxbun Can't you make the installation fail when you have a newer version than gcc-10 (gcc version 10) installed as default, like gcc-11? I believe that's the problem @infojunkie had, he had to downgrade gcc (and gpp?) to version 10 on his system.

But anyways, moving gl to optionalDependencies and trying to load it dynamically in the generateImages script should be the correct workaround for people having trouble to install gl, looks very good on a first look!

(and yes, I'll update the info in the [wiki for visual regression tests](https://github.com/opensheetmusicdisplay/opensheetmusicdisplay/wiki/Testing#visual-regression-tests-unixmacwsl2-only-for-now))

paxbun:
Oh, I thought `g++-11` was installed on my system, but it wasn't so my `$env:CXX = $(which g++-11)` was not effective... I managed to reproduce the same error @infojunkie had. I think making `gl` optional is enough now. It seems like the authors of `gl` is waiting for `angle`'s owner to approve for the fix to add missing `#include <limits>`, maybe we can fork `gl` or `angle` and add the missing`#include` if this bothers too many OSMD developers, as mentioned in https://github.com/stackgl/headless-gl/issues/214.

sschmidTU:
Thanks for checking again! Yes, I think this is fine for now, being optional.

paxbun:
By the way, this PR restores the removed `generate:blessed` NPM script, so this PR should be merged as soon as possible (or you can make a commit for the restoration directly to `develop`). Are there any additional changes needed in this PR?

sschmidTU (approved):
Merging. Thanks for the PR!
I'll add a comment to the import that it's dynamic due to the optional dependency.

sschmidTU:
@paxbun by the way, I think we can now change the generate scripts in package.json to always use webgl, since it's faster and now also works if gl couldn't be installed (then it falls back to not using gl).
(I'm in the process of updating the wiki for visual regression tests)

paxbun:
Before making the batch algorithms default, I want to mention one I had to mention in #1158; it needs to be guaranteed that ornaments connecting multiple measures are not rendered in the temporary canvas even after VexFlow is updated. Do you remember when I first made the PR, some ornaments were erroneously moved? That was because the ties and the connectors were rendered between the measures as below:

* Before fix (Bach - Air):
  image
  ```typescript
  const vexFlowContext: Vex.Flow.CanvasContext = this.canvas.getContext();
  ...
  vfStaff.setX(u * elementWidth);
  vfStaff.setY((<any>vsStaff).y as number + 100 + v * elementHeight);
  ...
  measure.draw(vexFlowContext);
  ```

* After fix (Bach - Air):
  image
  ```typescript
  const vexFlowContext: Vex.Flow.CanvasContext = this.canvas.getContext();
  const context: CanvasRenderingContext2D = vexFlowContext as unknown as CanvasRenderingContext2D;
  ...
  vfStaff.setX(0);
  vfStaff.setY((<any>vsStaff).y as number + 100);
  ...
  context.translate(u * elementWidth, v * elementHeight);
  measure.draw(vexFlowContext);
  context.translate(-u * elementWidth, -v * elementHeight);
  ```
  
I haven't read the source code of VexFlow, but I think there is a reason that the authors of VexFlow made `CanvasContext` instead of using `CanvasRenderingContext2D` and restricted the use of `translate` (even though the canvas is `CanvasRenderingContext2D` in runtime), because allowing to access such functions will make some ornaments be drawn incorrectly, and the current code abuses such behavior. There already was `(<any>vsStaff).y` before I opened the PR, but wouldn't this be a big problem preventing upgrading VF's version in the future?

sschmidTU:
I'm not sure that I understand the problem. Didn't you fix all problems with batch and webgl? I found no visual diffs with either.

Also, as for this:
> ```ts
> vfStaff.setY((<any>vsStaff).y as number + 100);
> ```
If I understand correctly, that's just an offset for the Skylinecalculator from the top so that nothing is drawn above the image border.

Also, we already have a feature branch upgraded to vexflow 4 (`feat/vexflow4`), the latest vexflow version, and there are no such problems there. You can try the branch for yourself.

paxbun:
@sschmidTU I meant that if in the future version of VF the properties `y` and `translate` are removed (as these are not exposed as a public TypeScript property), these codes will be a huge issue, but now that I thinking about it, maybe this was just a premature concern. Never mind.
By the way, on my Windows 10 machine (i7-7700HQ, GTX 1070 Mobile, & 32GB RAM), the non-batch algorithm was faster when the number of measures was less than 100, contrary to the test results on my mac where the non-batch algorithm was faster only when the number of measures was less than 5-10. Maybe we can add an automated benchmarking script (which executes the code I posted in #1158) in `test/` so people can test the performance of the algorithms on as many devices as possible and post the results to github easily.

sschmidTU:
> I meant that if in the future version of VF the properties `y` and `translate` are removed (as these are not exposed as a public TypeScript property), these codes will be a huge issue, but now that I thinking about it, maybe this was just a premature concern. Never mind.

Yeah, first of all this was already used before, so no change, and second of all, this applies to many vexflow things, we'll deal with it when we get there.

> on my Windows 10 machine (i7-7700HQ, GTX 1070 Mobile, & 32GB RAM), the non-batch algorithm was faster when the number of measures was less than 100
Interesting, yeah, I should do a bit more benchmarking as well.

PR #1165, opened 2022-04-30
Improved url detection

mardax007:
I've tested it in the sponsor version and it worked


sschmidTU:
why are you deleting the line that tests for xml?

Issue #1169, opened 2022-05-16
Unofficial MusicXML Test Suite


rvilarl:
https://lilypond.org/doc/v2.20/input/regression/musicxml/collated-files#test-cases

@sschmidTU I believe that it would be good to include this test set, what do you think?

sschmidTU:
I actually looked at that at some point. But for one, we cover a lot of that already with our test samples, and secondly, the links are dead. Good idea though!

PR #1170, opened 2022-05-16
MusicXML Test Suite

rvilarl:
addresses #1169


sschmidTU:
Nice. How did you get the xmls? (official links were dead)

Still not sure what exactly we should do with this. Maybe put it into a separate folder, and do separate visual regression tests for them. A lot of these tests are basically duplicates of what we're already testing with our samples, so if we just lumped everything together, we'd get duplicate visual regressions, basically (same issues, slightly different samples)

rvilarl:
In the page https://lilypond.org/doc/v2.20/input/regression/musicxml/collated-files#test-cases you find:

### Downloading the test suite

_The complete set of MusicXML test files contained in this suite can be downloaded [here](https://lilypond.org/doc/v2.20/input/regression/musicxml/MusicXML-TestSuite-0.1.zip) as a ZIP archive._

sschmidTU:
@rvilarl Let's put this test suite into a separate folder in `test/data`, like `test/data/lilypond_test_suite`, as I said, also because it has a lot of duplicate test cases to our own, and would significantly increase the runtime of visual regression tests.
I just ran the test suite only, and it takes around 7 seconds for generating the images and 28 seconds for visual regression (and this is on a pretty fast/recent machine). This would be in addition to the current test time, as I excluded our current test samples for this run.

So for the generate and test:visual scripts, I would add an option to include the lilypond samples and disable it by default.
I can do this adaptation to the `generateImages` script so that you don't have to.
In that case you can change the target of the PR to a feature branch, and I'll merge and continue from there.

One more slight issue is that most of these tests don't have titles set in the XML, so OSMD says "Loaded sheet Untitled score" a lot. I think in this case I can also fix the `generateImages` script so that it sets the title by filename if it doesn't have an XML title.

PR #1171, opened 2022-05-17
Vexflow 4.0.3

rvilarl:
@sschmidTU this should fix some open issues, please have a look at the visual regression.

### DIFF
--- a/package.json
+++ b/package.json
@@ -71,7 +71,7 @@
     "jszip": "3.7.1",
     "loglevel": "^1.8.0",
     "typescript-collections": "^1.3.3",
-    "vexflow": "^4.0.2"
+    "vexflow": "^4.0.3"
   },
   "devDependencies": {
     "@types/chai": "^4.3.0",
--- a/src/MusicalScore/Graphical/VexFlow/VexFlowConverter.ts
+++ b/src/MusicalScore/Graphical/VexFlow/VexFlowConverter.ts
@@ -567,6 +567,8 @@ export class VexFlowConverter {
         for (let i: number = 0, len: number = numDots; i < len; ++i) {
             Dot.buildAndAttach([vfnote], {all: true});
         }
+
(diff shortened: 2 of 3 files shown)


rvilarl:
### Fixes
**Grace note colliding**
test_clef_end_measure musicxml_1
**String numbers colliding vertically**
test_grace_note_fingerings_and_strings_Ysaye_excerpt musicxml_1
**Invisible Notes, rests disappear**
OSMD_function_test_invisible_notes musicxml_1

sschmidTU:
@rvilarl Nice, almost only improvements! Merging.

One slight new issue is grace notes of invisible notes not showing, which they technically should, I think.
image
In Musescore:
image
sample: `OSMD_function_test_invisible_notes.musicxml`

This is a pretty rare and artificial use case though.

Full diff with lilypond samples added:
(1 lilypond diff, 3 OSMD sample diffs)
[diff_vexflow4.0.3.zip](https://github.com/opensheetmusicdisplay/opensheetmusicdisplay/files/8732944/diff_vexflow4.0.3.zip)

sschmidTU:
@rvilarl the issue with the visible grace note on the invisible grace not rendering is line 571 in VexFlowConverter.ts you added:
```typescript
vfnote.render_options.draw = firstNote.PrintObject;
```

I'm not sure how this works out exactly though, because if you replace it with only setting it to true if PrintObject true, the grace note renders, but not the rests in measures 3 and 5. Not sure why it works like that.
image

Previously, making a note transparent (instead of not rendering it, so you can re-enable it in SVG on the fly) was done in `VexFlowVoiceEntry:color()`. Ideally, it should still work like previously.
But again, the visible grace note on an invisible note will basically never happen, so this is not too important.

rvilarl:
The question would be. If we leave it transparent, do we still shift notes if they overlap? Now the shift of overlapping notes is not performed if the note is not to be printed.

PR #1178, opened 2022-06-18
Update Cursor.ts to use <div> instead of <img>

joshstovall:
I made a small improvement to simplify the Cursor element. It uses a `<div>` instead of creating a `<canvas>` then coverting into an ``, etc.  I'm not sure it will render much faster, but does save a few lines of code.

### DIFF
--- a/src/OpenSheetMusicDisplay/Cursor.ts
+++ b/src/OpenSheetMusicDisplay/Cursor.ts
@@ -37,7 +37,7 @@ export class Cursor {
       this.cursorElementId = `cursorImg-${id}`;
     }
 
-    const curs: HTMLElement = document.createElement("img");
+    const curs: HTMLElement = document.createElement("div");
     curs.id = this.cursorElementId;
     curs.style.position = "absolute";
     if (this.cursorOptions.follow === true) {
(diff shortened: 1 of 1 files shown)


sschmidTU:
Cool! The question is what the performance impact is (if any), have you measured that? That would be good to know. Though I would estimate that relatively it will be insignificant, because the by far largest part in performance is vexflow rendering and bounding box calculations (including getting the pixels for that).
Also, we should make sure no other functionality is affected by this.

joshstovall:
Not exactly sure on the performance impact. Just saw the `// FIXME This approach needs to be improved` comment and figured I would contribute.

joshstovall:
I still wonder why is the cursor being generated as an image in the first place? Wouldn't that take up more initial processing than just appending a div or a SVG rect, since you first have to make the dummy canvas etc.

Or is it the playback render (moving the cursor) where the performance is benchmarked? I would still think the pixel transforms required to scale the image during playback would use more processing than if it were a div.

Another reason to make this change is:

Using the default cursor, if the score's background color is anything other than white, the cursor's gradient is still white resulting in a bad look. If the cursor were a div, the user could have much more flexibility with dynamic styling. 

Just curious 2 years later, the whole cursor as an image thing often makes me scratch my head wishing it were more simply a div. This way a dev could have more options to style the cursor, maybe changing its color during playback etc.

How would you go about running benchmarks on this? I would be happy to try it out.

Issue #643, opened 2020-01-07
Spacing under staff


dedan:
Hey OSMD team,

a happy new year and thanks for this amazing library!

When I render a score using OSMD I get an unnecessary space under the bass staff. Even though I use `drawingParameters: 'compact'`. Is there any way to remove this spacing?

Please see the screenshot below for an example:

<img width="1109" alt="Screen Shot 2020-01-07 at 20 43 08" src="https://user-images.githubusercontent.com/196905/71924240-e7112900-318e-11ea-8499-b1e73cba4d6d.png">

sschmidTU:
Hey, thanks for the comment and report!
I checked this problem and it's not caused by compact mode,
the problem is that some of our Staffline bounding boxes are way too large.

Before i explain why, let me say that i could not fix it yet, but i provided a workaround that shouldn't cause trouble in most rendering cases (only on develop branch until next release):
```js
osmd.EngravingRules.FixStafflineBoundingBox = true;
// (load/render)
```

This disables adding the bounding boxes of ActivitySymbols to Stafflines.
Their bounding boxes seem to cause some problems:
![image](https://user-images.githubusercontent.com/33069673/72457825-b7ce6d80-37c7-11ea-868a-ab5244a697cc.png)

![image](https://user-images.githubusercontent.com/33069673/72455785-b8fd9b80-37c3-11ea-97d8-b82e26f28704.png)
This only seems to happen when there are two stafflines in a system (e.g. piano).
Unfortunately i haven't been able to find the deeper cause yet.

The Beethoven piece is an interesting example, because it only happens for some stafflines, most of them have a correct bounding box:
<img src="https://user-images.githubusercontent.com/33069673/72455901-eea28480-37c3-11ea-80c0-7b5336c40a7b.png" height="300">

The bounding

bneumann:
I am pretty puzzled about these boxes at the end. It seems I ported them a while back but no idea what they are good for. The see to be some click are of sorts. @matt-uib you know what the StaffLineActivitySymbol class should be doing? They are not doing anything at the moment

matt-uib:
yes they are for designed for activating e.g. playback of a staff-line, but this will all come later and can be commented out currently. but their bounding boxes seem as they shall be and it doesn't look to me as these are the problem here... they only stick out a little bit vertically.

sschmidTU:
This is now fixed without a workaround needed in #1244.

[...]

PR #1244, opened 2022-09-29
Fix page bbox, fix PageRightMargin PageLeftMargin EngravingRules for RenderSingleHorizontalStaffline

sschmidTU:
Page bounding box fixed by not adding unused StaffActivitySymbol bounding boxes (see #643).
They made it so that PageRightMargin 0 was not possible: there was always extra white space to the right of the score.

EngravingRules.PageRightMargin and PageLeftMargin are now also effective for EngravingRules.RenderSingleHorizontalStaffline.

(you may have to adjust your settings if you use this option)

Setting all margins to 0, you can now actually get the minimum possible score/SVG/canvas size:
image

previous result (OSMD 1.5.7):
image

Settings used:
```js
osmd.EngravingRules.PageLeftMargin = 0;
osmd.EngravingRules.PageRightMargin = 0;
osmd.EngravingRules.RenderSingleHorizontalStaffline = true;
osmd.EngravingRules.RenderTitle = false;
osmd.EngravingRules.RenderPartNames = false;
osmd.EngravingRules.SystemLabelsRightMargin = 0;
osmd.EngravingRules.PageTopMargin = 0;
osmd.EngravingRules.PageBottomMargin = 0;
```

To enable horizontal scrolling:
http://localhost:8000/?horizontalScrolling=1

(for Canvas backend, you may want to set the right and left margin to at least 0.2 in some cases)

no visual regressions.

### DIFF
--- a/src/MusicalScore/Graphical/BoundingBox.ts
+++ b/src/MusicalScore/Graphical/BoundingBox.ts
@@ -3,8 +3,6 @@ import {ArgumentOutOfRangeException} from "../Exceptions";
 import {PointF2D} from "../../Common/DataObjects/PointF2D";
 import {SizeF2D} from "../../Common/DataObjects/SizeF2D";
 import {RectangleF2D} from "../../Common/DataObjects/RectangleF2D";
-import { StaffLineActivitySymbol } from "./StaffLineActivitySymbol";
-import { EngravingRules } from "./EngravingRules";
 import { GraphicalObject } from "./GraphicalObject";
 
 /**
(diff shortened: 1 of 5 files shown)


sschmidTU:
Note that this caused measure bounding boxes to be wrong and thus the whole measure cursor (type 3) to be too short:
image
(Also, the title label could be offset)

this was caused by recalculating the bounding box of the page in MusicSheetCalculator.calculatePageLabels(), which was done to fix the page bounding box and SVG size not shrinking to the score size (which it did),
but for some reason (re-)calculating the bounding box of the page reset the measure bounding boxes to a wrong width, which was previously set correctly in MusicSystemBuilder.setMeasureWidth(). I assume it was set correctly there, but not in a way that sets the same width when you call calculateBoundingBox() on a measure.

This is now fixed:
image

with debug bounding box for measures:

image

PR #1266, opened 2022-11-05
Cursor, move to previous note

joshstovall:
Can now use `cursor.previous()` to move cursor to the previous note.

Similar to `cursor.next()`, but instead of moving the cursor one note forward, it moves backwards.

This allows the user to step through notes both forward and backward. 👍

This has been requested [on the OSMD Discord Server](https://discordapp.com/channels/796778104419909732/796796488331427841/1034618206842785802) and I am sharing my implementation. I'm not sure if this is the best way to achieve this? But it seems to work well.

### DIFF
--- a/src/Common/DataObjects/Fraction.ts
+++ b/src/Common/DataObjects/Fraction.ts
@@ -9,7 +9,7 @@ export class Fraction {
   private numerator: number = 0;
   private denominator: number = 1;
   private wholeValue: number = 0;
-  private realValue: number;
+  public realValue: number;
 
   /**
    * Returns the maximum of two fractions (does not clone)
--- a/src/Common/DataObjects/Fraction.ts
+++ b/src/Common/DataObjects/Fraction.ts
@@ -9,7 +9,7 @@ export class Fraction {
   private numerator: number = 0;
   private denominator: number = 1;
   private wholeValue: number = 0;
-  public realValue: number;
+  private realValue: number;
 
   /**
    * Returns the maximum of two fractions (does not clone)
--- a/src/MusicalScore/MusicParts/MusicPartManagerIterator.ts
+++ b/src/MusicalScore/MusicParts/MusicPartManagerIterator.ts
@@ -248,7 +248,7 @@ export class MusicPartManagerIterator {
         if (this.currentVoiceEntries) {
             this.currentVoiceEntries = [];
         }
-        if (this.currentTimeStamp.realValue < this.musicSheet.SourceMeasures.length) {
+        if (this.currentTimeStamp.RealValue < this.musicSheet.SourceMeasures.length) {
             this.recursiveMoveBack();
             return;
         }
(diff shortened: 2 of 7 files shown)


src/Common/DataObjects/Fraction.ts
```diff
@@ -9,7 +9,7 @@ export class Fraction {
   private numerator: number = 0;
   private denominator: number = 1;
   private wholeValue: number = 0;
-  private realValue: number;
+  public realValue: number;
```
sschmidTU, line 12: this is not necessary, you can use the public getter RealValue.

src/MusicalScore/MusicParts/MusicPartManagerIterator.ts
```diff
@@ -499,6 +522,33 @@ export class MusicPartManagerIterator {
...
         }
     }
+
+    private recursiveMoveBack(): void {
+       if (this.currentVoiceEntryIndex > 0 ) {
+        this.currentVoiceEntryIndex--;
+            const currentContainer = this.currentMeasure.VerticalSourceStaffEntryContainers[this.currentVoiceEntryIndex];
+            this.currentVoiceEntries = this.getVoiceEntries(currentContainer);
+            this.currentVerticalContainerInMeasureTimestamp = currentContainer.Timestamp;
+            this.currentTimeStamp =Fraction.plus(this.currentMeasure.AbsoluteTimestamp, this.currentVerticalContainerInMeasureTimestamp);
```
sschmidTU, line 532: space missing before Fraction

src/MusicalScore/VoiceData/SourceMeasure.ts
```diff
@@ -83,7 +83,7 @@ export class SourceMeasure {
     private staffLinkedExpressions: MultiExpression[][] = [];
     private tempoExpressions: MultiTempoExpression[] = [];
     public rehearsalExpression: RehearsalExpression;
-    private verticalSourceStaffEntryContainers: VerticalSourceStaffEntryContainer[] = [];
+    public verticalSourceStaffEntryContainers: VerticalSourceStaffEntryContainer[] = [];
```
sschmidTU, line 86: again, use public getter starting with capital letter instead.

src/MusicalScore/VoiceData/VerticalSourceStaffEntryContainer.ts
```diff
@@ -13,7 +13,7 @@ export class VerticalSourceStaffEntryContainer {
         this.parentMeasure = parentMeasure;
     }
 
-    private timestamp: Fraction;
+    public timestamp: Fraction;
```
sschmidTU, line 16: again, use public getter starting with capital letter instead.

sschmidTU (changes requested):
See changes requested, otherwise it looks great, elegant implementation, and it seems to be working without issues! Thank you very much, great functionality to have!

sschmidTU:
And if you can enable the left arrow key in the demo, that would be great, otherwise I'll add it after the PR.
Here's the code from line 486 of index.js to replace:
```typescript
        window.addEventListener("keydown", function (e) {
            var event = window.event ? window.event : e;
            // left arrow key
            if (event.keyCode === 37) {
                openSheetMusicDisplay.cursor.previous();
            }
            // right arrow key
            if (event.keyCode === 39) {
                openSheetMusicDisplay.cursor.next();
            }
        });
```

sschmidTU (approved):
Looks great now! All review comments addressed. Merging.
Again, an extremely useful feature (asked for by many users), and an elegant implementation.
Thanks a lot!

sschmidTU:
There was a bug causing an infinite loop after you advanced past the end of the sheet, then tried to go back.
Fixed in #1302.

Issue #1272, opened 2022-11-09
Unnecessary space rendered for unused lyric lines


ErdoesiT:
If there are multiple lyrics (e. g. for multiple verses), OSMD always leaves space for this lyrics, even when there is nothing to display in the current line:

Example:
[Test cases - Partitur - 03 Lyrics.txt](https://github.com/opensheetmusicdisplay/opensheetmusicdisplay/files/9971941/Test.cases.-.Partitur.-.03.Lyrics.txt)

Expected result:
![image](https://user-images.githubusercontent.com/28756388/200855810-80235c3a-47a8-4934-8e8c-86cc2c4b4140.png)

Actual result:
![image](https://user-images.githubusercontent.com/28756388/200855880-8965aca5-654e-4b55-b207-ca130b9f825f.png)

sschmidTU:
Fixed in https://github.com/opensheetmusicdisplay/opensheetmusicdisplay/pull/1284#issuecomment-1357693105

before:
<img width="386" alt="image" src="https://user-images.githubusercontent.com/33069673/208438671-8a34bfa5-e381-4522-97f5-700b2d84f9f3.png">

after:
<img width="401" alt="image" src="https://user-images.githubusercontent.com/33069673/208438849-c4831127-ec54-4f5f-8ba7-c75cb80ea1bf.png">

Note that the space is still used if on the same line:
<img width="688" alt="image" src="https://user-images.githubusercontent.com/33069673/208438814-545ed5da-54ef-4353-94e8-bcf0efb3ba45.png">

Issue #1271, opened 2022-11-09
Lyrics not rendered correctly when lyric number is not a numeric value


ErdoesiT:
According to the MusicXML specification the number attribut of the lyric element can be any string:
https://www.w3.org/2021/06/musicxml40/musicxml-reference/elements/lyric/

Dorico, for example, exports lyric translations and refrain/chorus with alphanumeric identifiers. If this identifier starts with a number, this lyric replaces the lyric line with this number, otherwise the lyrics are rendered above the line for lyric number 1.

Example:
[Test cases - Lyrics.txt](https://github.com/opensheetmusicdisplay/opensheetmusicdisplay/files/9971850/Test.cases.-.Lyrics.txt)

Expected result:
![image](https://user-images.githubusercontent.com/28756388/200853520-dff299fe-fd26-4368-9865-d629d46d9573.png)

Actual result:
![image](https://user-images.githubusercontent.com/28756388/200853592-5adf6cd2-4b30-42a4-8b26-4c7706b320e4.png)

Workaround (edited manually):
[Test cases - Lyrics Workaround.txt](https://github.com/opensheetmusicdisplay/opensheetmusicdisplay/files/9971864/Test.cases.-.Lyrics.Workaround.txt)

PR #1284, opened 2022-12-07
Improved lyrics handling

ErdoesiT:
Changed the lyrics verse numbers from type number to type string according to the MusicXML specification. This fixes a lot of issues when importing MusicXML files that were created with Dorico or Sibelius. When using Dorico, refrain/chorus and translations are now recognized and rendered in italics. Issues #1271 and #1272 are fixed now and lyrics with non-numeric identifiers are no longer rendered on top of the staffs.

### DIFF
--- a/src/MusicalScore/Graphical/MusicSheetCalculator.ts
+++ b/src/MusicalScore/Graphical/MusicSheetCalculator.ts
@@ -70,6 +70,7 @@ import { FillEmptyMeasuresWithWholeRests } from "../../OpenSheetMusicDisplay/OSM
 import { IStafflineNoteCalculator } from "../Interfaces/IStafflineNoteCalculator";
 import { GraphicalUnknownExpression } from "./GraphicalUnknownExpression";
 import { GraphicalChordSymbolContainer } from ".";
+import { LyricsEntry } from "../VoiceData/Lyrics";
 
 /**
  * Class used to do all the calculations in a MusicSheet, which in the end populates a GraphicalMusicSheet.
@@ -546,9 +547,10 @@ export abstract class MusicSheetCalculator {
      * @param staffLine
      * @param lyricVersesNumber
      */
-    protected calculateSingleStaffLineLyricsPosition(staffLine: StaffLine, lyricVersesNumber: number[]): GraphicalStaffEntry[] {
+    protected calculateSingleStaffLineLyricsPosition(staffLine: StaffLine, lyricVersesNumber: string[]): GraphicalStaffEntry[] {
         let numberOfVerses: number = 0;
         let lyricsStartYPosition: number = this.rules.StaffHeight; // Add offset to prevent collision
+        const relevantVerseNumbers: Map<string, boolean> = new Map<string, boolean>();
         const lyricsStaffEntriesList: GraphicalStaffEntry[] = [];
         const skyBottomLineCalculator: SkyBottomLineCalculator = staffLine.SkyBottomLineCalculator;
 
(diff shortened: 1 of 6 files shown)


src/MusicalScore/ScoreIO/MusicSymbolModules/LyricsReader.ts
```diff
@@ -71,28 +71,9 @@ export class LyricsReader {
...
+                            let currentLyricVerseNumber: string = "1";
                             if (lyricNode.attributes() !== undefined && lyricNode.attribute("number")) {
-                                try {
-                                    currentLyricVerseNumber = parseInt(lyricNode.attribute("number").value, 10); // usually doesn't throw error, but returns NaN
-                                } catch (err) {
-                                    errorNumberParse1 = true;
-                                }
-                                errorNumberParse1 = errorNumberParse1 || isNaN(currentLyricVerseNumber);
-                                if (errorNumberParse1) {
-                                    try { // Sibelius format: "part1verse1"
```
sschmidTU, line 84: Nice that you could simplify the code and delete the Sibelius format fix without breaking it (originally I expected it to be broken).
Though I'm not sure that having the verse number as string instead of number is always a win.
Of course it is when the xml gives verse numbers as random strings instead of numbers, like Sibelius (which is bad at XML export).

src/MusicalScore/VoiceData/Lyrics/LyricsEntry.ts
```diff
@@ -1,8 +1,9 @@
 import {LyricWord} from "./LyricsWord";
 import {VoiceEntry} from "../VoiceEntry";
+import { FontStyles } from "../../../Common";
```
sschmidTU, line 3: This should be from Common/FontStyles, since it can be ambiguous and has led to compile issues in the past. But I can also fix that in a follow-up commit.

src/MusicalScore/Graphical/MusicSheetCalculator.ts
```diff
@@ -70,6 +70,7 @@ import { FillEmptyMeasuresWithWholeRests } from "../../OpenSheetMusicDisplay/OSM
 import { IStafflineNoteCalculator } from "../Interfaces/IStafflineNoteCalculator";
 import { GraphicalUnknownExpression } from "./GraphicalUnknownExpression";
 import { GraphicalChordSymbolContainer } from ".";
+import { LyricsEntry } from "../VoiceData/Lyrics";
```
sschmidTU, line 73: should be from Lyrics/LyricsEntry, see below. I can fix this after merging.

sschmidTU (approved):
Good work! Looks good to me, except for the minor import format issues I can fix after merging.

No visual regressions.

Nice that the Sibelius format fix could be eliminated from the code, removing a lot of lines, simplifying.
I'm not 100% sure I'm happy with verse numbers being strings instead of number, but with XML export issues like Sibelius using random strings for verse numbers, that seems to be the safer approach.

sschmidTU:
Also added your sample to our test samples:
`test_lyrics_number_string_and_space_for_unused_lines.musicxml`

before:
image

after:
image

sschmidTU:
Added the sample for unused space:
`test_lyrics_unused_space_issue1272.musicxml`

before:
image

after:
image

Note that the space is still used if on the same line:
image

src/MusicalScore/ScoreIO/MusicSymbolModules/LyricsReader.ts
sschmidTU: Oh, right, if this is what the MusicXML specification says, we unfortunately need to stick with it ;) thanks for the info!

Issue #1299, opened 2023-01-20
"Uncaught ReferenceError: globalThis is not defined" on Android 


mariuskurgonas:
Hello,

Steps to reproduce:

1. Create a new clean Android project lets say use a single activity template
2. Add a simple index.html file and latest OSMD minimized library to the assets directory of the project (ofcourse create the assets folder as it does not exist by default)
3. Make index.html load the library and have a div with id as per wiki tutorial. Also initialize the OSMD and load a simple music xml file.
4. On the main activity screen, add a webview, call functions on it like enableJavaScript etc.
5. Load the HTML page from the assets using something like: file:///android_asset/OSMD/index.html
6. Run the app on simulator

The outcome is - on any simulator or device that runs at least API 29 (Android version 10+) - music sheet shows up without any problems

The issue:

When running the app on say Android 9 (API 28) - music scheet does not show up and console gives the following error:

com.example.myapplication I/chromium: [INFO:CONSOLE(2)] "Uncaught ReferenceError: globalThis is not defined", source: file:///android_asset/OSMD/opensheetmusicdisplay.min.js (2)

It seems that the Android 9 and lower does not support concept of globalThis yet. As i am not a specialized web developer - is there any way to overcome this for cases where globalThis is not found by default?

Thanks you

Marius

sschmidTU:
@mariuskurgonas Can you try this build?
[opensheetmusicdisplay.min.js_globalThis_fix.zip](https://github.com/opensheetmusicdisplay/opensheetmusicdisplay/files/10469426/opensheetmusicdisplay.min.js_globalThis_fix.zip)

I added null checks on the two occurences of globalThis we use in OSMD's code.
If the problem remains, we'll have to think about whether we can avoid globalThis used by other parts of the build process, like maybe by webpack.

mariuskurgonas:
@sschmidTU thanks for a super quick reply 

I have added this build to the project but unfortunately its exactly the same problem

sschmidTU:
Are you sure the build wasn't cached?
Can you find out the exact line of code that causes the error?
You can find source maps for the build in our [Releases](https://github.com/opensheetmusicdisplay/opensheetmusicdisplay/releases).

mariuskurgonas:
Yes, i just checked on couple of different versions of Android API 27 and API 26 so on API 27 it gives the same error:

[...]

PR #1301, opened 2023-01-20
Fix ES2017 incompatibility by not using Array.flat() from ES2019 (e.g. Android API 28)

sschmidTU:


sschmidTU:
As discussed in #1299, the current state of changes seems to regain compatibility with Android API 28 (assuming ES2018 or ES2017), so we'll merge this for now.

Issue #1348, opened 2023-03-16
Fingering Below Single Stave


chbauman:
Currently, setting the option `this.rules.FingeringPosition = 1` (`fingeringPosition = "below"` in options) does not work if there is a single stave, fingering markings will always be placed on top. 

The reason is that in `calculateFingerings()`, this option is always overwritten to "Above" because `measure.isUpperStaffOfInstrument()` returns true.

sschmidTU:
Fixed in #1349.
Thanks for the report and fix!

PR #1349, opened 2023-03-16
Better fingering position selection

chbauman:
Keeping fingering position if it was set to a fixed value.

Adresses #1348


sschmidTU (changes requested):
Good change, thanks!
Can you please do the two small refactor changes?
I can also do these after merging if you don't have the time, but it would make the PR neater and give you more credit for the commit/code lines.

chbauman:
Thanks for the feedback. I implemented the suggested changes, now it should be good to merge.

sschmidTU (approved):
Looks good now, thanks for making the changes, and contributing the fix!
Merging.

PR #1357, opened 2023-03-29
Make `new-page="yes"` create a new system when `newSystemFromXML` is `true`

PolariTOON:
Currently, with `{newPageFromXML: false, newSystemFromXML: true}` options, system breaks are only inserted when the measure does not fit in the system or when the MusicXML mandates a *system* break.
*Page* breaks defined by MusicXML are not taken into account, resulting in the measures keeping getting added to the same system if they still fit.
But usually, every new page starts with a new system so, while my understanding of the MusicXML format is low, i think `new-page="yes"` should also be considered a system break, even if `new-system="yes"` is not explicitely set on the `<print />` element.
This PR attempts to fix that.


sschmidTU:
Hmm, that's a good idea, we can add an option for this, but
1. The MusicXML standard doesn't say new-page implies new-system
https://www.w3.org/2021/06/musicxml40/musicxml-reference/elements/print/
Granted, it does say a page break should be forced, and OSMD refuses that, but that leads to 2.:
2. Always implying a system break assumes that page-breaks in MusicXML are always meaningfully inserted, which is unfortunately not the case.
Our MusicXML samples are filled with unnecessary page breaks, here is one example that makes little sense:
image
Here, the XML gives a forced new-page for measure 7. There is no reason to do a system break here.

As is OSMD's logic in general, the layout in OSMD is unavoidably different to the layout in the original music program used to export the score, and a different number of measures will fit a system, so OSMD does not respect layout instructions in the MusicXML that don't make sense for OSMD (because they only make sense with the exact original layout in the original music program exported from).

I've created a test sample with a page break after measure 1 (Musescore 3):
image

[test_pagebreak_implies_systembreak.zip](https://github.com/opensheetmusicdisplay/opensheetmusicdisplay/files/11101174/test_pagebreak_implies_systembreak.zip)

option used:
`osmd.EngravingRules.NewSystemAtXMLNewSystemAttribute = true`

reference behavior on develop:
image

With this change/option, it does a system break:
image

So, as a conclusion, I would suggest adding an option for `page break implies system break`, and only enabling this behavior when that option is set as well.
I'll make more comments in the review/files changed section.
(if possible, please also add the sample above to test/data. unzipped of course.)

src/MusicalScore/Graphical/MusicSystemBuilder.ts
```diff
@@ -135,7 +135,7 @@ export class MusicSystemBuilder {
             const measureFitsInSystem: boolean = this.currentSystemParams.currentWidth + totalMeasureWidth + nextMeasureBeginInstructionWidth < systemMaxWidth;
             const doXmlPageBreak: boolean = this.rules.NewPageAtXMLNewPageAttribute && sourceMeasure.printNewPageXml;
             const doXmlLineBreak: boolean = doXmlPageBreak || // also create new system if doing page break
-                (this.rules.NewSystemAtXMLNewSystemAttribute && sourceMeasure.printNewSystemXml);
+                (this.rules.NewSystemAtXMLNewSystemAttribute && (sourceMeasure.printNewPageXml || sourceMeasure.printNewSystemXml));
```
sschmidTU, line 138: I would prefer switching the order here, printNewSystem first, printNewPage second. Logically the same of course, but printNewSystem makes more sense so it's more intuitive to come first.

Also, since the printNewPageXml value will only be respected with the new `PageBreakImpliesSystemBreak` EngravingRule, please extract this part of the condition into its own variable, something like:
```ts
const impliedSystemBreak: boolean = sourceMeasure.printNewPageXml && this.rules.PageBreakImpliesSystemBreak;

const doXmlLineBreak: boolean = doXmlPageBreak || // also create new system if doing page break
    (this.rules.NewSystemAtXMLNewSystemAttribute && (sourceMeasure.printNewSystemXml || impliedSystemBreak));
```

sschmidTU (changes requested):
See comments for changes requested, including the new EngravingRule.
Definitely a good option to have, but should not be default behavior.

PolariTOON:
As suggested, I added a new engraving rule to control this behavior, instead of modifying an existing one.
But i was unsure about how it should interact with `newSystemFromXML` and `newPageFromXML`. Obviously, it conflicts with `newPageFromXML` (perhaps i should have made it an `enum` instead ?), but instead of also make it depend on `newSystemFromXML`, i have choosen to make the new option work no matter `newSystemFromXML`'s value.
That way, an user can also decide to force to insert new systems at `new-page="yes"` but not at `new-system="yes"` (as it is already the default behavior).

src/MusicalScore/Graphical/MusicSystemBuilder.ts
```diff
@@ -134,7 +134,9 @@ export class MusicSystemBuilder {
             }
             const measureFitsInSystem: boolean = this.currentSystemParams.currentWidth + totalMeasureWidth + nextMeasureBeginInstructionWidth < systemMaxWidth;
             const doXmlPageBreak: boolean = this.rules.NewPageAtXMLNewPageAttribute && sourceMeasure.printNewPageXml;
-            const doXmlLineBreak: boolean = doXmlPageBreak || // also create new system if doing page break
+            const impliedSystemBreak: boolean = doXmlPageBreak || // also create new system if doing page break
```
sschmidTU, line 137: an implied system break is not a page break.
this variable should be:
```typescript
const impliedSystemBreak: boolean = sourceMeasure.printNewPageXml && this.rules.PageBreakImpliesSystemBreak;
```
as I wrote in my other comment. If we have a page break, we don't need a system/line break.

src/OpenSheetMusicDisplay/OSMDOptions.ts
```diff
@@ -178,6 +178,11 @@ export interface IOSMDOptions {
      *  at different measures. So this option may result in a system break after a single measure in a system.
      */
     newSystemFromXML?: boolean;
+    /** Whether to begin a new system ("line break") when given a new page in XML ('new-page="yes"').
```
sschmidTU, line 181: when given a new page, but newPageFromXML is false.
Again, when we do a page break because the newPageFromXML is true, we don't need a system break. A page break is an implicit system break.

sschmidTU (changes requested):
Functionally this seems fine now, just style/code understanding changes requested.

src/MusicalScore/Graphical/MusicSystemBuilder.ts
sschmidTU: Wait, ignore the previous comment, this was as before and necessary. Sorry, I got confused there.

sschmidTU: Actually this is fine anyways, because logically, a page break also implies a system break anyways.

sschmidTU (approved):
Looks good to me now, I'll just run some tests and then merge.
Thanks for the contribution and implementing the requested changes!

PolariTOON:
@sschmidTU I think you got confused by the similarity of subsequent docstrings, and added a new comment for the option just below (`newPageFromXML`).

sschmidTU:
... and I forgot to stage the OSMDOptions file in the previous commit. Just proving my point ;)

sschmidTU:
Comment fixed once more. Should hopefully be the last correction ;)

Issue #1314, opened 2023-01-28
[Question] Same width for all measures ?


CorentinTh:
Hi !
Is there a way to have the same width for all the measures ?

I need to align a [single horizontal staff line](https://github.com/opensheetmusicdisplay/opensheetmusicdisplay/blob/develop/src/OpenSheetMusicDisplay/OSMDOptions.ts#L175) with custom data for each note and division, so I'm trying to make so the measures all have the same width, is this possible ?  

Thank you in advance for your time 🙏🏻

sschmidTU:
This is currently not possible.
The problem is that some measures need a lot of space, in which case you would more than double the space used for some measures unnecessarily.
I'm not sure what you're trying to do exactly, but you can simply get the position of each note and place the custom data above it, see here:
https://github.com/opensheetmusicdisplay/opensheetmusicdisplay/wiki/Exploring-the-Demo
(paragraph "Drawing overlay lines over the score and getting a note's position")

CorentinTh:
Okey thanks for the details, I'll try with the note positions

Thank you for your time 🙏🏻

sschmidTU:
Note that this (same measure width for all measures) can now be tested on the branch `feat/fixedMeasureWidth`.
But I hope you'll see that it's not a good idea.
We have to set the width of all measures to the width of the biggest measure, so small measures get very bloated.

Otherwise, measures will have less space than Vexflow needs (see `formatter.preCalculateMinTotalWidth`), and the layout will fail, creating tons of overlaps:
![image](https://user-images.githubusercontent.com/33069673/219523915-09783350-3539-4621-80c0-962920faafd2.png)

You can try this by setting `osmd.EngravingRules.FixedMeasureWidthFixedValue = 10` or other values. (this EngravingRule is just for testing)

We could merge the FixedMeasureWidth option into the main build, but I don't see where this could be reasonable when MusicXML files can contain very small and very large measures, which won't behave well with a fixed width.

CorentinTh:
Thanks you very much for the feedback and the implementation, I'll give it a try 🙏🏻 
If you dont want to merge, I'll maybe use the `fixedMeasureWidth` branch

fheyen:
@sschmidTU I would appreciate merging `feat/fixedMeasureWidth`, as some people (including myself) could find it very useful.
Examples:
- You already know the file(s) that will be displayed, such as exercises, so you know they will look better or more readable with fixed measure widths
[...]

PR #1368, opened 2023-04-18
Fixed measure width option

sschmidTU:
See #1314.

An experimental option with certain drawbacks, but multiple people have requested it.

Activate option:
`osmd.EngravingRules.FixedMeasureWidth = true`

optional/experimental/debug given width for all measures:
`osmd.EngravingRules.FixedMeasureWidthFixedValue = 5`
(if you set this too low, the Vexflow layout will break, causing overlaps. minimum width depends on score)

### DIFF
--- a/src/MusicalScore/Graphical/EngravingRules.ts
+++ b/src/MusicalScore/Graphical/EngravingRules.ts
@@ -674,7 +674,7 @@ export class EngravingRules {
         this.MultipleRestMeasureDefaultWidth = 4;
         this.MultipleRestMeasureAddKeySignature = true;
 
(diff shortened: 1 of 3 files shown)


sschmidTU:
Note that the last system will currently not always be aligned with previous systems / measure widths.
image
(Clementi 36/1/1⁾

I've left a TODO comment in MusicSystemBuilder.ts in https://github.com/opensheetmusicdisplay/opensheetmusicdisplay/pull/1368/commits/ce81d521541839df8436aba70dc13cb9401dbb57.
(this would probably need a slight refactor where we save and apply the stretch factor of the previous system)

PR #1376, opened 2023-05-16
Fix Uncaught Exception for Undefined Object referenced in VexFlowOctaveShift.setEndNote()

taiyungwang:
Reproduction steps:

1. Download the MusicXML format for this [piece](https://musescore.com/user/6662591/scores/4383881)
(edit sschmidTU: [Chopin_-_Nocturne_Op_9_No_2_E_Flat_Major.zip](https://github.com/opensheetmusicdisplay/opensheetmusicdisplay/files/11491544/Chopin_-_Nocturne_Op_9_No_2_E_Flat_Major.zip) - MuseScore needs an account to download.)

2. Add it to the list of samples in the demo
3. Select the piece
4. Exception is thrown with:
`opensheetmusicdisplay.min.js:2 Uncaught TypeError: Cannot read properties of undefined (reading 'graphicalVoiceEntries')
    at l.setEndNote (opensheetmusicdisplay.min.js:2:451999)
    at I.calculateSingleOctaveShift (opensheetmusicdisplay.min.js:2:427623)
    at I.calculateOctaveShifts (opensheetmusicdisplay.min.js:2:289252)
    at I.calculateMusicSystems (opensheetmusicdisplay.min.js:2:240987)
    at I.calculate (opensheetmusicdisplay.min.js:2:231116)
    at e.GraphicalMusicSheet.reCalculate (opensheetmusicdisplay.min.js:2:189798)
    at _.render (opensheetmusicdisplay.min.js:2:719442)
    at opensheetmusicdisplay.min.js:2:730371`

### DIFF
--- a/src/MusicalScore/Graphical/VexFlow/VexFlowOctaveShift.ts
+++ b/src/MusicalScore/Graphical/VexFlow/VexFlowOctaveShift.ts
@@ -77,6 +77,9 @@ export class VexFlowOctaveShift extends GraphicalOctaveShift {
      * @param graphicalStaffEntry the staff entry that holds the end note
      */
     public setEndNote(graphicalStaffEntry: GraphicalStaffEntry): boolean {
+        if (GraphicalStaffEntry === undefined) {
+            return false;
+        }
         // this is duplicate code from setStartNote, but if we make one general method, we add a lot of branching.
         for (const gve of graphicalStaffEntry.graphicalVoiceEntries) {
             const vve: VexFlowVoiceEntry = (gve as VexFlowVoiceEntry);
--- a/src/MusicalScore/Graphical/VexFlow/VexFlowOctaveShift.ts
+++ b/src/MusicalScore/Graphical/VexFlow/VexFlowOctaveShift.ts
@@ -77,7 +77,7 @@ export class VexFlowOctaveShift extends GraphicalOctaveShift {
      * @param graphicalStaffEntry the staff entry that holds the end note
      */
     public setEndNote(graphicalStaffEntry: GraphicalStaffEntry): boolean {
-        if (GraphicalStaffEntry === undefined) {
+        if (!graphicalStaffEntry) {
             return false;
         }
         // this is duplicate code from setStartNote, but if we make one general method, we add a lot of branching.


src/MusicalScore/Graphical/VexFlow/VexFlowOctaveShift.ts
```diff
@@ -77,6 +77,9 @@ export class VexFlowOctaveShift extends GraphicalOctaveShift {
      * @param graphicalStaffEntry the staff entry that holds the end note
      */
     public setEndNote(graphicalStaffEntry: GraphicalStaffEntry): boolean {
+        if (GraphicalStaffEntry === undefined) {
```
sschmidTU, line 80: We need to check `graphicalStaffEntry` (lowercase), not `GraphicalStaffEntry`, which is a class, not the object given as a parameter.
Also, the check can be simplified to:
```ts
if (!graphicalStaffEntry) {
```

taiyungwang: Fixed it.  Didn't realize I copied the wrong string.

sschmidTU:
First of all, thanks for the report and pull request!
I think the solution should be a bit more goal-oriented though. Instead of just not setting the end note if the end note was undefined, we should try to find an alternative end point for the octave shift. Otherwise the octave shift might never end, or not be rendered at all.
I've debugged into this, and the start measure is 30:
image
```ts
if (!lastNoteOfFirstShift) {
          console.log(`didn't find last note in measure ${lastMeasureOfFirstShift.MeasureNumber}, start measure ${startMeasure.MeasureNumber}`);
```
`didn't find last note in measure -1, start measure 30`

The end measure should be 35, according to Musescore's rendering:
image

I'll have to check what's the best solution here and why the end of the OctaveShift isn't parsed correctly from the MusicXML. Maybe the safest "solution" if the end couldn't be parsed is to let the octave shift go until the end of the current system.

sschmidTU:
This will be fixed in a more expansive manner in #1377.
Still, again, thanks for the report and workaround, it was very helpful for our solution!

PR #1377, opened 2023-05-16
Fix octaveshift end note not found if last staffline measure IsExtraGraphicalMeasure (error)

sschmidTU:
also fix octaveshift bbox width negative

evolution of PR #1376 (thanks for the report @taiyungwang)

Fixes this octave shift erroring because the last measure in the staffline is an extra graphical measure (for key/rhythm change):
image
sample:
[Chopin_-_Nocturne_Op_9_No_2_E_Flat_Major.zip](https://github.com/opensheetmusicdisplay/opensheetmusicdisplay/files/11491544/Chopin_-_Nocturne_Op_9_No_2_E_Flat_Major.zip)

The method used for finding the last measure in the staffline was naive, simply taking the last one in the staffline, which can have IsExtraGraphicalMeasure = true, no staffentries, and measure number -1.
OctaveShift needs a staffentry in the measure to determine the stopping position. (This could also be changed, then we can also stop the octaveshift at an ExtraGraphicalMeasure, would need to be tested)

Now we find the first valid measure from the end backwards.

No visual regressions, though also no diffs, which means we currently don't have a test case for this.

### DIFF
--- a/src/MusicalScore/Graphical/VexFlow/VexFlowMusicSheetCalculator.ts
+++ b/src/MusicalScore/Graphical/VexFlow/VexFlowMusicSheetCalculator.ts
@@ -957,8 +957,8 @@ export class VexFlowMusicSheetCalculator extends MusicSheetCalculator {
 
       if (endStaffLine !== startStaffLine) {
         graphicalOctaveShift.endsOnDifferentStaffLine = true;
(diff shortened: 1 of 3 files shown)


sschmidTU:
During the testing of this fix, I've found further issues in #1378.

Since that will require a lot more changes, for clarity purposes, I think it's best to merge this fix first, then continue looking at the issues in #1378, to keep the code changes for both fixes somewhat separate.

PR #1384, opened 2023-05-20
AccidentalCalculator: Fix natural accidentals not remembered within measure

### LINKED ISSUE #1383: Natural accidentals are not remembered within measure
When there is an accidental, it is not remembered for subsequent notes within the same measure.
This can leads to redundant naturals or missing sharps / flats.

As an example, the following was rendered by MuseScore:
musescore

Now when rendered by OSMD it looks as follows:

before

Which is not correct.

chbauman:
Adresses and fixes #1383

Now it looks like this:
after


sschmidTU:
Hey @chbauman, thanks for the contribution, however this change causes some duplicate natural accidentals.
Minimal reproducing sample:
image

Musescore:
image

sample:
[test_accidental_natural_repeat.zip](https://github.com/opensheetmusicdisplay/opensheetmusicdisplay/files/11547653/test_accidental_natural_repeat.zip)

There's even a "double natural" in Dichterliebe measure 9:
image

So I suspect this is rather a workaround or in any case not taking all variables into account.
In this case it seems like your change works for extra accidentals/naturals in C major key signature but not in other key signatures like D major.

The accidental display code is clearly not optimal in OSMD and there are other issues e.g. with [courtesy accidentals](https://github.com/opensheetmusicdisplay/opensheetmusicdisplay/issues?q=is%3Aissue+courtesy), and I've had a quick look before, but it looks like the accidental code needs a slightly more holistic fix.

sschmidTU:
Closing, as we already/recently fixed #1383.
Still, thank you for working on it!

PR #1385, opened 2023-05-21
ColorOptions improvement, and New color mode: ColorByInstrument

joshstovall:
I have added a new color mode `ColorByInstrument`, which allows each instrument to be it's own color. 
```js
openSheetMusicDisplay.setOptions({
    coloringMode: 3,
    coloringSetCustom: ["#ff0000", "#ffa500", "#abad00", "#008000", "#0000ff", "#4b0082", "#ee82ee", "#9e8299"],
    colorOptions: {
        stems: true,
        beams: true,
        flags: true,
        rests: false,
        slurs: true
    }
});
```
 Colors are determined by `coloringSetCustom` and will simply repeat colors if there are more instruments than colors in `coloringSetCustom` array.

Screenshot 2023-05-20 172918

# 
The new `ColorOptions` consolidates some of the color options into a single interface.

This allows for toggling color inheritance for specific graphic object classes.  In other words, you can toggle various graphic elements (`stems`, `beams`, `flags`, `rests`, `slurs`) to inherit the color from the notehead or instrument.

This means `ColorStemsLikeNoteheads`, `ColorFlags`, and `ColorBeams` are now consolidated into the `ColorOptions`  and can be referenced as `ColorOptions.stems`, `ColorOptions.flags`,  `ColorOptions.beams`.

If this system proves useful, we could add other options here such as `ties`, `dynamics`, or `lyrics` to inherit as well in the future.

# 

I'm hoping this new system can be adopted, It opens up possibilities in the future for other ColoringModes such as `ColorByPart`, `ColorByVoice`, or `ColorByKey` (could be like `AutoColor`, but shifted by key so red is not always C and the color center can modulate with the key center).

These are just some ideas, but I think the changes in this PR are a core improvement to the color system. At the very least, I find useful and hope others do too. Feel free to chime in. Thanks 🤘

### DIFF
--- a/demo/index.js
+++ b/demo/index.js
@@ -461,6 +461,15 @@ import { TransposeCalculator } from '../src/Plugins/Transpose/TransposeCalculato
             coloringEnabled: true,
             // defaultColorNotehead: "#CC0055", // try setting a default color. default is black (undefined)
             // defaultColorStem: "#BB0099",
+            coloringMode: 1,
+            coloringSetCustom: ["#ff0000", "#ffa500", "#abad00", "#008000", "#0000ff", "#4b0082", "#ee82ee", "#9e8299"],        
+            colorOptions: { // colorOptions allows for toggling color inheritance for specific graphic object classes
+                stems: false,
+                beams: false,
+                flags: false,
+                rests: false,
+                slurs: false,
+            },
 
             autoBeam: false, // try true, OSMD Function Test AutoBeam sample
             autoBeamOptions: {
--- a/demo/index.js
+++ b/demo/index.js
@@ -461,14 +461,14 @@ import { TransposeCalculator } from '../src/Plugins/Transpose/TransposeCalculato
             coloringEnabled: true,
             // defaultColorNotehead: "#CC0055", // try setting a default color. default is black (undefined)
             // defaultColorStem: "#BB0099",
-            coloringMode: 1,
+            coloringMode: 3,
             coloringSetCustom: ["#ff0000", "#ffa500", "#abad00", "#008000", "#0000ff", "#4b0082", "#ee82ee", "#9e8299"],        
             colorOptions: { // colorOptions allows for toggling color inheritance for specific graphic object classes
-                stems: false,
-                beams: false,
-                flags: false,
+                stems: true,
+                beams: true,
+                flags: true,
                 rests: false,
-                slurs: false,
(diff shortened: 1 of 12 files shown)


sschmidTU:
@joshstovall  First of all, thanks for the PR, great new option, I like your example image!
However, when I use the new ColorByInstrument option, first of all it causes an error because it expects the `osmd.EngravingRules.ColoringSetCurrent` to be specified as well. This was already the case before the PR of course, though I think we shouldn't just error here, but set the color values to default values if the user didn't provide any (even though it should of course ideally be provided).
to reproduce in the demo:
```js
osmd.EngravingRules.ColoringMode = 3; // ColorByInstrument
```
Of course, when you set it via option, you do get a warning text in the console:
```js
osmd.setOptions({ coloringMode: 3 });
// $ Uncaught Error: Invalid coloringSetCustom: With coloringModes.colorByInstrument, you have to provide a coloringSetCustom parameter (array) with at least one color.
```
It would be nice if we treated both ways of setting the coloring mode the same of course (via option or EngravingRule), maybe we could achieve that via a setter on ColoringMode, if we want to use one.

One way to set the default values: (this shouldn't be necessary)
```js
osmd.setOptions({ coloringMode: 1 }); // AutoColoring mode, sets default color values
osmd.EngravingRules.ColoringMode = 3
```

I can add the setting of default values after the PR though, you don't need to do that in the PR (though if you want to, that's also nice).

---

What I would like to change in this PR while we're at it though is the default values for this coloring mode.
If I set the default colors via the last code snippet above, the second and fourth instrument remain black:
image

Can we set the default values as in your sample image above so that all instruments are colored (non-black)?
Ideally this would extend to more than 4 instruments, e.g. the 7th instrument in Actor - Prelude is also black currently.
image

By the way, out of curiosity, where in the code does it loop / modulo around the indices if we have more than 7 instruments? I'd expect an array out of bounds error, from a quick look at the code, as I don't see the index looping back to 0 after all colors were used.

joshstovall:
Thanks for the response!  To answer your last question, the modulo happens on line 80 of src/MusicalScore/Graphical/VexFlow/VexFlowVoiceEntry.ts 

```js
const colorIndex: number = instrumentIndex % this.rules.ColoringSetCurrent.size();
```

I will look into your other comments this weekend! I may have broken something in the process, but will do some testing and push an update soon.

PR #1395, opened 2023-06-12
Lyrics Spacing: Add x-padding for short notes with long lyrics

### LINKED ISSUE #1394: Long lyrics on short notes stretch the measure too much (length/width) -> use extra paddingRight on note
When lyrics entries (words) would overlap because their notes are too close together, our system to avoid lyrics overlaps (see `VexFlowMusicSheetCalculator.calculateElongationFactorFromStaffEntries`) elongates a measure to create more space between the notes and lyrics.
This usually works well, but for short notes (quarter or less) with long lyrics (e.g. >4 characters), it makes the measures very long, which is visually unappealing.
This is because notes are given less of the measure's width the shorter they are, so to create more space between 2 very short notes, you need to elongate a measure much more than for a quarter note or so.

So I've added extra padding for these short notes with long lyrics, currently 8 pixels (modifiable by EngravingRule `LyricsXPaddingFactorForLongLyrics`), which seems to work well:

before:
image

after:
image

Measure 2 here was elongated much more before, which I always thought was visually unappealing and should be improved. Now the measure widths are much more evenly distributed. In more extreme examples, this could lead to more measures fitting on one line.

<details><summary>
Via PNG generation (visual regression tests), the Beethoven piece does fit 1 more measure in the first system/line, and takes 1 system/line less to render:
</summary>

before:
image

after:
image

</details>

**Settings**
```js
osmd.EngravingRules.LyricsUseXPaddingForShortNotes = true; // whether to enable this right-side padding or not
osmd.EngravingRules.LyricsXPaddingFactorForLongLyrics = 0.8; // default value = 8 pixels
```

Added in PR #1395

these settings can also help:
```js
osmd.EngravingRules.LyricsXPaddingFactorForLongLyrics = 2.5; // can help when there is still a lot of elongation

// for testing: disable measure elongation (also try 2 or 1.6, default 2.5)
osmd.EngravingRules.MaximumLyricsElongationFactor = 1;
osmd.setOptions({ renderSingleHorizontalStaffline: true }); // removes the stretching of measures to the system/line width

// unlikely to help:
osmd.EngravingRules.HorizontalBetweenLyricsDistance = 0;
```

sschmidTU:
closes #1394

running visual regression tests now

### DIFF
--- a/src/MusicalScore/Graphical/EngravingRules.ts
+++ b/src/MusicalScore/Graphical/EngravingRules.ts
@@ -188,6 +188,11 @@ export class EngravingRules {
      * so that the measure doesn't need to be elongated too much to avoid lyrics collisions.
      * Default 0.8 = 8 pixels */
     public LyricsXPaddingFactorForLongLyrics: number;
+    /** How wide a text needs to be to trigger lyrics padding for short notes.
+     * This is visual width, not number of characters, as e.g. 'zzz' is about as wide as 'iiii'.
+     * Default 3.3.
+     */
+    public LyricsXPaddingWidthThreshold: number;
     public VerticalBetweenLyricsDistance: number;
     public HorizontalBetweenLyricsDistance: number;
     public BetweenSyllableMaximumDistance: number;
@@ -638,6 +643,7 @@ export class EngravingRules {
         this.LyricsYMarginToBottomLine = 0.2;
         this.LyricsUseXPaddingForShortNotes = true;
         this.LyricsXPaddingFactorForLongLyrics = 0.8;
+        this.LyricsXPaddingWidthThreshold = 3.3;
         this.VerticalBetweenLyricsDistance = 0.5;
         this.HorizontalBetweenLyricsDistance = 0.2;
         this.BetweenSyllableMaximumDistance = 10.0;
--- a/src/MusicalScore/Graphical/EngravingRules.ts
+++ b/src/MusicalScore/Graphical/EngravingRules.ts
@@ -181,6 +181,13 @@ export class EngravingRules {
     public LyricsHeight: number;
     public LyricsYOffsetToStaffHeight: number;
     public LyricsYMarginToBottomLine: number;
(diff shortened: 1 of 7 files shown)


sschmidTU:
If we use number of characters instead of text width (as we used to before the following commit), we can add unnecessary padding,
e.g. Brooke West sample, measure 11 on 'ling':
before (no padding):
image

after (unnecessary padding on 'ling'):
image

sschmidTU:
Current diff:
[diff_lyrics-x-padding.zip](https://github.com/opensheetmusicdisplay/opensheetmusicdisplay/files/11728821/diff_lyrics-x-padding.zip)

Good improvement for Beethoven, also nice for a few other samples. Not every sample has less space used total, but at least the critical measures demand less space, so space is more evenly distributed across measures.
Parlez-moi does use almost 150 vertical pixels less. (2650 vs 2798)

Brooke - West Point might take slightly more space, but it also fixes some lyrics collisions (e.g. measure 13 end). This is an odd sample where the padding isn't always helpful because the 8th use so much space already. I tried optimizing for all samples but couldn't get further with Brooke in the end.

The thing is that for an ideal solution, we would need to know the x-positioning of the notes beforehand, but we don't have that when the padding is added, because it's on a vexflow note level. So part of the overlap fixing still needs to be done via measure elongation where we do have all the positioning info we need (but can't move notes anymore).

In theory, we could do a 2-pass where we draw the measure once before the main rendering (without skybottomline, which would take a lot of processing time), take the x-positioning info for notes from that, and then do the note padding based on that information (and afterwards draw again).
But this would make OSMD slightly slower even without the skybottomline calculations.

sschmidTU:
As mentioned in the issue, a great way to test all the padding, spacing and overlaps is to deactivate measure elongation and use a single horizontal line:
```js
osmd.setOptions({ renderSingleHorizontalStaffline: true });
osmd.EngravingRules.MaximumLyricsElongationFactor = 1
```

This will show what the overlaps look like before measure elongation.
image
(Brooke - West sample)

Unfortunately, the "We" before measure 13 is too short to trigger the padding.
(again, we don't have note x-position info at this point, unfortunately)

PR #1397, opened 2023-06-13
Fix chord symbol centering for short chords

### LINKED ISSUE #1396: Short Chord Symbols: Improve Centering of label
Short chord symbols like "C" (compared to e.g. "Dmaj7") can appear off-center:
before:
image

after:
image

The reason was that they are positioned at the start of the staff entry's bounding box (its x-position), which is usually slightly to the left of the notehead's left edge (due to the margin), instead of in the center of the notehead.

Now short chord symbols are pushed slightly to the right.
**This x-shift is not applied to longer chord symbols** (e.g. Eb7), because their length shifts their visual center right already, and a further shift could cause chord symbol overlaps or the start appearing too far to the right.

See the first measures of our OSMD Function Test - Chord Symbols, before the change:
image

The "G" here appeared off-center (now fixed), but the D7 and Eb7 were fine (unchanged post-fix as their width is > 2).
(the "G" has width ~1.44, the "D7" ~2.44)

post-fix:
image

Settings:
```js
osmd.EngravingRules.ChordSymbolExtraXShiftForShortChordSymbols = 0.3; // default. set to 0 to disable this new x-shift
osmd.EngravingRules.ChordSymbolExtraXShiftWidthThreshold = 2.0; // default. chords with a width larger than this (e.g. Dmaj7) do not get the new x-shift.
```

sschmidTU:
Fixes #1396


sschmidTU:
test

sschmidTU:
visual regression testing didn't report the new sample because of an error in the script (and maybe too low threshold), see #1398 (will be fixed).

sschmidTU:
Before the fix to visual regression tests in #1398, only one visual diff for this PR was shown in the tests, in `test_chord_symbols_backup_node_timestamp_1270.musicxml`.

old diffs:
[diff_chord-symbol-centering_high-threshold.zip](https://github.com/opensheetmusicdisplay/opensheetmusicdisplay/files/11740256/diff_chord-symbol-centering_high-threshold.zip)

after the fix in #1398, there are 4 visual diffs shown, which I think are all significant enough to include in the diffs:
[diff_chord_symbols_centering_low-threshold.zip](https://github.com/opensheetmusicdisplay/opensheetmusicdisplay/files/11740261/diff_chord_symbols_centering_low-threshold.zip)

for example in `test_chord_symbol_centering_short_symbols.musicxml`, the diff I originally wanted to see in the tests:
image

PR #1401, opened 2023-06-14
fix tuplets not on single beam not bracketed

### LINKED ISSUE #1400: Tuplets (e.g. triplets) with rest note or not on one beam not bracketed
Tuplets not on one beam are not bracketed by default (if not given in the XML), which can create ambiguity and is a mistake according to Gould, like the first triplet here:

image

(On first sight it might look like the first note is a full rest note, but it's actually part of the tuplet)

fIxed:
image

The second triplet is a good example of the bracket being unnecessary and not used by most scores/engravers.

This also applies to unbeamed triplets:
before:
image

after:
image

new test sample:
image

Gould - Behind Bars (THE reference on engraving) says the bracket is necessary if the notes aren't all on one single beam:

image

sschmidTU:
Fixes #1400


sschmidTU:
Now the XML value (bracket='no' or 'yes') is respected by default again. (this was broken in this PR before the most recent commit)

If we set `osmd.EngravingRules.TupletsBracketedUseXMLValue = false`:
image

with `true` (default), we get the desired rendering as before the PR:
image

sample: `test_slurs_long_steep_arc_flattening_chopin.musicxml`

sschmidTU:
Current visual diff is exactly what we wanted to change:
before:
image

after:
image

sample: `test_tuplet_bracket_necessary.musicxml` (newly added in this PR)

src/MusicalScore/Graphical/VexFlow/VexFlowMeasure.ts
```diff
@@ -965,12 +966,21 @@ export class VexFlowMeasure extends GraphicalMeasure {
         if (!this.rules.AutoBeamTabs && this.isTabMeasure) { // could also use an option tabBeams to disable beams there completely
             return;
         }
-        let notesToAutoBeam: StemmableNote[] = [];
```
sschmidTU, line 968: all the changes to autobeam generation are just to save note.NoteBeam for the auto-generated beam and also not create brackets for autogenerated tuplets if not necessary. (which needs to check note.NoteBeam)

src/MusicalScore/Graphical/VexFlow/VexFlowMeasure.ts
```diff
@@ -1135,9 +1172,29 @@ export class VexFlowMeasure extends GraphicalMeasure {
                     if (tupletStaveNotes.length > 1) {
                       const tuplet: Tuplet = tupletBuilder[0];
                       const notesOccupied: number = tuplet.Notes[0][0].NormalNotes;
-                      const bracketed: boolean = tuplet.Bracket ||
```
sschmidTU, line 1138: here's where the main code changes outside of autobeams start (don't add bracket if not necessary)

PR #1408, opened 2023-06-20
Fix short lyrics not looking vertically centered below notes. Fix dash overlap in short intervals

### LINKED ISSUE #1407: Short Lyrics Not Vertically Centered Well (look slightly left-shifted)
before:
image

after:
image

[test_lyrics_centering.zip](https://github.com/opensheetmusicdisplay/opensheetmusicdisplay/files/11802256/test_lyrics_centering.zip)

solution: shift very short lyrics slightly to the right.

New EngravingRules:
```js
osmd.EngravingRules.LyricsExtraXShiftForShortLyrics = 0.5;
osmd.EngravingRules.LyricsExtraXShiftForShortLyricsWidthThreshold = 1.4; // width of '+': 1.12, 'II': 1.33 (benefits from x-shift), 'III': 1.99 (doesn't benefit)
```

`III` (width: 1.99) already looks centered without x-shifting:
image

Of course this could be made more granular by having a gradient of x-shift between certain widths, but the difference here would be academic, maybe about 1-2 pixels in a lot of cases.

This is very similar to #1396, just for lyrics and not chord symbols.

sschmidTU:
Fixes #1407

### DIFF
--- a/src/MusicalScore/Graphical/EngravingRules.ts
+++ b/src/MusicalScore/Graphical/EngravingRules.ts
@@ -189,6 +189,12 @@ export class EngravingRules {
     public LyricsHeight: number;
     public LyricsYOffsetToStaffHeight: number;
     public LyricsYMarginToBottomLine: number;
+    /** Extra x-shift (to the right) for short lyrics to be better vertically aligned.
+     * Also see ChordSymbolExtraXShiftForShortChordSymbols, same principle, same default value.
+     */
+    public LyricsExtraXShiftForShortLyrics: number;
+    /** Threshold of the lyric entry's width below which the x-shift is applied. Default 1.4. */
+    public LyricsExtraXShiftForShortLyricsWidthThreshold: number;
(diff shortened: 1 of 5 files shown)


sschmidTU:
Visual diffs are fine now, almost all improvements.
[diff_lyrics-centering-short-notes-prcommit4_y-pos-fix.zip](https://github.com/opensheetmusicdisplay/opensheetmusicdisplay/files/11804829/diff_lyrics-centering-short-notes-prcommit4_y-pos-fix.zip)

The only slightly suboptimal change is that the dash in Beethoven Geliebte measure 4 could be slightly more to the right, but it's quite minor, and x-position is quite sensitive currently (also see Mozart/Holzer measure 11-12, Christbaum measure 9 and 11. It's hard to find a solution that distinguishes these cases and is ideal for all of them)

Also, it would be nicer, more flexible and more precise to use a GraphicalRectangle instead of a GraphicalLabel with the `-` character. But that's more of a refactor unless we need the higher precision and flexibility, as the current solution is fine.

PR #1420, opened 2023-07-06
Added ETC - Extended Transpose Calculator

ammatwain:
Hello  @sschmidTU  , it's time to secure the code in a safe place!.
There's probably still a lot of work to be done, but it's also time for you to decide on the final implementation strategies. The engine is called **ETC** ***(Equal Temperament Calculator)***, and the TransposeCalculator is called **ExtendedTransposeCalculator**.
You won't find transposition routines in ETC.
ETC primarily deals with transforming Pitch (which is a kind of vector) into a scalar value called **Comma** . The **Comma is the difference between the chromatic semitone and the diatonic semitone**. In some theories, an octave is divided into 53 commas, which already accommodate the majority of calculations. **ETC uses an octave of 77 Commas** for best calculation precision. More specifically, in the system I have imagined, a **Comma unit represents the seventy-seventh part of the octave**.

comma77fragment

Once you obtain the Comma of a Pitch, the Comma of current key, the Comma of key to transpose (for example), transposition becomes a game of subtractions/additions of Commas. After performing these operations, ETC provides routines to transform the Comma back into Pitch.
I've created another object, **TransposeOptions**, which might be the right way to pass the parameters, but it needs some consideration. The basic idea of TransposeOptions would be to interface between the user's intentions and the TransposeCalculator. Given the various transposition modes, the values to write in Sheet.Transpose might be unclear for someone who hasn't read the code. In my implementation, it is TransposeOptions that writes the correct value into the Sheet.Transpose setter (*I think I was a bit hasty in choosing the name TransposeOptions. Perhaps it would be more suitable to call it TransposeSettings*).
I have tried to comment as much as possible, but I haven't finished yet. Some function names need to be improved to be self-descriptive.
The transposition system is profoundly different from the previous one, but unlike the previous one, it's designed without compromises or simplifications.

In this pull request, I have also included my demo; essentially, it is the default one, but simplified to highlight the transposition functions. You may find duplicate sheet music which is already present in the official demo. I apologize for that oversight; I realized it only after clicking on the PR :D. In any case, it was included just to show you the TransposeOptions functions, so you can remove it if you'd like.

There's nothing stopping us from proceeding step by step. As you may have noticed, my implementation of "Transpose By Halftones" works wonderfully from **all** keys to **all** keys. I would suggest taking a gradual approach, which means replacing the body of functions in the "old" TransposeCalculator with ETC functions, BUT... to handle ONLY semitone transposition (and without using TransposeOptions). This way, you will preserve the existing set of applications that depend on OSMD, thoroughly test ETC, and think about how to best implement the new features. Then, you can add the extended transposition functionalities with a major release, like version 1.9 or 2.0.

Hey! I've been working on this tool for months, and if you decide to adopt it, I expect to be adequately rewarded, which means I want access to the extended part of OSMD  :)


sschmidTU:
Just leaving my current (unfinished) thoughts here: First of all, thank you very much for the work on transposing, it looks like this works very well now and solves some transposing issues, see https://github.com/opensheetmusicdisplay/opensheetmusicdisplay/discussions/1391 (discussion).

Of course this will need some time to review. Also, it looks like there are still changes being added recently.

By the way, I'll change the PR title to Extended Transpose Calculator now, as that's also the name of the new class currently, and it's clearer. I wasn't a big fan of the name Equal Temperament Calculator, because it doesn't reveal very well that it's about transposing or that it applies to modern temperaments / scores as well, and OSMD is per se not interested in equal temperament, which is a tuning, and doesn't necessarily change the score display. But yeah, you already changed the class name anyways.

sschmidTU:
`osmd.EngravingRules.RenderCount` was added in https://github.com/opensheetmusicdisplay/opensheetmusicdisplay/commit/1e16c6d341da5c26562f99a6aee93536d2aabd04.

sschmidTU:
Yes, transposing microtones is a headache, basically we don't support it, since we don't support microtonal or other non-standard key signatures either, not the least because Vexflow doesn't support that either.
We try to support microtones and other rare/modern score features when the required workload is reasonable, but since about 99% of our users and developers never use microtones and support in Vexflow is limited, we can't fully support them.
So If I were you, I wouldn't worry about microtonal transposition, unless you're interested in it ;)

ammatwain:
> Yes, transposing microtones is a headache, basically we don't support it, since we don't support microtonal or other non-standard key signatures either, not the least because Vexflow doesn't support that either. We try to support microtones and other rare/modern score features when the required workload is reasonable, but since about 99% of our users and developers never use microtones and support in Vexflow is limited, we can't fully support them. So If I were you, I wouldn't worry about microtonal transposition, unless you're interested in it ;)

ammatwain:
For now, I'm really interested in using only a limited set of quarter tones (often found in guitar bends), just to see how I can experiment with them a bit ;) (Sorry, I clicked the wrong button and closed the pull request...)
Ah! I recompiled everything, inserting RenderCount into the famous three functions we've been talking about these days. Now ETC is definitely less aggressive when opening the sheet music.
As usual, and for your convenience, I'm providing the link to the demo:
[https://ammatwain.github.io/extended-transpose-calculator/](https://ammatwain.github.io/extended-transpose-calculator/)

sschmidTU:
test

ammatwain:
Hello Simon, I closed the pull request because my fork is out of sync with the original OSMD. I will keep you updated.

PR #1424, opened 2023-07-10
Fix key signatures accidentals wrong for key changes in certain clefs

### LINKED ISSUE #1423: The changes in key signature appear as if they were in the TREBLE CLEF.
**The changes in key signature appear as if they were in the TREBLE CLEF.**
These tests were conducted using a simple html page with clean OSMD and **without** enabling transposition.

|*Image 1*
|:--
|The first image is taken from MuseScore. In this image, the changes in key signatures that occurred in the clefs of G, C and F appear consistent with the context of their respective clefs:
| musescore|

|*Image 2*
|:--
| The second image is produced by OSMD. In this image, the changes in key signatures that occurred in the clef of C and F **appear** as if they had occurred in the clef of G: 
|osmd2-elab|

|*Image 3*
|:--
| The third  (and interesting) image is produced with OSMD, and it's the typical situation where there is little space for the entire musical staff. The staff starts again and updates the current clef. In this case, the subsequent key signature is in the correct position, but the problem reoccurs in the next measure:
|osmd3-elab|

|*Image 4*
|:--
|The fourth image is produced by OSMD, with a modification in `MusicSystemBuilder.AddInstructionsAtMeasureBegin()` method, around line 504: Changing `if (isSystemStartMeasure}` to `if (true /* isSystemStartMeasure */ )`. The modification is intended to force the drawing of the clef at the beginning of each measure. ***In this case, everything appears correct, except for an overabundance of clef symbols.***:
|osmd4-elab|

If I have to speculate something, I speculate that in the timeline of events to be drawn, there should regularly be a "currentClef" (or something similar) that is "null|undefined" and this state should bring back into play the last active clef of the staff. However, instead of taking the last active clef in the staff, for some reason, perhaps due to ambiguous conditionals, the default clef which is certainly the "G" clef, is brought back into play.

I apologize, I haven't gained enough experience in the OSMD graphics engine to explore a solution more in-depth. I pass along this information in the hope that it may be useful to more experienced individuals.

I am attaching the test file.

***NB: In any case, the notes always appear to be represented correctly.***

[Stress_Test_Clef_and_Key.musicxml.zip](https://github.com/opensheetmusicdisplay/opensheetmusicdisplay/files/11990079/Stress_Test_Clef_and_Key.musicxml.zip)

sschmidTU:
Fixes #1423

before:
image

after:
image

(yes, it's a little... unfortunate that we didn't catch this before ;) )

before:
image

after:
image

### DIFF
--- a/src/MusicalScore/Graphical/GraphicalMeasure.ts
+++ b/src/MusicalScore/Graphical/GraphicalMeasure.ts
@@ -42,6 +42,8 @@ export abstract class GraphicalMeasure extends GraphicalObject {
 
     public parentSourceMeasure: SourceMeasure;
     public staffEntries: GraphicalStaffEntry[];
+    /** The clef of the first note of the measure (the clef the measure starts with). */
+    public InitiallyActiveClef: ClefInstruction;
     /**
      * The x-width of possibly existing: repetition start line, clef, key, rhythm.
      */
--- a/src/MusicalScore/Graphical/MusicSystemBuilder.ts
+++ b/src/MusicalScore/Graphical/MusicSystemBuilder.ts
@@ -79,6 +79,7 @@ export class MusicSystemBuilder {
                 continue; // previous measure was probably multi-rest, skip this one
             }
             for (let idx: number = 0, len: number = graphicalMeasures.length; idx < len; ++idx) {
+                // graphicalMeasures[idx].InitiallyActiveClef = this.activeClefs[idx]; // too early to know clef
                 graphicalMeasures[idx].resetLayout();
             }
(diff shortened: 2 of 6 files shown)


sschmidTU:
Surprisingly only one visual diff, which fixes a key signature change in Elite Syncopations measure 54 (also in bass clef):
image

ammatwain:
Great job, Simon!

sschmidTU:
Yeah, it's funny that this can get overlooked for so long. As a musician, I see one sharp, and I know "G major". I see two sharps, and I know "D major". I'm not even looking at the placement of the accidentals, I know which ones they should be ;)

Issue #1315, opened 2023-01-28
Alignment problems


michaelcchu:
Title: "Excerpt from Clarinet Quintet, K. 581"
[Mozart_Clarinet_Quintet_Excerpt.mxl](https://opensheetmusicdisplay.github.io/demo/Mozart_Clarinet_Quintet_Excerpt.mxl) 

[OSMD Demo](https://opensheetmusicdisplay.github.io/demo/):
<img height="400" alt="image" src="https://user-images.githubusercontent.com/50280000/215287757-e274e15e-27e2-4da5-bb8c-a21b0aac45b7.png"><img height="400" alt="image" src="https://user-images.githubusercontent.com/50280000/215288322-b057c896-1d9e-4fce-bacb-3f54c53433dc.png"><img height="400" alt="image" src="https://user-images.githubusercontent.com/50280000/215287764-411fa4e4-9605-4291-a4ea-07084548145d.png">

[Soundslice](https://www.soundslice.com/musicxml-viewer/):
<img height="400" alt="image" src="https://user-images.githubusercontent.com/50280000/215286925-e2f3edca-7995-46e3-b0c1-196efdc87def.png"><img height="400" alt="image" src="https://user-images.githubusercontent.com/50280000/215287479-be7f3213-91ae-4045-b9da-5ae785d8dfb2.png">

rvilarl:
@sschmidTU this one is resolved in VexFlow 4

sschmidTU:
Also see #797

michaelcchu:
What is VexFlow 4?

sschmidTU:
Fixed in #1427.

<img width="759" alt="image" src="https://github.com/opensheetmusicdisplay/opensheetmusicdisplay/assets/33069673/5e5006dc-c002-4b9b-883f-fe3b5e157c73">

<img width="162" alt="image" src="https://github.com/opensheetmusicdisplay/opensheetmusicdisplay/assets/33069673/1aa778a8-9560-4e62-956c-4ec80c1a03a7">

PR #1427, opened 2023-07-13
Fix note x-alignment for staves with different key signatures

### LINKED ISSUE #1425: Horizontal alignment issues with transposing parts
Here is another issue that should be fixed, and I am wondering if there is an easy fix for it.

When a score includes transposing parts (which results in parts having different key signatures), the music is not correctly aligned horizontally.

Have a look at the screenshot below, I have circled the bar where the problem is mostly shown, the notes of the first and second part are not aligned because of the different key signature:

clarShot

Similarly, look at this shot of a score with multiple instruments; again, I have circled in red where the problem is mostly shown, the "Violin piccolo in Eb" part is even going out of the bar line on the right:

brandShot

I have taken those shots from the demo page:

https://opensheetmusicdisplay.github.io/demo/

And you can try yourself with the same files from the link below:

[OSMD_transp_examples.zip](https://github.com/opensheetmusicdisplay/opensheetmusicdisplay/files/12043371/OSMD_transp_examples.zip)

All parts should be horizontally aligned from the time signature by starting from the most-far-right time signature (in the case of the last score above, aligned with the "Violin piccolo in Eb" part).

I am eager to know if there is a way to fix that or if you plan to fix that. I have spent hours trying to figure out and fix the code myself, but it looks like it'd be almost impossible for me to do it. I guess who wrote the code should be able to do it faster or point me in the right direction?

Thank you!

All the best,

sschmidTU:
Fixes #1425. #1315

before:
image

after:
image

new sample: `test_noteStartX_key_signatures.musicxml`

<details><summary>before:</summary>

image

</details>

after:
image

`Mozart_Clarinet_Quintet_Excerpt.mxl`

<details><summary>before:</summary>

image

</details>

after:
image

(sample from #1425)
The Brandenburg concerto, a very big score, is now also correctly aligned, very pleased!

<details><summary>before:</summary>

image

</details>

after:

image

edit: This fix is an extended version of the one Vexflow did in their repository:
https://github.com/0xfe/vexflow/pull/1390

In our case, this fix alone was not enough, because of how we use `formatToStave()`, which overwrote the new `x` and `start_x` values after they had been set correctly. (which was quite tricky to debug)

### DIFF
--- a/src/MusicalScore/Graphical/MusicSystemBuilder.ts
+++ b/src/MusicalScore/Graphical/MusicSystemBuilder.ts
@@ -165,13 +165,6 @@ export class MusicSystemBuilder {
                 // don't set prevMeasureEndsPart in this case! because we will loop with the same measure again.
             }
         }
-        // for (const gMeasures of this.measureList) {
-        //     const staves: VF.Stave[] = [];
-        //     for (const gMeasure of gMeasures) {
-        //         staves.push((gMeasure as VexFlowMeasure).getVFStave());
-        //     }
-        //     (VF.Stave as any).formatBegModifiers(staves);
-        // }
         if (this.currentSystemParams.systemMeasures.length > 0) {
             if (this.rules.MaxSystemToDrawNumber === this.musicSystems.length) {
                 this.finalizeCurrentSystem(this.measureList[this.measureList.length - 1], !this.rules.StretchLastSystemLine, false);
--- a/src/MusicalScore/Graphical/MusicSystemBuilder.ts
+++ b/src/MusicalScore/Graphical/MusicSystemBuilder.ts
@@ -21,9 +21,6 @@ import {MidiInstrument} from "../VoiceData/Instructions/ClefInstruction";
 import {CollectionUtil} from "../../Util/CollectionUtil";
 import {SystemLinePosition} from "./SystemLinePosition";
 import { MusicSheet } from "../MusicSheet";
-import { VexFlowMeasure } from "./VexFlow";
-import Vex from "vexflow"; // unfortunately necessary here for horizontal alignment
-import VF = Vex.Flow;
 
 export class MusicSystemBuilder {
     protected measureList: GraphicalMeasure[][];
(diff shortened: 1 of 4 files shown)


sschmidTU:
Visual regression only shows one change, the fix for Mozart Clarinet Quintet. Nice!
[diff_noteStartX-fix.zip](https://github.com/opensheetmusicdisplay/opensheetmusicdisplay/files/12043411/diff_noteStartX-fix.zip)

sschmidTU:
The `modifierX` change in stave.js was something that I needed to add compared to the way vexflow fixed it in https://github.com/0xfe/vexflow/pull/1390 because of a nasty interaction where the modifier.x was set correctly by these changes, but then overwritten in formatToStave().

sschmidTU:
Actually it's still wrong in the second system because there's no timesignature there.
So, let's fix that.
image

(Mozart clarinet quintet)

sschmidTU:
Done.
image

fablau:
This is fantastic! Is this update already in the repo?

ammatwain:
Another shot on target! Great job, @sschmidTU!

sschmidTU:
@fablau in public OSMD, yes. Or do you mean in the sponsor repository? I'm working on fixing a bug with it after merge in JustifiedMusicSystemBuilder, but I think it's actually a different issue that caused this and that needs to be rolled back. I'm working on merging it right now, should be done soon (today).

fablau:
Yeah, I meant the public OSMD, even though it'll be nice to have the same update in the sponsor one. I was asking because I tried the demo and it looks like it is still set to the previous version ;)

I'll download the latest public version then. Thank you so much again!

sschmidTU:
@fablau Oh, you mean the demo. The demo is always on the state of the last release (currently, 1.8.0). We can't do a new release every day and for every fix (if nothing else, people watching our npm releases would go crazy with being spammed).
But I'll definitely make a new release very soon as there has been one more important fix already.

Also, the fix is now merged into osmd-extended (with the extended-specific bug there fixed as well, at least for x-alignment).

fablau:
Awesome! Thank you Simon!!

PR #1428, opened 2023-07-14
Support inverted triangle notehead

### LINKED ISSUE #1418: Inverted triangle NoteHeadValue displays wrong
I have notes where one of instrument (cowbell) has a NoteHeadValue == inverted triangle.

Expected:
image

Actual:  
image

MusicXml attached 
[Money beat.txt](https://github.com/opensheetmusicdisplay/opensheetmusicdisplay/files/11903195/Money.beat.txt)

sschmidTU:
Fix #1418

image

[test_notehead_inverted_triangle_stem_up_down.zip](https://github.com/opensheetmusicdisplay/opensheetmusicdisplay/files/12051266/test_notehead_inverted_triangle_stem_up_down.zip)

Note that the x and y positioning especially for the smaller notes isn't ideal, but hard to optimize. If we e.g. shift everything one more pixel, the triangle of the small notes is not well visible, basically "eaten up" by the stem, or not connected to it. There are some intricacies of the glyph code here because this is a custom glyph that wasn't originally supported by Vexflow, so without a real deep dive into the custom glyph code, I think this is as far as we can go right now. Probably the x-shift would need to scale with note scaling.

visual diffs:
[diff_notehead_inverted_triangle.zip](https://github.com/opensheetmusicdisplay/opensheetmusicdisplay/files/12051575/diff_notehead_inverted_triangle.zip)
(only the 2 new test samples affected)

### DIFF
--- a/src/MusicalScore/Graphical/VexFlow/VexFlowConverter.ts
+++ b/src/MusicalScore/Graphical/VexFlow/VexFlowConverter.ts
@@ -207,6 +207,8 @@ export class VexFlowConverter {
                 return codeStart + "D" + codeFilled;
             case NoteHeadShape.TRIANGLE:
                 return codeStart + "T" + codeFilled;
+            case NoteHeadShape.TRIANGLE_INVERTED:
+                return codeStart + "TI";
             case NoteHeadShape.X:
                 return codeStart + "X" + codeFilled;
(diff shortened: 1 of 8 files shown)


sschmidTU:
test

PR #1429, opened 2023-07-14
Fix cursor type 3 when first measure(s) invisible

sschmidTU:
after:
image

before:
image

after:
image

before:
image

```js
osmd.Sheet.Instruments[0].Visible = false;
osmd.cursor.cursorOptions.type = 3
```


sschmidTU:
test

PR #1471, opened 2023-10-11
Update InstantaneousTempoExpression.ts

fablau:
Added extra tempo indications such as "Maestoso", "Animato" and "Allegro Assai"

### DIFF
--- a/src/MusicalScore/ScoreIO/MusicSymbolModules/ExpressionReader.ts
+++ b/src/MusicalScore/ScoreIO/MusicSymbolModules/ExpressionReader.ts
@@ -535,6 +535,13 @@ export class ExpressionReader {
                 fontStyle = FontStyles.Italic;
             }
         }
+        const fontWeightAttr: Attr = wordsNode.attribute("font-weight");
+        if (fontWeightAttr) {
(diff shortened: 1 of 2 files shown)


fablau:
I have also added the ability to parse the "font-weight" attribute to show bolded text correctly.

sschmidTU:
`<words>` nodes now being bold if given is great, thanks!
visual regression diffs:
[diff_bold-words-plus-allegrovivace-etc.zip](https://github.com/opensheetmusicdisplay/opensheetmusicdisplay/files/13221080/diff_bold-words-plus-allegrovivace-etc.zip)

Could you add/drop at least one sample to this PR that shows the changes in expressions (allegro vivace etc.)? Would be nice. Though it's also not the end of the world without it, as there are no negative changes in the visual regression.

Also, sorry for the delay, I've been a bit busy with other projects recently (a couple osmd-related as well).

fablau:
No problem Simon!

Here is an example, before and after:


Before:
before

After:
After

I am attaching the same XML sample.

Thanks again!

[maestosoTest.musicxml.zip](https://github.com/opensheetmusicdisplay/opensheetmusicdisplay/files/13221792/maestosoTest.musicxml.zip)

PR #1474, opened 2023-10-17
Improve Shortening length of measures with long lyrics on short notes

sschmidTU:
Improvement on #1394. Still not perfect, but better.

visual regression diffs:
[current-lyrics-measurelength8_PaddingFactor1.0_commit.zip](https://github.com/opensheetmusicdisplay/opensheetmusicdisplay/files/12969270/current-lyrics-measurelength8_PaddingFactor1.0_commit.zip)

West Point got shortened the most, due to less measure elongation factor needed:
image

Some samples also need slightly more space now though, due to a lower threshold for activating the padding:
image

This could be avoided by adding some checks, like whether there are no lyrics on the next note (then skip padding).

Almost all affected scores are shorter though, I think the above is the only one that's longer.

---

There's a slight issue with the png generation in the generateImages script using less padding, probably because the text is somehow measured to be shorter, but I'd categorize that as jsdom wonkiness. That's not the only imperfection in jsdom/png generation, and you can just increase `EngravingRules.LyricsXPaddingWidthThreshold` for this case.

Most visible in Parlez Moi:
image
(`quide et` are two syllables, like this, in SVG:)
image
edit: fixed in 2024:
image

also Mozart - Chloe depending on settings:
image


sschmidTU:
Actually scores can be longer if `RenderSingleHorizontalStaffline` isn't used and you don't set `osmd.EngravingRules.MaximumLyricsElongationFactor = 1.0`, which I had locally.
With `RenderSingleHorizontalStaffline` false and max factor 1.0 there can still be overlaps.
max factor 1.2 looks good, but still sometimes uses slightly more space than before.
Will need more investigation.

sschmidTU:
visual diffs:
[diff_lyrics-padding-pr-to-develop-final.zip](https://github.com/opensheetmusicdisplay/opensheetmusicdisplay/files/14067751/diff_lyrics-padding-pr-to-develop-final.zip)

Many scores got a lot shorter and more compact, because we need less measure elongation to avoid overlaps.
(see `EngravingRules.MaximumLyricsElongationFactor`).

Land der Berge took 4 systems, now 3.
Parlez-mois took 16 systems, now 12.
Beethoven Geliebte went until measure 5 in the first system and measure 11 in the next one now it's 6 and 13.

OSMD Function Test Chord symbols:
Before:
image

After:
image

the only slight downside is elongating a few quarter notes that didn't need it, because unfortunately our note to note distance calculation (or rather RelativePosition.x calculation) is sometimes incorrect.
You can see this in `OSMD_function_test_measuresToDraw...`:
before:
image

after:
image
(look at "Weit")

This also happens in measure 1 in Beethoven, see the debug info in commit 5d59b4b423ac30846fa08bbf3c7195e5e5a3d625.

PR #1483, opened 2023-11-16
Bold italic fix

fablau:
I forgot to add a check for those texts with bold and italic combined. Here is the added code check for that.


sschmidTU:
Thanks for the report and PR @fablau !
I refactored this, so I've created a new PR #1487 that replaces this one, which will be closed, but was of course a valuable basis.
This PR is also not mergable, probably because you continued working on an old non-merged branch. For future PRs, you should start from the current OSMD develop branch again instead of reusing non-merged branches, if possible.

PR #1490, opened 2023-12-12
Fix tabs x-alignment broken (no release affected)

### LINKED ISSUE #1489: Tabs not x-aligned since key/time signatures not rendered anymore (no release affected)
This issue was a regression introduced in #1467 and no release has been made since then, so it didn't affect any releases yet.

before:
image

after:
image

I tried a complex solution (on the branch `fix/tab-x-alignment`) which updates the start x-position of modifiers and notes for each vertically linked stave when the start x of another stave is updated, but it lead to visual regressions in other samples that didn't seem easy to fix:
[diff_tab-x-alignment-1.zip](https://github.com/opensheetmusicdisplay/opensheetmusicdisplay/files/13651757/diff_tab-x-alignment-1.zip)

e.g.:
image
(Mozart Clarinet Quintet)

The current solution (branch `fix/tab-x-alignment-invisible-modifiers`) is much easier and similar to what we do elsewhere: technically, we render the key and time signature, but with transparent color. The advantage here is that you can still get the SVG element for these and instantly make them appear during runtime by setting the fill/stroke to a non-transparent color.

sschmidTU:
Fixes #1489


sschmidTU:
CI tests were broken because they apparently can't handle UTF-16. reformatted to UTF-8. (this happened before)

Issue #1413, opened 2023-06-23
Display Buzz Roll Tremolo (unmeasured)


sschmidTU:
after:
<img width="266" alt="image" src="https://github.com/opensheetmusicdisplay/opensheetmusicdisplay/assets/33069673/0c1c8f9c-eadd-43e5-9117-aa748035ebf4">

thickness 0.15:
<img width="267" alt="image" src="https://github.com/opensheetmusicdisplay/opensheetmusicdisplay/assets/33069673/8f239884-2cc9-433e-aab0-565f2a1f1d45">

`osmd.EngravingRules.TremoloBuzzRollThickness = 0.15`

before:
<img width="238" alt="image" src="https://github.com/opensheetmusicdisplay/opensheetmusicdisplay/assets/33069673/a06df855-4792-4002-8270-0d3a6b90b9eb">

real-life example:
<img width="370" alt="image" src="https://github.com/opensheetmusicdisplay/opensheetmusicdisplay/assets/33069673/4a0e6703-410d-4984-92fd-c051eca6b1b0">

We render this manually (outside of Vexflow) by drawing a path (tracing the Z point by point and filling the path).

PR #1493, opened 2024-01-03
fix(Buzz Roll): Fix error in SVG export for buzz rolls

sschmidTU:
Small fix for #1413 

Looks like SVG export doesn't always have bounding boxes for the SVG elements.
(probably due to how we fake the browser environment / DOM in the generateImages script)

The error was `(...).getBbox() is not a function`


sschmidTU:
test

PR #1498, opened 2024-01-09
feat(Color): Words nodes / expressions use XML color by default.

### LINKED ISSUE #1492: Support color attr in <words> element
According to [MusicXML spec](https://www.w3.org/2021/06/musicxml40/musicxml-reference/elements/words/) document, <words> element can have an optional color attribute.  Please support color attribute.
```
<direction placement="above">
  <direction-type>
    <words color="#ff0000">text in read</words>
  </direction-type>
</direction>
```

Is interpretWords the right place to add such support?
https://github.com/opensheetmusicdisplay/opensheetmusicdisplay/blob/314f8f55dcc4b7581c3fdc2d92285fa46d8727a5/src/MusicalScore/ScoreIO/MusicSymbolModules/ExpressionReader.ts#L523

sschmidTU:
Closes #1492

### DIFF
--- a/src/MusicalScore/Graphical/EngravingRules.ts
+++ b/src/MusicalScore/Graphical/EngravingRules.ts
@@ -411,6 +411,7 @@ export class EngravingRules {
     public DynamicExpressionMaxDistance: number;
     public DynamicExpressionSpacer: number;
     public IgnoreRepeatedDynamics: boolean;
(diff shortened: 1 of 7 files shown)


sschmidTU:
No visual regressions.
Looks like our current samples don't use color for words / expressions. The new test sample remedies that.

PR #1499, opened 2024-01-09
Fix crescendo wedges for multi instrument scores

### LINKED ISSUE #1480: Problems with forks (crescendos)
Here is another issue I can't figure out. Look at the attached XML file.

Here is how MuseScore displays it, which is correct:

ForkRight

And here is how OSMD displays it:

ForkWrong

Which is clearly wrong. Look at the forks for the 2nd, 3rd, and 4th instruments. They are minimized at the beginning of the second measure. Why? No idea.

Also, why is the fork for the first instrument that low vertically? It should be aligned with the forte (F) of the next bar, ideally.

But the biggest issue here is how the 2nd, 3rd, and 4th instrument forks are badly displayed.

Thoughts are welcome!

Thanks.

[ForkProblem.musicxml.zip](https://github.com/opensheetmusicdisplay/opensheetmusicdisplay/files/13257712/ForkProblem.musicxml.zip)

sschmidTU:
Fixes #1480

before:
image

after:
image

### DIFF
--- a/src/MusicalScore/Graphical/EngravingRules.ts
+++ b/src/MusicalScore/Graphical/EngravingRules.ts
@@ -411,6 +411,7 @@ export class EngravingRules {
     public DynamicExpressionMaxDistance: number;
     public DynamicExpressionSpacer: number;
     public IgnoreRepeatedDynamics: boolean;
+    public ExpressionsUseXMLColor: boolean;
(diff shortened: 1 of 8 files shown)


sschmidTU:
No visual regressions -> added test sample `test_crescendo_multi-instrument.musicxml`

PR #1508, opened 2024-01-30
Fix multiline subtitles and composer not parsed / rendered correctly from credit-words

### LINKED ISSUE #1503: Multiline subtitles exported from MuseScore only show first line
I have exported a MusicXML file from MuseScore 4.2.0, and noticed that only the first line of the subtitles (and composer) is shown. Is this something that can be fixed, or is the way MuseScore represents multiline credits different from the specification?

[test.musicxml.zip](https://github.com/opensheetmusicdisplay/opensheetmusicdisplay/files/14030099/test.musicxml.zip)

MuseScore 4.2.0:
image
OSMD 1.8.5:
image

sschmidTU:
Fixes #1503 

This might create significant changes for users, because we now use the credit-words if given, instead of the metadata.
So users might see different composer, title and subtitle labels rendered after this change.

visual regressions:
[diff_composer-subtitle-multiline-creditwords-now-rendered.zip](https://github.com/opensheetmusicdisplay/opensheetmusicdisplay/files/14104968/diff_composer-subtitle-multiline-creditwords-now-rendered.zip)

example:
before:
image

after:
image
`BrookeWestSample.mxl`

If you want to use the previous way of parsing and displaying subtitle and composer (though I highly recommend to adapt your XMLs instead, if you now see unwanted (extra) subtitle or composer lines), you can use this deprecated option, though I would like to remove it and the old code soon:
```js
osmd.EngravingRules.SheetComposerSubtitleUseLegacyParsing = true;
```


sschmidTU:
test

sschmidTU:
Build broke from karma-webpack 5.0.1 release a few hours ago, fixed in #1509.

PR #1509, opened 2024-02-01
Fix npm test (karma-webpack): delete karma-webpack patch, as 5.0.1 works out of the box

sschmidTU:
karma-webpack 5.0.1, released a few hours ago, failed npm test, and with that, our CI builds.
It is incompatible with the patch we used for it, but it seems to work out of the box patch-less, so we can just remove our patch.
We originally added the patch because karma-webpack 5.0.0 had build issues on MacOS (with different drives/mounts), but it looks like they fixed [that](https://github.com/webpack/webpack/issues/12759) in [5.0.1](https://github.com/codymikol/karma-webpack/releases/tag/v5.0.1).
Original OSMD patch issue: #1119


sschmidTU:
test

Issue #1062, opened 2021-09-17
Cross Stave Notes Wrong x-position (shifted)


zhongyita:
some part display incorrect as below: 
![image](https://user-images.githubusercontent.com/35086572/133758877-b90314a2-166e-43ec-8477-bcd0ea04746d.png)

It happens in 6th measure of 24-short.xml which put in attach files.
[24-short.zip](https://github.com/opensheetmusicdisplay/opensheetmusicdisplay/files/7184457/24-short.zip)

sschmidTU:
This is fixed now:
![image](https://user-images.githubusercontent.com/33069673/134253361-aa750074-be15-49d9-ba53-2f0b9f16fcab.png)

Here's a simplified sample for testing:
![image](https://user-images.githubusercontent.com/33069673/134253569-bfd2deff-75f0-4474-9073-8280e8fdd7b3.png)
[cross-quarters2.zip](https://github.com/opensheetmusicdisplay/opensheetmusicdisplay/files/7206519/cross-quarters2.zip)

However, this still needs regression testing, and also, we are currently looking at a related issue with cross hand notes that may need some adjustments in the implementation.

sschmidTU:
This also improves 3 other samples, while one problem in another sample is more pronounced now.

In Dichterliebe, this improves many measures immensely, especially measure 5:
![image](https://user-images.githubusercontent.com/33069673/134258060-6bc8467c-d6c9-49bb-a1e3-de992a9412e5.png)
![image](https://user-images.githubusercontent.com/33069673/134258002-7376398b-618a-488b-8c15-c045f23970f4.png)

In Haydn - Concertante, this fixes an overlap:
![image](https://user-images.githubusercontent.com/33069673/134257125-546de3d8-44eb-47ec-a1f3-d5ca4e1bfab7.png)
![image](https://user-images.githubusercontent.com/33069673/134257185-a6f327f4-ac12-4e6e-a450-ba4d33189f50.png)

There's just one problem: The beginning of Elite Syncopations (Joplin).
It was always there, but it's more pronounced now (simplified sample):
![image](https://user-images.githubusercontent.com/33069673/134259611-1d56cca9-a14e-4b7c-9892-002eb7570616.png)
![image](https://user-images.githubusercontent.com/33069673/134259579-3c14cd3e-fb29-45e3-8a5d-265c2279b2ac.png)
The last 3 notes F, G, A should be simultaneous.
[...]

PR #1515, opened 2024-02-15
Fix tab tuplet / triplet x-alignment, tab tuplet brackets

sschmidTU:
before:
image

after:
image

tuplet brackets can be disabled like this:
```js
osmd.EngravingRules.TupletNumbersInTabs = false;
// or just show the tuplet number without the bracket:
osmd.EngravingRules.TabTupletsBracketed = false;
```

tuplet brackets are placed below by default, unless there's a placement tag in the XML.

This is somewhat related to #1062 and #1489, though it's a different cause and part of the code.


sschmidTU:
No visual regressions so far, which is explained by our previous tab test samples not having triplets.
So, I added two test samples, one with tuplet brackets below and above, one with a bend effect, to get test coverage.

PR #1516, opened 2024-02-15
SVG error fix for renderCurve()

fablau:
Added check to renderCurve() function to make sure all numbers are used to avoid SVG path error: "<path> attribute d: Expected number." That may happen when there are slurs starting and ending on the same note (often slures not connected to any note in the original source score).

I am attaching the XML file that causes the error, note the last slur definition, that's causing it. It'd be nice if OSMD could show that slur in some way going towards the end of the bar as shown in the original source score below:

slurOSMD

[testSlurOSMD.musicxml.zip](https://github.com/opensheetmusicdisplay/opensheetmusicdisplay/files/14299039/testSlurOSMD.musicxml.zip)

### DIFF
--- a/src/MusicalScore/Graphical/VexFlow/SvgVexFlowBackend.ts
+++ b/src/MusicalScore/Graphical/VexFlow/SvgVexFlowBackend.ts
@@ -190,6 +190,18 @@ export class SvgVexFlowBackend extends VexFlowBackend {
     }
 
     public renderCurve(points: PointF2D[]): Node {
+
+        //Here we want to make sure all points are numbers. If not, we will use the previous point.
+        //This avoids the "<path> attribute d: Expeced number." SVG error.
+        for (let i: number = 0; i < points.length; i++) {
+            if (isNaN(points[i].x)) {
+                points[i].x = points[i - 1].x;
+            }
+            if (isNaN(points[i].y)) {
+                points[i].y = points[i - 1].y;
+            }
+        }
+
         const node: Node = this.ctx.openGroup("curve");
         this.ctx.beginPath();
         this.ctx.moveTo(points[0].x, points[0].y);


sschmidTU:
Hey, thanks for the PR!
The code change seems to basically only remove the error though, with no rendering changes, so I would like to avoid adding a few for loops just to not print a console log ;)
But still, it's nice to have one pointer to an origin of the error in the code.
What I would prefer though is to have a fix that also fixes the rendering, or doesn't cost for loops to avoid the error message.
The message is also a hint that something is not working correctly, so before the issue is fixed, at least for developers, it's better to know when there is a (small) error somewhere.

We do have the functionality to let a slur end at the end of the staffline instead of a note, e.g. for slurs that go over multiple stafflines, so this should be possible to fix completely.

By the way, on a musical note, this sample doesn't really make sense, because the slur ends nowhere. There would be no musical difference if you removed it. I'm guessing this sample came into existence by removing some measures after this slurs.

sschmidTU:
Ah, if the slur was meant to end at a repeat (so it basically goes to the repeat beginning's first note), it makes sense.

I think the solution will be either in `calculateCurve()`, to just make the slur go to the end of the staffline (or measure), if there's no end note, like in line 303 in GraphicalSlur:
```js
// Slur continues to next StaffLine - must reach the end of current StaffLine
                endLowerLeft.x = this.staffEntries[this.staffEntries.length - 1].parentMeasure.PositionAndShape.RelativePosition.x
                    + this.staffEntries[this.staffEntries.length - 1].parentMeasure.PositionAndShape.Size.width;
```
Though that might make some slurs with erroneous info in the XML render that were supposed to end somewhere else than the end of the staffline, where we simply couldn't get the info for the endNote.

Or maybe it could be solved in SlurReader (XML parsing), which might be simpler.
(note that openSlurDict contains the slurs for which we haven't found an end (note) yet)

But overall, I think this might not be that simple, so if you could also have a look, that would be great.

fablau:
Thanks Simon for the hints. I'll look into that and see if I can find an intelligent solution.

src/MusicalScore/Graphical/VexFlow/SvgVexFlowBackend.ts
```diff
@@ -190,6 +190,18 @@ export class SvgVexFlowBackend extends VexFlowBackend {
     }
 
     public renderCurve(points: PointF2D[]): Node {
+
+        //Here we want to make sure all points are numbers. If not, we will use the previous point.
+        //This avoids the "<path> attribute d: Expeced number." SVG error.
+        for (let i: number = 0; i < points.length; i++) {
```
alexshenker, line 196: consider checking if `points.length > 1`, otherwise points[-1] is possible. And if `isNaN(points[0].x)`

sschmidTU: @alexshenker how do you get array.length === -1 on an array? I don't think I've ever seen that before.
The NaN check on .x is definitely a good idea.

sschmidTU: @alexshenker you mean if the array is undefined? I don't think that's possible within current OSMD code.
current osmd code is basically this:
```ts
const curvePointsInPixels: PointF2D[] = []; // in VexFlowMusicSheetDrawer.drawSlur()
//...
curvePoints.push(new PointF2D(...))
this.backend.renderCurve(curvePointsInPixels);
```
in above code, it's impossible for `curvePointsInPixels.length >= 0` to be false.
Maybe this can happen in your code?

alexshenker: @sschmidTU Bad phrasing on my part.
I meant when calculating `points[i - 1].x`: if `i === 0` and if `points[0]` is`NaN`, line 198 would attempt to assign
`points[0].x = points[-1].x;`
I imagine this is improbable, but wouldn't hurt to handle this edge case.

PR #1517, opened 2024-02-15
Refactor: Move enums used in OSMDOptions to common

sschmidTU:
Helps avoid a large import chain to DrawingParameters -> EngravingRules etc. when importing OSMDOptions.


sschmidTU:
test

PR #1525, opened 2024-03-18
Fix wedge (crescendo) offset not used, changing end points of wedges

### LINKED ISSUE #1477: Decrescendo / wedge / Hairpin wrong length
Hello @sschmidTU and team.

Another issue I have encountered, this time involving hairpins.

Here is how this passage should look like:

SibHari

And here is how OSMD is showing it:

wrongHari

Notice the first crescendo hairpin in the first bar. It wrongly extends until the end of the bar, and I am wondering why. I see nothing wrong in the MusicXML file (attached).

Thoughts?

[DynamicsIssue.musicxml.zip](https://github.com/opensheetmusicdisplay/opensheetmusicdisplay/files/13227885/DynamicsIssue.musicxml.zip)

sschmidTU:
fixes #1477

before:
image

after:
image

before:
image

after:
image


sschmidTU:
This new behavior can be disabled like this:
```js
osmd.EngravingRules.UseEndOffsetForExpressions = false;
```

Though it looks better and more correct in basically all of our test cases, as far as I can see.

PR #1526, opened 2024-03-21
merge osmd-public 1.8.8: fix tab note rect

sschmidTU:
Fixes black rectangle behind tab note in certain viewers for SVGs


sschmidTU:
Wrong repository

Issue #1534, opened 2024-04-05
measure width factor, global and per measure (shorten individual measures)


sschmidTU:
We added support for a custom XML attribute `osmdWidthFactor`, e.g.:
```xml
<measure number="1" osmdWidthFactor="0.7">
```
This will make measure 1 only 70% as wide as OSMD would make it otherwise.
after:
<img width="170" alt="image" src="https://github.com/opensheetmusicdisplay/opensheetmusicdisplay/assets/33069673/66c523f1-892a-4ca8-b1f9-da6eb6a70c1d">

before:
<img width="196" alt="image" src="https://github.com/opensheetmusicdisplay/opensheetmusicdisplay/assets/33069673/ea43aa25-7a26-495b-a2aa-750c7597c601">

You can also set this factor manually in SourceMeasure.widthFactor:
```js
osmd.Sheet.SourceMeasures[0].widthFactor = 0.7;
osmd.render();
```

We will also soon add a similar attribute for a global scale factor `x` that will scale all measure widths by `x`.

Note that if you make the factor too low (shrink measures too much), there might be overlaps and other layout issues, especially with lyrics. So use this with caution and at your own risk ;)

sschmidTU:
First part (measure width factor) done in #1535.
Next will be global scale factor for all measures.

sschmidTU:
Completed in PR #1536.

```js
osmd.Sheet.MeasureWidthFactor = 0.8;
osmd.render();
```

<img width="807" alt="image" src="https://github.com/opensheetmusicdisplay/opensheetmusicdisplay/assets/33069673/e4c15d3a-d04d-49d1-9c97-0dc1e86c56cd">
(you can see that a factor of 0.8 for all measures is almost too much for the end of measures 4 and 6 here. So, be cautious with the global measure width factor)

This can also be read from the MusicXML file, e.g.:
```xml
<score-partwise version="2.0" osmdMeasureWidthFactor="1.0">
```

PR #1535, opened 2024-04-05
Add SourceMeasure.widthFactor, parse osmdWidthFactor attribute from xml

sschmidTU:
Adds widthFactor for individual measures (osmdWidthFactor), first part of #1534.
(second part will be global scale factor for all measures)

e.g.:
```xml
<measure number="1" osmdWidthFactor="0.7"/>
```
This will make measure 1 only 70% as wide as OSMD would make it otherwise.
after:
image

before:
image

You can also set this factor manually in SourceMeasure.widthFactor:
```js
osmd.Sheet.SourceMeasures[0].widthFactor = 0.7;
osmd.render();
```

We will also soon add a similar attribute for a global scale factor `x` that will scale all measure widths by `x`.

Note that if you make the factor too low (shrink measures too much), there might be overlaps and other layout issues, especially with lyrics. So use this with caution and at your own risk ;)


sschmidTU:
no visual regressions.

PR #1536, opened 2024-04-05
Add Sheet.MeasureWidthFactor (scales all measure widths)

sschmidTU:
second part of #1534, similar to SourceMeasure.WidthFactor, which (additionally) scales for an individual measure.

This might be more fitting to set on GraphicalMusicSheet instead of MusicSheet, but
1) osmd.GraphicSheet doesn't exist yet at the time of xml parsing
2) measure width factor is also on SourceMeasure (.WidthFactor)


sschmidTU:
no visual regressions.

PR #1542, opened 2024-05-02
Added SVGelement property to GraphicalLine

fablau:
Added SVGelement property to GraphicalLine to be consistent with grap…hicalSlurs


sschmidTU:
Thanks, LGTM, merging.
I'll also add setting the SVGElement to the usage in `drawGlissando()`, which seems to be the only other applicable usage (there's also repeated usage in `drawSampledLine()`, but none of those GraphicalLines are saved).

sschmidTU:
Also renamed the property to SVGElement (capital E) to be consistent.

fablau:
Thank you Simon... and good catch about the missed capital E!

PR #1610, opened 2025-07-09
add support for color display for MoodExpression, UnknownExpression and ContinuousDynamicExpression

Crazy-Ryan:
PR [#1498](https://github.com/opensheetmusicdisplay/opensheetmusicdisplay/pull/1498) have add support for the font color for InstantaneousTempoExpression. The font color for some other expressions still do not take effect. 

The current PR adds support for MoodExpression, UnknownExpression and ContinuousDynamicExpression. 

Before:
image

After:
image

(I tried to add support for other expression like InstantaneousDynamicExpression as well but did not succeed).

### DIFF
--- a/src/MusicalScore/VoiceData/Expressions/MultiExpression.ts
+++ b/src/MusicalScore/VoiceData/Expressions/MultiExpression.ts
@@ -131,6 +131,14 @@ export class MultiExpression {
        }
        return fontStyle;
     }
+
+    public getColorXMLOfFirstEntry(): string {
+        let colorXML: string;
+        if (this.expressions.length) {
+            colorXML = this.expressions[0].expression.ColorXML;
+        }
+        return colorXML;
+     }
     //public getFirstEntry(staffLine: StaffLine, graphLabel: GraphicalLabel): AbstractGraphicalExpression {
     //    let indexOfFirstNotInstDynExpr: number = 0;
     //    if (this.expressions[0].expression instanceof InstantaneousDynamicExpression)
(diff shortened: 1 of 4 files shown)


src/MusicalScore/VoiceData/Expressions/MultiExpression.ts
```diff
@@ -131,6 +131,14 @@ export class MultiExpression {
        }
        return fontStyle;
     }
+
+    public getColorXMLOfFirstEntry(): string {
+        let colorXML: string;
+        if (this.expressions.length) {
```
sschmidTU, line 137: this should be length >= 1, but I'll fix that.

sschmidTU (approved):
Looks good, merging.
(except for the length -> length >= 1 change, I'll do that after merging)
Thank you for the contribution, nice addition!

src/MusicalScore/VoiceData/Expressions/MultiExpression.ts
Crazy-Ryan: I assume they are equivalent in this case except that `length >= 1` is more expressive/readable

（correct me if I am wrong, as I am not very familiar with TypeScript ）

sschmidTU: Yes, in most cases they should be functionally equivalent since 0 and undefined are usually treated as false, but that could theoretically change. And since Typescript has types, it's unclean to treat a number as a boolean.
But yeah, it's mostly a code style change, for clean TypeScript.

Crazy-Ryan: Gotcha, thank you for the explanation.

PR #1631, opened 2025-12-19
Fix: Tuplet multi-voice alignment in VexFlow rendering

rle-mino:
# Problem

Notes in different voices that should be aligned at the same timestamp were being rendered at different horizontal positions when tuplets were involved. This was particularly noticeable in complex pieces like Debussy's "Clair de Lune" where multiple voices have tuplet notes with different durations (e.g., eighth notes in one voice, quarter notes in another) but the same tuplet ratio.

## Root Cause:
  1. VexFlow was using tick fractions with different denominators for tuplet notes (e.g., 3072/6) vs non-tuplet notes (2048/1), preventing proper alignment
  2. More critically, VexFlow calculated tick values based on note type (eighth = 2048 ticks) rather than the actual duration from MusicXML, causing misalignment when different voices had tuplet notes with the same start timestamp but different visual durations

# Solution

## Modified /src/MusicalScore/Graphical/VexFlow/VexFlowMeasure.ts (lines 1405-1427) to:

1. Normalize tick denominators: Convert all tick fractions to denominator of 1 (e.g., 3072/6 → 3072/1)
2. Recalculate tuplet note ticks: For notes that are part of a tuplet, calculate tick values based on their actual graphical note length from MusicXML using VF.RESOLUTION rather than relying on VexFlow's default tick calculation

```typescript
// Calculate ticks from the actual graphical note length
// graphicalLength.RealValue is the note length as a fraction of a whole note (e.g., 0.5 for half note)
// VF.RESOLUTION is the number of ticks for a whole note
const graphicalLength: Fraction = voiceEntry.notes[0].graphicalNoteLength;
const noteTicks: number = Math.round(graphicalLength.RealValue * VF.RESOLUTION);
ticks.numerator = noteTicks;
ticks.denominator = 1;
```
This ensures that all notes at the same timestamp have their x-positions correctly aligned by VexFlow's formatter, regardless of whether they're in different voices or have different durations.

## Testing

Added comprehensive test suite in /test/MusicalScore/Graphical/VexFlow/VexFlowMeasure_TupletAlignment_Test.ts with 4 tests:

- ✅ Tick denominator normalization for tuplet notes
- ✅ Correct tick calculation based on graphical length
- ✅ Voice alignment at same timestamps
- ✅ Handling of different normal-type values (16th, 32nd)

Also created test MusicXML file: /test/data/test_tuplet_multivoice_alignment.musicxml

All tests pass, including existing regression tests.

### Before
Screenshot 2025-12-19 at 21 46 32

> Before: Notes that should be vertically aligned appear at different horizontal positions

### After
Screenshot 2025-12-19 at 21 45 16

> After: Notes at the same timestamp are properly aligned across all voices

## Files Changed

- src/MusicalScore/Graphical/VexFlow/VexFlowMeasure.ts - Core fix
- test/MusicalScore/Graphical/VexFlow/VexFlowMeasure_TupletAlignment_Test.ts - Test suite (new)
- test/data/test_tuplet_multivoice_alignment.musicxml - Test data (new)

### DIFF
--- a/src/MusicalScore/Graphical/VexFlow/VexFlowMeasure.ts
+++ b/src/MusicalScore/Graphical/VexFlow/VexFlowMeasure.ts
@@ -1402,29 +1402,28 @@ export class VexFlowMeasure extends GraphicalMeasure {
                     // if denominator === 0, addTickable() below goes into an infinite loop.
                     // continue; // previous solution, but can lead to valid notes skipped, further problems, see #1073
                 }
-                // Normalize tick denominator to 1 and apply correct tuplet-based tick values
-                // VexFlow's formatter needs all voices to have denominator=1 to align notes correctly
-                const ticks: VF.Fraction = vexFlowVoiceEntry.vfStaveNote.getTicks();
-                if (ticks.denominator !== 1 && ticks.denominator !== 0) {
-                    const normalizedValue: number = ticks.value();
-                    ticks.numerator = normalizedValue;
-                    ticks.denominator = 1;
-                }
 
-                // For tuplet notes, VexFlow calculates ticks based on the note type (e.g., eighth = 2048),
-                // but we need to use the actual duration from MusicXML to ensure proper alignment
+                // Fix tick values for tuplet notes to ensure cross-voice alignment.
+                // VexFlow calculates ticks based on note type and time-modification, but when
+                // different voices use different normal-type values in their tuplets, the calculated
+                // ticks can differ even for notes at the same timestamp. We use graphicalNoteLength
+                // (which represents the actual musical duration) to calculate correct tick values.
                 if (voiceEntry.notes.length > 0 && voiceEntry.notes[0].sourceNote) {
                     const sourceNote: Note = voiceEntry.notes[0].sourceNote;
                     if (sourceNote.NoteTuplet) {
-                        // Calculate ticks from the actual graphical note length
-                        // graphicalLength.RealValue is the note length as a fraction of a whole note (e.g., 0.5 for half note)
-                        // VF.RESOLUTION is the number of ticks for a whole note
                         const graphicalLength: Fraction = voiceEntry.notes[0].graphicalNoteLength;
-                        const noteTicks: number = Math.round(graphicalLength.RealValue * VF.RESOLUTION);
-                        ticks.numerator = noteTicks;
-                        ticks.denominator = 1;
+                        // Calculate ticks using VexFlow Fraction to preserve precision.
+                        // graphicalLength.RealValue is the note length as a fraction of a whole note.
+                        // VF.RESOLUTION (e.g., 16384) is the number of ticks for a whole note.
+                        // We use Fraction arithmetic to avoid floating-point precision issues.
+                        const vfTicks: VF.Fraction = vexFlowVoiceEntry.vfStaveNote.getTicks();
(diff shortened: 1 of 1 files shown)


sschmidTU:
Thanks for the PR! Very good form with the test and description.
However, the visual regression tests on our test samples show some issues. In some cases, this patch makes the placement incorrect.
E.g. Mozart - An Chloe measure 5:
Before:
image

After:
image

also `test_fraction_infinite_loop_greatestCommonDenominator_measure3.musicxml`:

Before:
image

After:
image

(This rhythm is esoteric/rare, but at least the sample is short)

A guide to run visual regression tests is in visual_regression.sh (might not run on Windows, maybe with WSL2. runs on Linux or MacOS).

Maybe the rounding sometimes misfires? Have you tried always rounding down or up maybe?
Otherwise you'll have to look at these visual diffs and try to debug these samples.
Attaching the diffs for convenience:
[diff_tuplet_alignment.zip](https://github.com/user-attachments/files/24382193/diff_tuplet_alignment.zip)

rle-mino:
Hi 👋

First, thank you very much for the answer and the help you provided, especially for the visual testing part (I completely missed the `test:visual` command which could have helped me a lot!). Now, more details about the 2nd try.

---
## Why the first try failed:

1. Converting fractional ticks to integers via ticks.value() loses precision for triplets. For example, 4096/3 = 1365.33... becomes 1365, breaking the tick arithmetic.
2. Using Math.round(graphicalLength.RealValue * VF.RESOLUTION) converts the fraction to a floating-point number and back, introducing rounding errors.

## The (new) fix introduced in 70bfde3d

1. The fix uses VexFlow's Fraction class to preserve precision:
```typescript
if (sourceNote.NoteTuplet) {
    const graphicalLength: Fraction = voiceEntry.notes[0].graphicalNoteLength;
    const vfTicks: VF.Fraction = vexFlowVoiceEntry.vfStaveNote.getTicks();
    vfTicks.numerator = graphicalLength.Numerator * VF.RESOLUTION;
    vfTicks.denominator = graphicalLength.Denominator;
    vfTicks.simplify();
}
```

### Why it works:
1. No floating-point conversion: We multiply integers directly (Numerator * VF.RESOLUTION) and keep the denominator as-is, preserving exact fractional values.
2. VexFlow Fraction handles the math: By using vfTicks.simplify(), VexFlow reduces the fraction properly without precision loss.
3. Only affects tuplet notes: Non-tuplet notes keep VexFlow's default tick calculation, minimizing the impact on existing behavior.

### Still a visual regression...
...For an already invalid result!

#### The `test_note-ticks-0_issue1073.musicxml` case
Here is the develop version of `test_note-ticks-0_issue1073.musicxml` :

test_note-ticks-0_issue1073 musicxml_1_Blessed
> As we can see, at the end of the second measure, the left-hand notes are not correctly placed.


Here is the fix/vexflow-tuplet-tick-normalization version of `test_note-ticks-0_issue1073.musicxml`

test_note-ticks-0_issue1073 musicxml_1_Current
> Alignment fixed. Looking at the MusicXML file, this makes sense.


#### Cases that I broke with the first try (now fixed)
Mozart - An Chloe measure 5 FIXED ✅ 

Screenshot 2026-01-06 at 00 43 35

test_fraction_infinite_loop_greatestCommonDenominator_measure3 FIXED ✅ 

test_fraction_infinite_loop_greatestCommonDenominator_measure3 musicxml_1


Tell me if you need anything else here.

PR #1642, opened 2026-02-01
Fix grace note fingering positioning

isc:
Hi! This is my first contribution to OSMD. Thank you for making this library available — it's been very helpful for my piano training project.

---

## Summary

- Fix incorrect fingering positions on grace notes
- Grace notes now have collision offsets calculated per voice entry instead of across all notes in the staff entry

## Problem

Fingerings on grace notes were incorrectly shifted because the collision detection (`baseFingeringXOffset`) included all notes in the staff entry, without considering that grace notes are at different horizontal positions.

This caused issues like:
- Second grace note's fingering appearing above the first grace note
- Sequential grace note chords having fingerings shifted incorrectly

### Before (develop)
Screenshot 2026-02-01 at 14 01 22

### After (this PR)
Screenshot 2026-02-01 at 14 02 04

## Solution

Calculate fingering collision offsets separately for each grace note voice entry (chord), since they are rendered at different X positions. Non-grace notes continue to have offsets calculated together as before.

## Test plan

- [x] Added test MusicXML file `test_grace_note_fingerings_position.musicxml`
- [x] Added unit test verifying `baseFingeringXOffset === 0` for single grace notes
- [x] All 194 existing tests pass
- [x] Tested with sequential grace notes with fingerings
- [x] Tested with sequential grace note chords with fingerings

## Question for maintainers

I added a unit test that verifies `baseFingeringXOffset === 0` for single grace notes. However, this tests an implementation detail rather than the visual output.

Since the MusicXML test file is included in `test/data/`, it will be part of visual regression testing. **Would you prefer:**
1. **Keep the unit test** - runs automatically in CI, catches regressions quickly
2. **Remove the unit test** - rely only on visual regression testing with the MusicXML file

Happy to adjust based on your preference.

---
🤖 Generated with [Claude Code](https://claude.ai/code)

### DIFF
--- a/src/MusicalScore/Graphical/VexFlow/VexFlowMeasure.ts
+++ b/src/MusicalScore/Graphical/VexFlow/VexFlowMeasure.ts
@@ -1632,6 +1632,9 @@ export class VexFlowMeasure extends GraphicalMeasure {
             }
             let offsetX: number = this.rules.FingeringOffsetX;
             let modifierPosition: number; // VF.Stavemodifier.Position
+            // baseFingeringXOffset is now calculated per voice entry for grace notes
+            // (in VexFlowStaffEntry.setModifierXOffsets), so it correctly handles
+            // collision avoidance within each chord without cross-chord interference.
             switch (fingeringPosition) {
                 default:
                 case PlacementEnum.Left:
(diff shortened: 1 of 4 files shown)


sschmidTU (approved):
LGTM, merging.
Code looks good, visual regression tests pass.
Interesting that visual regression tests show no changes, which means this situation didn't happen in any of our samples. So, it's good that you added a test sample for it.
Thanks for the contribution!
Claude is surprisingly good with OSMD (a few months ago I had thought OSMD and Vexflow might be too idiosyncratic). Though in my experience you often have to guide him a little bit, and make corrections and give insights in multiple prompts.
Anyways, this PR is a straight improvement, no objections,
thanks for the contribution!

PR #1646, opened 2026-02-14
Fix octave shift dashed line too short on intermediate systems

isc:
## Summary

- Fix octave shift (8va/8vb) dashed continuation line stopping too short on intermediate systems when the shift spans 3+ systems
- The condition at line 1030 compared a system **ID** (`i`) against a **count** (`systemsInBetweenCount - 1`), which always evaluated `false` when the start system ID >= 1
- One-line fix: compare against `endStaffLine.ParentMusicSystem.Id - 1` instead

### Before 
Screenshot 2026-02-14 at 18 17 37

### After
Screenshot 2026-02-14 at 18 19 22

## Test plan

- [x] Added regression test in `VexFlowMeasure_Test.ts` that verifies `graphicalEndAtMeasureEnd` is set on intermediate systems
- [x] Added test MusicXML file with an 8va spanning 3 systems (with grace notes on the intermediate system)
- [x] All 195 tests pass (194 existing + 1 new)

### DIFF
--- a/src/MusicalScore/Graphical/VexFlow/VexFlowMusicSheetCalculator.ts
+++ b/src/MusicalScore/Graphical/VexFlow/VexFlowMusicSheetCalculator.ts
@@ -1027,7 +1027,7 @@ export class VexFlowMusicSheetCalculator extends MusicSheetCalculator {
             const nextOctaveShift: VexFlowOctaveShift = new VexFlowOctaveShift(octaveShift, nextShiftFirstMeasure.PositionAndShape);
             let nextShiftLastMeasure: GraphicalMeasure = this.findLastStafflineMeasure(nextShiftStaffline);
 
-            if (i < systemsInBetweenCount - 1) {
+            if (i < endStaffLine.ParentMusicSystem.Id - 1) {
               // if not - 1, the last octaveshift will always go to the end of the staffline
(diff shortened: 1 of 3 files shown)


sschmidTU:
Thanks for submitting the PR and the helpful sample.
I've "merged" this manually in a modified commit, as I've both added comments and slightly changed the test sample.
The sample's octave shift now ends before the end of the measure, which should test the in-between condition even more clearly (probably not necessary since you can tell from the existing or not existing "hook" down at the end of the shift, but maybe a little more clear in some cases).

By the way, the multi-line rendering for the test sample can be easily enforced with:
```js
osmd.EngravingRules.RenderXMeasuresPerLineAkaSystem = 1;
```
edit: added this to visual regression tests for this sample, see commit below

Sorry for not merging and crediting properly, I'll try to do that again in the future, but this was a single-line fix and I would have had to merge and then do the changes to both files. But I credited you in the commit message.

ps: I first thought we could solve this issue without exchanging the `systemsInBetweenCount` variable for the ` endStaffLine.ParentMusicSystem.Id` variable, but that doesn't work in a simple way, it looks like. I feel like the `systemsInBetweenCount` variable was meant to be used here, but the way the `for` loop was set up makes it not work.

PR #1662, opened 2026-04-07
feat: Migrate ESLint to v10 with flat config

isamu:
## Thank you

Thank you for this incredible project! OSMD has been an invaluable tool for the music technology community, and I truly appreciate all the work that has gone into it.

I noticed that ESLint 8 has reached end-of-life and the project is still using the legacy `.eslintrc.js` configuration format. I'd like to propose migrating to the latest ESLint with flat config support. If this aligns with the project's direction, please feel free to adopt it. If not, no worries at all — thank you for your consideration.

## Summary

- Migrate ESLint 8 → 10 with flat config (`eslint.config.mjs`)
- Delete legacy `.eslintrc.js`
- Update all ESLint-related packages to their latest versions
- Remove unused plugins

## Package Changes

| Package | Before | After |
|---------|--------|-------|
| `eslint` | ^8.57.0 | ^10.2.0 |
| `@eslint/js` | — | ^10.0.1 (new) |
| `typescript-eslint` | — | ^8.58.0 (new) |
| `@typescript-eslint/eslint-plugin` | ^8.0 | ^8.58.0 |
| `@typescript-eslint/parser` | ^8.0 | ^8.58.0 |
| `@stylistic/eslint-plugin-ts` | ^3.1.0 | removed (deprecated) |
| `@stylistic/eslint-plugin` | — | ^5.10.0 (replacement) |
| `eslint-plugin-jsdoc` | ^46.4.5 | ^62.9.0 |
| `eslint-plugin-no-null` | ^1.0.2 | removed (unused, rule was off) |
| `eslint-plugin-import` | ^2.27.5 | removed (no rules configured) |
| `eslint-plugin-node` | ^11.1.0 | removed (no rules configured) |

## Configuration Changes

- `.eslintrc.js` → `eslint.config.mjs` using `defineConfig()` from `eslint/config`
- `parserOptions` → `languageOptions.parserOptions` with `projectService: true`
- `ignorePatterns` → top-level `ignores` config object
- `overrides` → separate config objects with `files` patterns
- Lint script simplified: removed `-c .eslintrc.js --ext .ts` flags (unnecessary with flat config)

## Source Changes

- Removed two obsolete `eslint-disable no-null/no-null` comments from `MusicSheetReader.ts` (plugin was removed)
- Fixed a trailing whitespace in `MusicSheetReader.ts`

## Compatibility

[...]

PR #1663, opened 2026-04-07
lint: Resolve no-undef with proper globals configuration

isamu:
## Summary

Follow-up to #1662 — configures proper global variable definitions so that `no-undef` works correctly across the codebase.

## Changes

- Add `globals` package for standard global variable definitions
- Add `browser` + `node` globals for TypeScript source files
- Add `mocha` + `chai` globals for test TypeScript files
- Add `browser` + `node` + `mocha` globals for test JS/MJS utility files


sschmidTU:
Apparently the `globals` package is no longer necessary in eslint 9 (only in eslint 8), can you remove it from package.json?
I removed it and eslint works fine.

> This package is used by ESLint 8 and earlier. For ESLint 9 and later, you should depend on this package directly in [your ESLint config](https://eslint.org/docs/latest/use/configure/language-options#predefined-global-variables).

https://www.npmjs.com/package/globals
https://eslint.org/docs/latest/use/configure/language-options#predefined-global-variables


Also, please merge the develop branch, as checking out this pr without merging develop leads to lint errors.

isamu:
Thanks for the review! I looked into this a bit more — it turns out `globals` is actually still required for ESLint 9+. The wording on the npm page is a little confusing, but it means "install it directly as a dep" (in ESLint 8 it was bundled via `env`, in 9+ you need to install/import it yourself). The official ESLint 9 docs recommend the same approach:

https://eslint.org/docs/latest/use/configure/language-options#predefined-global-variables

On my side, removing `globals` from `package.json` causes `eslint.config.mjs` to fail loading with `ERR_MODULE_NOT_FOUND`. Could it be a leftover `node_modules` state on your end? A fresh `rm -rf node_modules && npm install` might reproduce it.

sschmidTU:
Strange, eslint works for me even after `rm -rf node_modules && npm install` when removing globals from package.json.
Can you run `npx eslint --version` (or `npm ls eslint`) and verify you have eslint v9.39.4 as well?
looks like it's installed as a sub module of eslint 9.39.4:
```sh
npm ls globals
opensheetmusicdisplay@1.9.7
└─┬ eslint@9.39.4
  └─┬ @eslint/eslintrc@3.3.5
    └── globals@14.0.0
```

Maybe you're using yarn again instead of npm? Again, this repository uses npm as the standard, never yarn.

also maybe relevant what your node and npm versions are.
I'm using npm 10.8.2 on Linux and 11.9.0 on Windows.

isamu:
You were right, sorry for the confusion! I tried again with a clean `rm -rf node_modules && npm install` and it works fine — `globals` gets hoisted to the top level as a transitive dep via `@eslint/eslintrc`, so the import resolves without needing it in `package.json`.

I'll push a commit removing `globals` from `package.json`. Thanks for your patience!

sschmidTU:
no problem, thanks for the adjustments!
LGTM, merging.

PR #1666, opened 2026-04-07
lint: Add block scoping to case clauses with lexical declarations

isamu:
## Summary

- Wrap case clause bodies in `{}` blocks where `const`/`let` declarations are used
- Remove `no-case-declarations: "off"` override from ESLint config

## Files changed

- `src/MusicalScore/Graphical/GraphicalUnknownExpression.ts` (2 cases)
- `src/MusicalScore/Graphical/MusicSystemBuilder.ts` (1 case)
- `src/MusicalScore/Graphical/VexFlow/VexFlowMeasure.ts` (1 case)
- `src/MusicalScore/ScoreIO/MusicSymbolModules/RepetitionCalculator.ts` (1 case)

### DIFF
--- a/eslint.config.mjs
+++ b/eslint.config.mjs
@@ -139,7 +139,6 @@ export default defineConfig(
             "no-prototype-builtins": "off",
             "no-redeclare": "error",
             "for-direction": "off",
-            "no-case-declarations": "off",
             "no-constant-binary-expression": "off",
             "no-constant-condition": "off",
             "no-irregular-whitespace": "error",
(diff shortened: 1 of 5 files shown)


sschmidTU:
This seems very verbose and unnecessary. I don't like this rule unless there's a technical reason for it. It just adds unnecessary (?) lines with brackets on them, adding indentation layers.

isamu:
Fair point — the added brackets do make the code more verbose without a strong technical benefit here. Closing this PR.

PR #1671, opened 2026-04-16
Visual polish

codesenyou:
This PR polishes the music sheet rendering with:
1. Better tie pairing detection
2. Thicker and more visible ties
3. Beam/stem collision dection that helps avoiding overlaps
4. Better space between bass and treble when cross staves notes are present


codesenyou:
Sorry, wrong repo I wanted to PR my fork

PR #1672, opened 2026-05-05
Fix lyrics overlap for pickup measure / last note in measure multi-syllable mid-word continuation

### LINKED ISSUE #1659: Lyrics overlapping
Hello, 
I encountered a situation where the lyrics are overlapping. If the lyrics at the end of a measurement are long enough they can easily overlap with the next one, in default settings. It seems a bit rare but worth mentioning. Could greatly vary between languages.

Maybe there is some settings, I could use to fix it, but I could not fin any...

What I found is that if I set
```js
osmd.EngravingRules.LyricsAlignmentStandard = TextAlignmentEnum.CenterBottom;
```
I can get a little more room, but not a solution. Again, not a big problem but something one can encounter.

Image

Image

sschmidTU:
Fixes #1659

before:
image

after:
image

### DIFF
--- a/src/MusicalScore/Graphical/EngravingRules.ts
+++ b/src/MusicalScore/Graphical/EngravingRules.ts
@@ -293,7 +293,7 @@ export class EngravingRules {
      *  has a syllable + dash glyph right at its start — the usual measure-end buffer doesn't apply.
      *  Set to 0 for maximum cross-measure clearance (no reduction), or up to
      *  LyricsXPaddingReductionForLastNoteInMeasure for the same reduction as a normal last note.
-     *  Default 0.6 (half of the regular 1.2). */
+     *  Default 1.0 (regular note 1.2). */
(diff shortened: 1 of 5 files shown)


sschmidTU:
LGTM, merging.

Visual Changes to other samples are minimal now, many samples are actually slightly more compact now, maybe because of the included overlap calculation fix.

visual diffs from visual regression tests:
[diff_lyrics_overlap_pickup_fix4_dashEndPaddingReduced_1.0.zip](https://github.com/user-attachments/files/27404194/diff_lyrics_overlap_pickup_fix4_dashEndPaddingReduced_1.0.zip)

I think these are all improvements or irrelevant changes, except maybe for Schubert an die Musik measure 9 being a bit too wide, which looked better beforehand, but I think that's just a random side effect of other measures saving space, so this measure gets a bit more incidentally.

PR #1675, opened 2026-05-27
fix: ensure currentVoiceEntries is safely cloned in MusicPartManagerIterator

LukasLokajicek:
MusicPartManagerIterator.clone() assigns currentVoiceEntries by reference rather than by value. Any code that
advances the returned clone (e.g. via moveToNextVisibleVoiceEntry or moveToNext) calls moveToNext() on the clone,
which zeroes the shared array in-place with this.currentVoiceEntries.length = 0. This silently destroys the
original iterator's voice entry list, so subsequent calls to CurrentVisibleVoiceEntries(), GNotesUnderCursor(), or
NotesUnderCursor() on the original cursor return empty arrays.

### DIFF
--- a/src/MusicalScore/MusicParts/MusicPartManagerIterator.ts
+++ b/src/MusicalScore/MusicParts/MusicPartManagerIterator.ts
@@ -14,10 +14,10 @@ import {InstantaneousDynamicExpression} from "../VoiceData/Expressions/Instantan
 import {MultiTempoExpression} from "../VoiceData/Expressions/MultiTempoExpression";
 import {AbstractExpression} from "../VoiceData/Expressions/AbstractExpression";
 import log from "loglevel";
-import { MusicSheet } from "../MusicSheet";
-import { NoteHeadShape } from "../VoiceData/Notehead";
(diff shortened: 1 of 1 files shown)


sschmidTU:
Thanks again!

Would be nice to add a test for this based on the reproduction:
> ## Reproduction
> const cursor = osmd.cursor;
> cursor.reset();
> cursor.next(); // advance to first note
> 
> // At this point GNotesUnderCursor() correctly returns notes
> console.log(cursor.GNotesUnderCursor().length); // e.g. 2
> 
> // Clone and advance — this is the common pattern for "peek at next position"
> const clone = cursor.iterator.clone();
> clone.moveToNextVisibleVoiceEntry(false);
> 
> // Now the original is corrupted
> console.log(cursor.GNotesUnderCursor().length); // 0  ← BUG
> The pattern appears in typical playback/training apps that call `getCursorPosition()` (which peeks ahead via a clone) before reading notes via `GNotesUnderCursor()`.

to make sure this doesn't happen again in future.
You can either add it to the PR or I can add it after merging.

sschmidTU:
LGTM, merging. No visual regressions.
I'll add a test for this after merging the PR.

PR #1676, opened 2026-05-28
fix(Tuplet): Align cross-staff tuplet notes with simultaneous notes in other voices

isc:
## Problem

When a tuplet voice is split across staves — e.g. a left-hand triplet whose top note is notated on the upper staff — the notes of the tuplet that fall on a given staff are not horizontally aligned with simultaneous notes in other voices on that staff.

A concrete example is the left hand of Schumann's *Kinderszenen No. 1*: on every second beat the first note of the triplet should sit exactly above the bass note it is played with, but it is shifted noticeably to the right.

## Root cause

When a voice is missing a member on a staff (because that member is notated on the other staff), `VexFlowMeasure` fills the gap with ghost notes via `VexFlowConverter.GhostNotes()`, which decomposed the gap into standard note-value rests with `durations(frac, false)`.

A tuplet member such as a triplet eighth lasts `1/12` of a whole note, which is **not a dyadic fraction** and cannot be represented exactly by standard rests. The greedy decomposition overshoots:

```
1/12  ->  1/16 + 1/64 + 1/128  =  11/128   (1024 + 256 + 128 = 1408 ticks)
```

instead of the correct `RESOLUTION / 12 ≈ 1365.33` ticks. The voice's accumulated ticks then drift (~42.67 ticks per gap), so every following note is placed in a different `TickContext` than the simultaneous notes in other voices and is no longer vertically aligned.

This is the cross-staff / ghost-note counterpart of the StaveNote tuplet tick fix in #1631.

## Fix

`GhostNotes()` now always emits a **single** ghost note and sets its ticks to exactly match the gap (`frac * VF.RESOLUTION`, kept as an exact rational via `VF.Fraction`), mirroring the StaveNote approach. Since ghost notes are invisible, the standard-rest decomposition was never needed; a single exact-tick ghost is simpler, behaves identically for dyadic gaps, and (per review) slightly improves spacing in some samples with no visual regressions.

## Tests

- New sample `test/data/test_tuplet_crossstaff_alignment.musicxml`: a two-staff part with a cross-staff triplet (3rd note on the upper staff) over plain eighth notes on the lower staff.
- Two unit tests in `VexFlowMeasure_TupletAlignment_Test.ts`:
  - the cross-staff ghost filler has the exact triplet-eighth length (no overshoot) and the voice's ticks sum to the measure duration;
  - the beat-2 tuplet note and the simultaneous eighth note share the same cumulative tick position.

Full suite passes locally (197 tests).

### DIFF
--- a/src/MusicalScore/Graphical/VexFlow/VexFlowConverter.ts
+++ b/src/MusicalScore/Graphical/VexFlow/VexFlowConverter.ts
@@ -229,6 +229,29 @@ export class VexFlowConverter {
     }
 
     public static GhostNotes(frac: Fraction): VF.GhostNote[] {
+        // When a gap's duration is not a dyadic fraction (denominator not a power of two) -
+        // e.g. a tuplet member like a triplet eighth (1/12) - it cannot be represented exactly
+        // by standard note-value rests. durations() then decomposes it greedily and overshoots
+        // (1/12 -> 1/16 + 1/64 + 1/128), so the voice's accumulated ticks drift and every
+        // following note is pushed out of vertical alignment with simultaneous notes in other
+        // voices. This typically happens with cross-staff tuplets, where the off-staff members
+        // are filled with ghost notes on this staff. Emit a single ghost note whose ticks match
+        // the gap exactly, mirroring the StaveNote tuplet tick fix in graphicalMeasureCreatedCalculations().
+        const denominator: number = frac.Denominator;
+        // dyadic = denominator is a power of two (1, 2, 4, 8, ...), i.e. log2 is an integer.
+        const isDyadic: boolean = denominator > 0 && Number.isInteger(Math.log2(denominator));
+        if (!isDyadic) {
+            const ghostNote: VF.GhostNote = new VF.GhostNote({
+                duration: VexFlowConverter.durations(frac, true)[0],
+            });
+            // ticks = frac * VF.RESOLUTION, kept as an exact rational via VF.Fraction
+            // (e.g. 1/12 -> 16384/12 -> 4096/3) so the voice's resolution multiplier stays correct.
+            const vfTicks: VF.Fraction = ghostNote.getTicks();
+            vfTicks.numerator = frac.Numerator * VF.RESOLUTION;
+            vfTicks.denominator = frac.Denominator;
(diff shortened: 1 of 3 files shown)


sschmidTU:
Looks good to me.
I'm fine with always emitting a single ghost note, especially as it seems to improve/compact spacing in some samples.
Especially `test_tuplet_multivoice_alignment.musicxml`:
Before:
image

After:
image

It looks like it was using more spacing than necessary between notes 3 and 4 before.
(it's easier to see by switching between the 2 images on top of each other).

Here are the visual regressions for always single ghost note:
[diff_pr1676_always_nonDyadic_path.zip](https://github.com/user-attachments/files/28379950/diff_pr1676_always_nonDyadic_path.zip)

(there are no visual regressions if we keep the dyadic path, except for the new sample)

@isc I used this quick edit for testing, that should be functionally equivalent to what you want, right?
(of course, we'd just remove the `if` instead of using `if(true || !isDyadic)`)
```ts
        const isDyadic: boolean = denominator > 0 && Number.isInteger(Math.log2(denominator));
        if (true || !isDyadic) {
            // (rest of the code is unchanged)
            const ghostNote: VF.GhostNote = new VF.GhostNote({
                duration: VexFlowConverter.durations(frac, true)[0],
            });
            // ticks = frac * VF.RESOLUTION, kept as an exact rational via VF.Fraction
            // (e.g. 1/12 -> 16384/12 -> 4096/3) so the voice's resolution multiplier stays correct.
            const vfTicks: VF.Fraction = ghostNote.getTicks();
            vfTicks.numerator = frac.Numerator * VF.RESOLUTION;
            vfTicks.denominator = frac.Denominator;
            vfTicks.simplify();
            return [ghostNote];
        }
```

isc:
Yes, that's functionally equivalent to what I want — thanks for digging into the visual regressions! I've simplified it accordingly: removed the dyadic/non-dyadic branch entirely, so `GhostNotes()` now always emits a single exact-tick ghost note. The standard-rest decomposition wasn't needed for invisible ghosts anyway, and as you found it's a bit more compact. Tests still pass (197).

sschmidTU:
Great, thanks!
LGTM, merging.

PR #1677, opened 2026-05-29
fix(Notes): Don't stagger a unison between two voices in a three-voice stave

isc:
## Problem

In a three-voice stave, when two of the voices play the **same pitch** at the same time (a unison), the noteheads are pushed apart horizontally instead of sharing one position. Example: a left-hand passage where an inner (triplet) voice and a tied bass voice both land on the same G — they render staggered, as if they were different notes.

The two-voice path already handles this correctly (it does **not** stagger a same-shape unison). The three-voice path did not.

## Cause

`ModifierContext.formatNotes` (VexFlowPatch `stavenote.js`) shifts the middle voice right whenever it "intersects" the upper or lower voice:

```js
// If middle voice intersects upper or lower voice
if ((!noteU.isrest && !noteM.isrest && noteU.minLine <= noteM.maxLine + 0.5) ||
    (!noteM.isrest && !noteL.isrest && noteM.minLine <= noteL.maxLine)) {
  xShift = voiceXShift + 3;      // shift middle note right
  noteM.note.setXShift(xShift);
}
```

A unison makes the middle voice share a staff line with the upper (or lower) voice, so this generic collision rule fires and staggers it — even though identical pitches should overlap.

## Fix

Only shift the middle note for a **genuine** collision. If the intersection is a *mergeable unison*, leave the noteheads overlapping. "Mergeable" mirrors the existing two-voice stagger conditions: same staff line, single noteheads, matching notehead shape (not one half/whole and the other not), and matching dot count. Genuine collisions (e.g. a second) are still staggered.

## Tests

- New sample `test/data/test_three_voice_unison_alignment.musicxml`: three voices where beat 1 is a unison (voices 1 & 2 on G4, voice 3 an octave below) and beat 2 is a second-interval collision (A4 vs G4).
- New `VexFlowMeasure_ThreeVoiceUnison_Test.ts`:
  - the beat-1 unison is not staggered (all notes `x_shift === 0`);
  - the beat-2 second is still staggered (`x_shift > 0`).

Full suite passes locally (197 tests). The change is scoped to exact, mergeable unisons in a three-voice ModifierContext, so non-unison collisions are byte-for-byte unchanged; visual regression may show intended improvements on samples that contain three-voice unisons.

### DIFF
--- a/src/MusicalScore/Graphical/VexFlow/VexFlowConverter.ts
+++ b/src/MusicalScore/Graphical/VexFlow/VexFlowConverter.ts
@@ -229,15 +229,22 @@ export class VexFlowConverter {
     }
 
     public static GhostNotes(frac: Fraction): VF.GhostNote[] {
-        const ghostNotes: VF.GhostNote[] = [];
-        const durations: string[] = VexFlowConverter.durations(frac, false);
-        for (const duration of durations) {
-            ghostNotes.push(new VF.GhostNote({
-                duration: duration,
-                //dots: dots
-            }));
-        }
-        return ghostNotes;
+        // Fill the gap with a single ghost note whose ticks exactly match it. A gap can be a
+        // tuplet member (e.g. a triplet eighth = 1/12), which is not a dyadic fraction and so
+        // can't be represented exactly by standard note-value rests: decomposing it overshoots
+        // (1/12 -> 1/16 + 1/64 + 1/128) and drifts the voice's accumulated ticks, breaking
+        // vertical alignment with simultaneous notes in other voices (e.g. cross-staff tuplets).
+        // Setting the ticks directly as an exact rational keeps the voice's resolution multiplier correct.
+        const ghostNote: VF.GhostNote = new VF.GhostNote({
+            duration: VexFlowConverter.durations(frac, true)[0],
+        });
+        // ticks = frac * VF.RESOLUTION, kept as an exact rational via VF.Fraction
+        // (e.g. 1/12 -> 16384/12 -> 4096/3) so the voice's resolution multiplier stays correct.
+        const vfTicks: VF.Fraction = ghostNote.getTicks();
+        vfTicks.numerator = frac.Numerator * VF.RESOLUTION;
+        vfTicks.denominator = frac.Denominator;
+        vfTicks.simplify();
+        return [ghostNote];
     }
 
     /**
--- a/src/VexFlowPatch/src/stavenote.js
+++ b/src/VexFlowPatch/src/stavenote.js
@@ -48,13 +48,15 @@ function centerRest(rest, noteU, noteL) {
   rest.minLine -= delta;
 }
 
-// Whether two single-notehead notes from different voices form a unison that can
-// share the same horizontal position (overlapping noteheads) instead of being
-// staggered. Mirrors the stagger conditions used in the two-voice path: a
-// notehead cannot cleanly overlap if the shapes differ (one half/whole and the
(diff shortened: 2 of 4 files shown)


isc:
One follow-up thought on factoring, happy to do it in this PR or skip it — your call.

The decision "can these two noteheads share a column?" now lives in two places with slightly different logic:
- the **two-voice** path computes `staggerConditions` inline (`halfNoteCount === 1 || wholeNoteCount === 1 || dots differ`, plus `StaggerSameWholeNotes`);
- the **three-voice** path uses the new `unisonMergeable` helper.

That divergence is exactly what caused this bug (the two-voice path already de-staggered unisons; the three-voice path didn't). I could extract a single shared predicate, e.g.:

```js
// noteheads can share a column iff: same line, matching shape (not one half/whole
// and the other not), matching dots, and (optionally) not two identical whole notes
function mergeableUnison(a, b, staggerSameWholeNotes) { ... }
```

and have both paths call it — the two-voice condition becomes `lineDiff === 0 && !mergeableUnison(noteU, noteL, stagger_same_whole_notes)`, which is functionally identical to today's `staggerConditions` (so no two-voice behavior change), while the three-voice path keeps its extra single-notehead guard. It's a small, behavior-preserving consolidation that keeps the two paths from drifting apart again.

I kept this PR minimal for now. Want me to add that refactor here (I can run the visual regression to confirm no changes), or leave it as a separate follow-up?

sschmidTU:
Yes, let's do the factoring into the mergeableUnison function, that's cleaner (avoids code duplication) and could prevent future bugs as well. Thanks!

sschmidTU:
LGTM.
Visual regressions are clean except for the fix in the new sample and one nice fix in Cornelius Christbaum measure 25 and 33.
[diff_pr1677_unison_stagger2_develop_merged.zip](https://github.com/user-attachments/files/28382781/diff_pr1677_unison_stagger2_develop_merged.zip)

This is ready from your side, right? Or are you still working on changes?
I would merge otherwise.
You could merge develop so the Files Changed on Github is cleaner, but it's not necessary.

sschmidTU:
Oh and you could make the unisonMergeable function comment a jsdoc (/** instead of //).

isc:
Done — factored the decision into a shared `mergeableUnison(a, b, staggerSameWholeNotes)` used by both paths (latest commit). The two-voice path's condition is now `lineDiff === 0 && !mergeableUnison(noteU, noteL, stagger_same_whole_notes)`, logically identical to the old inline `staggerConditions`; the three-voice path keeps its extra single-notehead guard.

I ran a visual-regression self-diff to confirm no rendering changes: built the bundle before and after the refactor, regenerated the `allSmall` set with `--osmdtesting`, and compared every image with `compare -metric AE`. **270 samples, 0 pixels different** — so the consolidation is behavior-preserving, and the only behavioral change in this PR remains the original three-voice unison fix. (197 unit tests still pass.)

isc:
Done — converted the `mergeableUnison` comment to a jsdoc block. Thanks for running the visual regression, and nice catch that it also fixes the Cornelius *Christbaum* unisons (m25/33) as a bonus. This is ready from my side — good to merge. (I left the branch off the pre-#1676 develop; the Files Changed is just the three files, so I skipped merging develop, but happy to if you'd prefer.)

isc:
Merged `develop` in as you suggested — the Files Changed is now just the three-voice unison fix (the tuplet PR #1676 is in the base, so it no longer shows up). Clean merge, 200 tests pass. Ready to merge.

sschmidTU:
Great, thanks again! Merging.

PR #1678, opened 2026-05-30
fix(Notes): merge unison noteheads that differ only in dots

isc:
## Summary

`mergeableUnison()` lets a same-line unison across two voices share one horizontal position (overlapping noteheads). It made one exception: when the two notes differ in **dot count**, it kept them staggered side by side — even though their noteheads have the same shape, and engraved editions (and MuseScore) overlap them, sharing a single augmentation dot.

A typical case is a dotted eighth in unison with an eighth-triplet note — e.g. the left hand of Schumann *Kinderszenen* No. 1, m10, where the dotted-eighth voice and the triplet voice start on the same pitch. OSMD drew two separate noteheads (with two staggered dots) instead of one.

So this **narrows the dot-count guard** in `mergeableUnison()` rather than removing it outright: a dot-differing unison now overlaps, **except** when *both* voices are dotted with *different* dot counts (e.g. a single- vs triple-dotted note). There the two sets of augmentation dots would collide at the shared column, and engraving has no settled rule for that contrived case (it doesn't occur in real music), so those stay staggered. The notehead-shape guard (half/whole vs not) and `StaggerSameWholeNotes` are unchanged.

```js
// merge dot-differing unisons, but not when both voices are dotted with
// different counts (their augmentation dots would collide at the shared column)
if (a.note.dots > 0 && b.note.dots > 0 && a.note.dots !== b.note.dots) return false;
```

## Why this is a fix, not a style option

This differs from `StaggerSameWholeNotes`. Two identical **whole notes** are stemless, so merging them would hide that there are two voices — a genuine editorial choice, hence a toggle. A dotted-vs-plain unison is **stemmed**: the shared notehead keeps both stems (and the dotted voice's dot), so the two voices stay legible. Staggering them is simply incorrect engraving. *(Earlier commits in this PR added an `EngravingRules.StaggerDottedUnisons` option; per the reasoning above I removed it in favor of always merging dotted-vs-plain unisons — fine to squash.)*

## Tests

- New two-voice sample + unit test asserting the noteheads overlap (no stagger x-shift) for a dotted-vs-plain unison.
- **Visual-regression self-diff** (built develop baseline and this branch, regenerated the `allSmall` set with `--osmdtesting`, compared every image with `compare -metric AE`): **3 / 272 samples differ**, all dotted-vs-plain unison merges, all localized to the affected noteheads (no layout reflow):
  - `test_dotted_unison_alignment` (new sample)
  - `Cornelius_Christbaum` (the unison you noted in #1677, m21/25/33)
  - `Dichterliebe01` (several dotted unisons in the piano part)

  `OSMD_Function_Test_Voice_Alignment` no longer differs from develop: its only dot-differing unison (m2) is a single- vs triple-dotted note — the both-dotted case we intentionally keep staggered, so it matches develop.

### DIFF
--- a/src/MusicalScore/Graphical/EngravingRules.ts
+++ b/src/MusicalScore/Graphical/EngravingRules.ts
@@ -412,6 +412,10 @@ export class EngravingRules {
     public SoftmaxFactorVexFlow: number;
     /** Stagger (x-shift) whole notes that are the same note, but in different voices (show 2 instead of 1). */
     public StaggerSameWholeNotes: boolean;
+    /** Stagger (x-shift) a unison in different voices when the two notes differ only in their dot count
+     * (e.g. a dotted eighth in unison with an eighth triplet note). When false, the noteheads overlap and
+     * share a single augmentation dot, matching MuseScore and most engraved editions. */
+    public StaggerDottedUnisons: boolean;
     public MaxInstructionsConstValue: number;
     public NoteDistances: number[] = [1.0, 1.0, 1.3, 1.6, 2.0, 2.5, 3.0, 4.0];
     public NoteDistancesScalingFactors: number[] = [1.0, 2.0, 4.0, 8.0, 16.0, 32.0, 64.0, 128.0];
@@ -907,6 +911,7 @@ export class EngravingRules {
         // if too high, score gets too big, especially half notes. with half note quarter quarter, the quarters get squeezed.
         // if too low, smaller notes aren't positioned correctly.
         this.StaggerSameWholeNotes = true;
+        this.StaggerDottedUnisons = true;
 
         // Render options (whether to render specific or invisible elements)
         this.AlignRests = AlignRestOption.Never; // 0 = false, 1 = true, 2 = auto
--- a/src/MusicalScore/Graphical/EngravingRules.ts
+++ b/src/MusicalScore/Graphical/EngravingRules.ts
@@ -412,10 +412,6 @@ export class EngravingRules {
     public SoftmaxFactorVexFlow: number;
     /** Stagger (x-shift) whole notes that are the same note, but in different voices (show 2 instead of 1). */
     public StaggerSameWholeNotes: boolean;
-    /** Stagger (x-shift) a unison in different voices when the two notes differ only in their dot count
-     * (e.g. a dotted eighth in unison with an eighth triplet note). When false, the noteheads overlap and
-     * share a single augmentation dot, matching MuseScore and most engraved editions. */
-    public StaggerDottedUnisons: boolean;
     public MaxInstructionsConstValue: number;
     public NoteDistances: number[] = [1.0, 1.0, 1.3, 1.6, 2.0, 2.5, 3.0, 4.0];
     public NoteDistancesScalingFactors: number[] = [1.0, 2.0, 4.0, 8.0, 16.0, 32.0, 64.0, 128.0];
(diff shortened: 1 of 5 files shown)


sschmidTU:
Thanks! The change is definitely an improvement, though the new test sample (`test_dotted_unison_alignment`) doesn't make sense to me musically.
There is no visible difference between voice 1 and 2.
It would be better if the voice with the dotted quarter had an 8th note following it (ideally a pitched note and not a rest note, which would be musically almost the same as the quarter note in the other voice), and the quarter note had another quarter note following it, which makes the stem direction / voicing visible, revealing the difference in note values between the voices.

Current state:
After:
image

Before:
image

sschmidTU:
Corrected my previous comment: dotted quarter, not dotted eighth.
Also suggested adding a non-rest note quarter note to the other voice.

isc:
Good point — thanks. I've reworked `test_dotted_unison_alignment.musicxml` so the note values are actually visible instead of being hidden behind rests:

- **Voice 1**: dotted quarter G4 → eighth **A4** (a pitched note, not a rest)
- **Voice 2**: quarter G4 → quarter **D4**

So beat 1 is the dotted-vs-plain unison being tested (one shared notehead, two stems, a single dot), and the continuation makes the differing note values and the two-voice stem directions legible — exactly as you suggested.

The unit test is unchanged (it asserts on beat 1, which is unchanged) and still passes. Pushed in 8b35f92.

sschmidTU:
Thanks, looks great now! Merging.

After:
image

Before:
image

sschmidTU:
This has some positive visual regressions / fixes by the way:
measure 21 in Cornelius Christbaum
After:
image

Before:
image

Dichterliebe measure 13-14
After:
image

Before:
image

sschmidTU:
There is technically a deterioration/regression in OSMD_Function_Test_Voice_Alignment, but that's a very contrived/technical sample. I'm fine with accepting this as a compromise in favor of the fixes in the other samples.
(the dots overlap)
After:
image

Before:
image
(see measure 2)
This is not an issue in any realistic/classical music sample in our test samples.

So maybe @isc you want to have a quick attempt at a fix, but again, that would not be necessary, I can also just merge right away.

isc:
Thanks for catching the `OSMD_Function_Test_Voice_Alignment` m2 regression — I took a shot at it rather than leaving it as a compromise, and I think it cleans up nicely.

**The fix (2e45763):** the dot collision only happens when *both* voices are dotted with *different* dot counts (m2 is a single-dotted eighth in unison with a triple-dotted quarter — 1 vs 3 dots), so their augmentation dots pile up at the shared column. I checked the references and engraving has no settled rule for that case (it doesn't occur in real music — Gould doesn't cover it), so I now keep *those* staggered:

```js
if (a.note.dots > 0 && b.note.dots > 0 && a.note.dots !== b.note.dots) return false;
```

This is strictly narrower than the original dot guard I removed: when one side has **0 dots** — the common dotted-vs-plain or dotted-vs-tuplet unison — there's no second dot to collide, so it still merges. So all the real-music fixes are untouched:

- I scanned `Cornelius_Christbaum` and `Dichterliebe01`: **every** dot-differing unison in them is dotted-vs-plain (e.g. Cornelius m21 is 0 vs 1 dot). None is both-dotted-with-different-counts.
- Self-diff (this branch with vs without the guard, same render) differs in **exactly one spot**: m2 of the voice-alignment sample. m2 reverts to the staggered layout (= develop), the dots no longer overlap, and nothing else in any sample changes.
- Unit tests still pass (the new sample is dotted-vs-plain, so it still overlaps as asserted).

So the contrived regression is gone and the Cornelius/Dichterliebe/Schumann merges are preserved. Ready whenever you are.

sschmidTU:
Great, the regression in `OSMD_Function_Test_Voice_Alignment.musicxml` is fixed, thanks!
Merging.

Issue #1587, opened 2025-02-23
Fingering order not correct for multiple voices


fablau:
Hello Simon and team.

Here is a new issue I'd like to tackle regarding fingerings.

It looks like that if the XML file has a chord composed of different voices and each voice has a different fingering, they are not displayed in the correct order.

Look at the example below for piano solo:

![Image](https://github.com/user-attachments/assets/13e44027-b5b3-4c47-ad25-9144ad5ddec0)

The fingerings on the first, second, and third beat are assigned to two different voices, and their vertical order should be preserved.

Here is how OSMD display them:

![Image](https://github.com/user-attachments/assets/5ea2cc34-2c8e-4c5a-a4a9-2e68f9e0ef6e)

Their order is inverted. Any way to fix this issue?

I am attaching the MusicXML file for this example in the .zip file below.

Thanks!

[FingeringTest.zip](https://github.com/user-attachments/files/18927480/FingeringTest.zip)

sschmidTU:
Fixed. Thanks for the report!

<img width="268" alt="Image" src="https://github.com/user-attachments/assets/b3aed1d1-cf6a-4cf8-aa51-189eaf506bcc" />

This was a rare situation where the visual note order is not the same as the fingering order in our processing.

fablau:
Awesome! Thanks!

fablau:
Looking forward to having these fixings applied to the Extended version as well!

sschmidTU:
Yup, will be applied to extended soon! Still busy fixing things ;)

fablau:
Great thanks!

sschmidTU:
@fablau merged to osmd-extended!

fablau:
Awesome! Thank you so much!

fablau:
Simon, actually, there is still the same problem in the bottom staff of the piano in this example:

![Image](https://github.com/user-attachments/assets/48b7302e-f16c-49bd-a1f8-23d5b7f7a864)

It should be like this:

![Image](https://github.com/user-attachments/assets/e71b93ce-d843-4a27-ac5e-a21248d5cdfb)

I am attaching the MusicXML file of this example. Thanks!

[BottomStaffExample.musicxml.zip](https://github.com/user-attachments/files/19526605/BottomStaffExample.musicxml.zip)

fablau:
Wonderful, thank you, Simon!

PR #1685, opened 2026-06-11
fix(Fingering): stack Above/Below fingerings in the pitch order of their notes

isc:
## Summary

`calculateFingerings()` collects the Above/Below fingerings of a staff entry voice by voice and stacks them in that order. When a beat combines notes from several voices — e.g. Beethoven *Pathétique* 2nd mvt m24, where voice 1 plays a two-note chord (fingerings 3 and 5) over the beat's lowest note held by voice 2 (fingering 2) — the second voice's fingering always lands at the outer end of the stack, regardless of its note's position in the chord: the stack reads 3/5/2 bottom-to-top instead of 2/3/5.

This generalizes the #1587 fix: the `topNote` reverse heuristic added there handles the case where each voice contributes **one** fingering (voice 1's note on top), but not the case where one of the voices is a **chord** — `fingerings[0]` is then the chord's bottom note, not the top note, so the heuristic doesn't trigger.

## Solution

When every collected fingering belongs to a distinct pitched note, sort the fingerings by their note's pitch so the stack mirrors the chord: the lowest note's fingering closest to the staff for Above placement, reversed for Below — same convention the existing heuristics produce for the single-note-per-voice case, so the #1587 samples render pixel-identical (verified below). The previous `reverse()` heuristics are kept as a fallback for bulk fingerings (several per note, Sibelius-style, where sorting by the shared note would be a no-op anyway) and unpitched notes.

## Screenshots

New test sample (a chord in voice 1 over the beat's lowest note in voice 2, mirrored on the bass staff — modeled on Pathétique 2nd mvt m24). Expected: treble reads 5/3/1 top-to-bottom (lowest note's fingering closest to the staff), bass reads 1/3/5 (highest note's fingering closest to the staff).

Before:
before

After:
after

## Tests

- New sample `test_fingering_two_voices_pitch_order.musicxml` + unit test asserting the exact top-to-bottom label order of all four stacks (both staves, i.e. both placements). All 214 tests pass.
- **Visual-regression self-diff** (built develop baseline and this branch, regenerated the `allSmall` set with `--osmdtesting`, compared every image with `compare -metric AE`): **3 / 282 samples differ**, and nothing else moves (no layout reflow):
  - `test_fingering_two_voices_pitch_order` (the new sample, above)
  - `test_fingering_reversed_rare_1587`, `..._onefingering` and `..._bottom_staff_multiple_fingerings_per_note` are all **pixel-identical** to develop ✅
  - `OSMD_Function_Test_Tabulature_hayden_study_1` — positive change: the guitar chords spread across voices now read in pitch order. E.g. the B2(1)+D#4(3)+G4(2) chord (m16) used to read 3/2/1 top-to-bottom (voice order); it now reads 2/3/1 — the highest note's "2" on top, matching the notes. Same for m10/11/15:

    Before:
    hayden before

    After:
    hayden after

  - `test_string_number_collisions` — one stack changes (m2, the G#4+A♭4 enharmonic unison chord in voice 1 over F4 in voice 2): the lowest note's fingering (F4's "1") now sits at the bottom of the stack, 1/3/1 top-to-bottom instead of 1/1/3. The enharmonic-unison pair keeps its XML order (the sort is stable, ties by half-tone don't move):

    Before:
    strings before

    After:
    strings after


sschmidTU (approved):
Thanks for the fix! LGTM, merging.
Confirmed no undesired visual regressions.

Issue #1554, opened 2024-08-27
is there any other way to lazyload scores?


dajun-TT:
we want to lazyload all scores because some xml files are huge, but we found it seem still have not this function.
![image](https://github.com/user-attachments/assets/ea3b0600-b4f1-4b68-9d88-773e5186a1b1)
can we have any other method?

sschmidTU:
You can use the `drawFromMeasureNumber` and `drawUpToMeasureNumber` options to only draw part of the score in `setOptions()`.
It would definitely be nice to give the option to programmatically draw page by page or only x measures, then y measures, etc., but you're the first one to ask for it. I guess most other people don't have gigantic scores.

sschmidTU:
The problem is also that it's not that simple to just stop drawing at a certain measure and then continue uninterrupted later.
You can't draw half a slur that goes over 4 measures, and then continue drawing it after you stopped at measure 2. You have to know the highest points (skyline) in measure 4 in order to know how to begin drawing the slur.

dajun-TT:
> The problem is also that it's not that simple to just stop drawing at a certain measure and then continue without any extra considerations later.
> You can't draw half a slur that goes over 4 measures, and then continue drawing it after you stopped at measure 2. You have to know the highest points (skyline) in measure 4 in advance in order to know how to begin drawing the slur, and for that you have to basically draw measure 4 before drawing the slur.

thanks, we will try it later

fablau:
Actually, I just posted the same exact question on this thread:

https://github.com/opensheetmusicdisplay/opensheetmusicdisplay/issues/1547#issuecomment-2330286364

Mostly on small devices such as on mobile, that'd avoid unresponsiveness and slowness and an overall better user experience.

Can't that be done at the "page" level instead? As I said earlier: I might want to render the first two pages, then when the user scrolls at the bottom of the second page, OSMD could render the next two pages, and so on similarly to the infinite scrolling of many websites. That'd be really neat!

sschmidTU:
Lazy rendering was implemented in OSMD 2.0. See https://github.com/opensheetmusicdisplay/opensheetmusicdisplay/pull/1690

PR #1690, opened 2026-06-15
Add lazy/incremental rendering option (renderNext API) to render scores "line by line"

### LINKED ISSUE #1604: Draw sheet system-by-system
Can osmd draw system-by-system to reduce first render wait time?

sschmidTU:
Closes #1604, #1554

Render large scores progressively instead of blocking on one full render(), for faster first paint.

E.g. here, after osmd.load(), we only render the first 2 systems ("lines") of the somewhat long Bach Prelude in C major:
image
```js
osmd.load(bachXml).then( () => {
  osmd.renderNext({systems: 2});
} );
```
We could then call `osmd.renderNext({systems: 2});` again to render the next 2 systems.

In our demo, you can try this interactively by enabling lazy/incremental rendering in the UI and loading a large score like Gounod - Meditation or the Bach Prelude. When you scroll down, OSMD will automatically render more (powered by `osmd.enableIncrementalRenderingOnScroll()`).

Works for the endless vertical-scroll page format and for `RenderSingleHorizontalStaffline` (one staffline growing to the right).
Usually you specify the next x `measures` to draw, which will snap to full systems (so it may not be exactly x measures added).
The `systems` option allows you to render the next x systems ("lines"), so you don't need to count and estimate how many measures you need to draw to add vertical height (a new system).
And you should be able to estimate the added height by getting the height of the first music system.
(The `systems` option of course doesn't work for the `RenderSingleHorizontalStaffline` option, as that only has one system)

Public API (OpenSheetMusicDisplay):
- renderNext(options?: { measures?: number, systems?: number }): IRenderNextResult -- render and append the next batch. By default it advances ~8 visual measures (a multi-rest counts as one); pass `systems` to advance by whole music systems (lines) instead, ending each batch exactly on a system boundary (#1604). `systems` is vertical only -- a single horizontal staffline is one system, so it ignores `systems` and uses `measures`. The first call auto-starts a fresh session (clears, lays out from measure 0); later calls append. Returns progress { done, renderedMeasures, totalMeasures, lastRenderedMeasure, nextUnrenderedMeasure } where the measure handles are the GraphicalMeasure[] across all staves at that position.
- renderRemaining() -- finish rendering the score synchronously (e.g. before PDF/image export).
- resetIncrementalRendering() -- abandon a session; auto-called by render()/reset().
- IncrementalRenderingActive / IncrementalRenderingComplete / IncrementalRenderProgress.
- enableIncrementalRenderingOnScroll(options? & { scrollElement? }) / disableIncrementalRenderingOnScroll() -- passive scroll-to-load near the edge (page bottom for vertical, right edge for single-horizontal), rAF fill, and auto-detach on completion.

How it works:
- Vertical (renderAppendGrowing): re-lays the whole prefix each batch (the greedy builder is forward-stable), draws only the new systems, defers the prefix's last (unstretched) system, and reconciles if a growing layout moves an earlier system.
- Horizontal (renderAppendGrowingHorizontal): single growing system at SheetMaximumWidth; single-staff grows the laid-out prefix, multi-staff lays the whole score out once and reuses it (only the drawn x-window grows) so lower stafflines never drift.
- Skyline reuse cache (gated by EngravingRules.LazyConsistentGraphic) reuses stable interior systems' sky/bottom lines across batches, turning the per-batch O(prefix) skyline pass into O(new systems). Off for a normal render().
- Drawer x-window / system-index gates draw each element exactly once; inert at their +/-Infinity / -1 defaults, so a normal render() is unaffected.
- Canvas backend snapshots and redraws its bitmap on resize (a canvas resize clears it); the SVG backend grows in place.

All gates and caches are gated behind LazyConsistentGraphic (forced false by render()) or are no-ops at their defaults, so normal rendering is unchanged. Demo: an "Incremental rendering (load on scroll)" toggle plus a renderRemaining() guard before PDF export.

Verified renderNext() output multiset-identical to a normal render() on Bach / Gounod / Actor, both vertical and single-horizontal.

### DIFF
--- a/demo/index.html
+++ b/demo/index.html
@@ -89,6 +89,12 @@
                             <label>Follow Cursor</label>
                         </div>
                     </div>
+                    <div class="item" style="margin-top: 5px;">
+                        <div class="ui toggle checkbox">
+                            <input type="checkbox" name="public" id="incremental-checkbox">
+                            <label>Incremental rendering (load on scroll)</label>
+                        </div>
+                    </div>
                 </div>
             </div>
 
--- a/demo/index.js
+++ b/demo/index.js
@@ -94,6 +94,7 @@ import { TransposeCalculator } from '../src/Plugins/Transpose/TransposeCalculato
         nextCursorBtn,
         resetCursorBtn,
         followCursorCheckbox,
+        incrementalCheckbox,
         showCursorBtn,
         hideCursorBtn,
         debugReRenderBtn,
@@ -233,6 +234,7 @@ import { TransposeCalculator } from '../src/Plugins/Transpose/TransposeCalculato
         nextCursorBtn = document.getElementById("next-cursor-btn");
         resetCursorBtn = document.getElementById("reset-cursor-btn");
         followCursorCheckbox = document.getElementById("follow-cursor-checkbox");
+        incrementalCheckbox = document.getElementById("incremental-checkbox");
         showCursorBtn = document.getElementById("show-cursor-btn");
         hideCursorBtn = document.getElementById("hide-cursor-btn");
         debugReRenderBtn = document.getElementById("debug-re-render-btn");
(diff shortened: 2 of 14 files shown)


fablau:
This update is great!

PR #1695, opened 2026-07-12
fix(Ties): match tied notes by sounding pitch instead of letter name (#1694)

### LINKED ISSUE #1694: Enharmonic ties (e.g. F♯ tied to G♭) are dropped and can mis-bind to a distant note. Tie matching compares letter name, not pitch
## Environment                                                                                                                                                                                                                                        
                                                                                                                                                                                                                                                     
  - opensheetmusicdisplay: 2.0.0 (also reproduced on 1.9.9, the relevant code is identical between the two)                                                                                                                                         
  - Loaded via the standard OpenSheetMusicDisplay(container).load(xml).render() path; default EngravingRules.                                                                                                                                        
  - Source of the files: MuseScore 4 exports (MusicXML), which freely emit ties across enharmonic respellings.  

## Summary                                                                                                                                                                                                                                            
                                                                                                                                                                                                                                                     
  When a tie connects two notes that are the same sounding pitch but spelled differently (F♯ ↔ G♭, E♯ ↔ F, C♭ ↔ B, …), OSMD fails to match the tie's stop to its start. Tie matching keys on the note's letter name (FundamentalNote) + octave rather
  than the actual pitch, so F♯ (letter F) and G♭ (letter G) are treated as unrelated.                                                                                                                                                                
                                                                                                                                                                                                                                                     
  Two observable consequences:                                                                                                                                                                                                                       
  1. Minimal case: the tie is simply not rendered.                                                                                                                                                                                                   
  2. Aggravated case: the unmatched start tie stays open in the tie dictionary and later binds to the next note that shares the same letter name — often several measures away. That produces a spurious long tie (rendered as a stray curve sweeping
  across measures / under a later system's clef when it crosses a system break), and the wrongly-tied note is sustained incorrectly during playback.      
                                                                                           
                                                                                                                                                                                                                                                     
  ### Steps to reproduce       

Render this file (single staff; F♯4 tied across the barline to its enharmonic equivalent G♭4):                                                                                                                                                     
                                                                                                                                                                                                                                                     
```xml
  <?xml version="1.0" encoding="UTF-8"?>                                                                                                                                                                                                             
  <!DOCTYPE score-partwise PUBLIC "-//Recordare//DTD MusicXML 4.0 Partwise//EN" "http://www.musicxml.org/dtds/partwise.dtd">                                                                                                                         
  <score-partwise version="4.0">                                                                                                                                                                                                                     
    <part-list><score-part id="P1"><part-name>Music</part-name></score-part></part-list>                                                                                                                                                             
    <part id="P1">                                                                                                                                                                                                                                   
      <measure number="1">                                                                                                                                                                                                                           
        <attributes><divisions>1</divisions><key><fifths>0</fifths></key><time><beats>4</beats><beat-type>4</beat-type></time><clef><sign>G</sign><line>2</line></clef></attributes>                                                                 
        <note><rest/><duration>2</duration><type>half</type></note>                                                                                                                                                                                  
        <note><pitch><step>F</step><alter>1</alter><octave>4</octave></pitch><duration>2</duration><type>half</type><tie type="start"/><notations><tied type="start"/></notations></note>                                                            
      </measure>                                                                                                                                                                                                                                     
      <measure number="2">                                                                                                                                                                                                                           
        <note><pitch><step>G</step><alter>-1</alter><octave>4</octave></pitch><duration>2</duration><type>half</type><tie type="stop"/><notations><tied type="stop"/></notations></note>                                                             
        <note><rest/><duration>2</duration><type>half</type></note>                                                                                                                                                                                  
      </measure>                                                                                                                                                                                                                                     
    </part>                                                                                                                                                                                                                                          
  </score-partwise>        
```


### Expected                                                                                                                                                                                                                                           
                                                                                                                                                                                                                                                     
  A tie curve connecting the F♯4 in m1 to the G♭4 in m2 (same sounding pitch).                                                                                                                                                                       
                           

<img width="527" height="200" alt="Image" src="https://github.com/user-attachments/assets/54755a5a-47df-4910-910d-37e75c5b4916" />

                                                                                                                                                                                                                          
###  Actual                                                                                                                                                                                                                                             
                                                                                                                                                                                                                                                     
  No tie is rendered (0 tie elements in the SVG). The <tied type="stop"/> on G♭4 finds no open tie of letter G and is discarded; the F♯4 tie remains open.                                                                                           
                                   
<img width="658" height="215" alt="Image" src="https://github.com/user-attachments/assets/aad47198-05b2-4038-ba6f-11cf1d8ffdf3" />


                                                                                                                                                                                                                  
####  Root cause                                                                                                           

> VoiceGenerator.findCurrentNoteInTieDict matches an open tie to the current note by comparing letter name + octave only:                                                                                                                            


```javascript
                                                                                                                                                                                                                                                     
  if (openTie.Pitch.FundamentalNote === note.Pitch.FundamentalNote &&                                                                                                                                                                                
      openTie.Pitch.Octave === note.Pitch.Octave) {                                                                                                                                                                                                  
    return tieNumber;                                                                                                                                                                                                                                
  } 

```

Because FundamentalNote is the diatonic letter (F vs G), enharmonic partners never match. The stop is dropped, the start dangles, and a later same-letter note (e.g. a plain F in a subsequent measure) becomes the accidental match - the source  
  of the "tie jumps several bars" artifact.                                                                                                                                                                                                          
       

## Suggested fix                                                                                                                                                                                                                                      
                                                                                                                                                                                                                                                     
  Keep the exact letter+octave match as the primary pass, then fall back to comparing the absolute half-tone so enharmonic spellings match:                                                                                                          
           
```javascript
                                                                                                                                                                                                                                                            
  // after the exact-match loop returns nothing:                                                                                                                                                                                                     
  for (const key in this.openTieDict) {                                                                                                                                                                                                              
    const openTie = this.openTieDict[key];                                                                                                                                                                                                           
    if (openTie.Pitch.getHalfTone() === note.Pitch.getHalfTone()) {                                                                                                                                                                                  
      return parseInt(key, 10);                                                                                                                                                                                                                      
    }                                                                                                                                                                                                                                                
  }                                                                                                                                                                                                                                                  
  return -1;                                                                                                                                                                                                                                         
```
                                                                                                                                                                                                                                                    
  Making it a fallback (rather than replacing the letter comparison) preserves current behavior whenever an exact-spelling partner exists, and only rescues the enharmonic case. Happy to open a PR if that direction looks right.                   
   

##  Notes                                                                                                                                                                                                                                              
                                                                                                                                                                                                                                                     
  - Reproduces identically on 1.9.9 and 2.0.0.                                                                                                                                                                                                       
  - Real-world trigger: MuseScore emits enharmonic ties around chromatic passages / key changes, so this shows up in ordinary user files, not just synthetic ones.

ymxlx:
## What

Fixes #1694 — ties between **enharmonic spellings** (e.g. F♯ tied to G♭, as MuseScore 4 commonly exports) are dropped, or worse mis-bound to a distant note, because tie matching compared note **letter name** (`FundamentalNote`) instead of **sounding pitch**.

## How

`findCurrentNoteInTieDict()` (`VoiceGenerator.ts`) keeps its existing exact match first — `FundamentalNote` + `Octave` (plus the tab-string fallback). Only if that finds nothing does it now run a second pass matching on `Pitch.getHalfTone()` (an absolute, octave-inclusive semitone value). Because it's fallback-only, every normal (non-enharmonic) tie is matched exactly as before — no behavioral change to existing scores.

`Tie.ts` gets a doc-comment update to describe ties by sounding pitch. No public API change.

## Tests

New fixture `test/data/test_tie_enharmonic_spelling_1694.musicxml`: F♯4 (`step=F alter=1 octave=4`) tied across a barline to G♭4 (`step=G alter=-1 octave=4`) — both compute to half-tone 54, but differ in `FundamentalNote` (F vs G). Two tests added:
- a domain-layer `MusicSheetReader` test that the enharmonic notes are tied, and
- a `VexFlowMeasure` render test that an actual tie curve is emitted in the SVG.

## Verification

- `npm run build` (eslint + webpack) — clean.
- `npm test` — full suite green (one unrelated `SlurFlattenToObstacle` assertion fails identically on a clean `develop`, i.e. pre-existing/flaky).
- Both new tests fail on pre-fix code — in fact, without the fix the reader **spins in an infinite loop** on this fixture, so the fix also prevents a hang, not just a mis-tie.

Branch targets `develop` per the contributing conventions.

### DIFF
--- a/src/MusicalScore/Graphical/AccidentalCalculator.ts
+++ b/src/MusicalScore/Graphical/AccidentalCalculator.ts
@@ -67,17 +67,18 @@ export class AccidentalCalculator {
 
         const tie: Tie = graphicalNote.sourceNote.NoteTie;
         if (tie && graphicalNote.sourceNote !== tie.StartNote) {
-            // A continued tie note normally repeats the start note's pitch, so its accidental is
-            // suppressed (it's the same sounding-and-spelled note held on). But an enharmonic tie
-            // (e.g. F#–Gb, see #1694) connects two different spellings of one sounding pitch: the
-            // continued note carries its own accidental and must still render it. Only suppress
-            // when the spelling actually matches.
+            // The continued note of a tie carries its accidental over implicitly, so it's
+            // normally suppressed. The exception is an enharmonic tie (e.g. F#–Gb, #1694),
+            // which respells one sounding pitch with a *different letter name* — that note must
+            // draw its own accidental. Distinguish by letter (FundamentalNote), not by the
+            // accidental enum: a natural-to-natural tie across a barline (e.g. B♮ tied to B under
+            // a B♭ key signature) is the same written note and its NATURAL/NONE enums may differ,
+            // yet it must stay suppressed (#1695 review). Only a different letter falls through.
             const startPitch: Pitch = tie.StartNote.Pitch;
-            const sameSpelling: boolean = startPitch !== undefined
-                && startPitch.FundamentalNote === pitch.FundamentalNote
-                && startPitch.Accidental === pitch.Accidental;
-            if (sameSpelling) {
-                return; // don't add accidentals on a continued tie note of the same spelling
+            const sameLetter: boolean = startPitch !== undefined
+                && startPitch.FundamentalNote === pitch.FundamentalNote;
+            if (sameLetter) {
+                return; // same written note held across the tie — accidental is implied
             }
         }
 
--- a/src/MusicalScore/Graphical/AccidentalCalculator.ts
+++ b/src/MusicalScore/Graphical/AccidentalCalculator.ts
@@ -67,8 +67,18 @@ export class AccidentalCalculator {
 
         const tie: Tie = graphicalNote.sourceNote.NoteTie;
         if (tie && graphicalNote.sourceNote !== tie.StartNote) {
-            return; // don't add accidentals on continued tie note
-            // note that continued tie notes may incorrectly use the same pitch object, with the same accidentalXml value.
+            // A continued tie note normally repeats the start note's pitch, so its accidental is
+            // suppressed (it's the same sounding-and-spelled note held on). But an enharmonic tie
+            // (e.g. F#–Gb, see #1694) connects two different spellings of one sounding pitch: the
+            // continued note carries its own accidental and must still render it. Only suppress
+            // when the spelling actually matches.
+            const startPitch: Pitch = tie.StartNote.Pitch;
+            const sameSpelling: boolean = startPitch !== undefined
+                && startPitch.FundamentalNote === pitch.FundamentalNote
+                && startPitch.Accidental === pitch.Accidental;
+            if (sameSpelling) {
+                return; // don't add accidentals on a continued tie note of the same spelling
+            }
         }
 
         const isInCurrentAlterationsToKeyList: boolean = this.currentAlterationsComparedToKeyInstructionList.indexOf(pitchKey) >= 0;
--- a/src/MusicalScore/ScoreIO/VoiceGenerator.ts
+++ b/src/MusicalScore/ScoreIO/VoiceGenerator.ts
@@ -1204,7 +1204,8 @@ export class VoiceGenerator {
   }
 
   /**
-   * Search the tieDictionary for the corresponding candidateNote to the currentNote (same FundamentalNote && Octave).
+   * Search the tieDictionary for the corresponding candidateNote to the currentNote.
+   * Prefer the existing spelling/string match, then fall back to sounding pitch for enharmonic ties.
    * @param candidateNote
    * @returns {number}
    */
@@ -1221,6 +1222,14 @@ export class VoiceGenerator {
           if (tieTabNote.StringNumberTab === tieCandidateNote.StringNumberTab) {
             return parseInt(key, 10);
           }
+        }
+      }
+    }
+    for (const key in openTieDict) {
+      if (openTieDict.hasOwnProperty(key)) {
+        const tie: Tie = openTieDict[key];
+        if (tie.Pitch.getHalfTone() === candidateNote.Pitch.getHalfTone()) {
+          return parseInt(key, 10);
         }
       }
     }
--- a/src/MusicalScore/VoiceData/Tie.ts
+++ b/src/MusicalScore/VoiceData/Tie.ts
@@ -6,7 +6,7 @@ import { PlacementEnum } from "../VoiceData/Expressions/AbstractExpression";
 import log from "loglevel";
 
 /**
- * A [[Tie]] connects two notes of the same pitch and name, indicating that they have to be played as a single note.
+ * A [[Tie]] connects two notes of the same sounding pitch, indicating that they have to be played as a single note.
  */
 export class Tie {
 
--- a/test/MusicalScore/Graphical/VexFlow/VexFlowMeasure_Test.ts
+++ b/test/MusicalScore/Graphical/VexFlow/VexFlowMeasure_Test.ts
@@ -23,6 +23,7 @@ import { OctaveEnum } from "../../../../src/MusicalScore/VoiceData/Expressions/C
 import { Tuplet } from "../../../../src/MusicalScore/VoiceData/Tuplet";
 import { Note } from "../../../../src/MusicalScore/VoiceData/Note";
 import { PointF2D } from "../../../../src/Common/DataObjects/PointF2D";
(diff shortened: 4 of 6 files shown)


sschmidTU (changes requested):
Thanks for the PR!
The tie creation works now, but the second note loses its accidental, which is not correct.
It looks like a slur from F# to G instead of a tie from F# to Gb.
Before:
image

After:
image

You can take another try at fixing this if you want or I can also have a go myself.

ymxlx:
Thanks for the clear before/after — that pinned it exactly.

**Root cause:** `AccidentalCalculator.checkAccidental` bailed out early on *any* continued tie note (`if (tie && sourceNote !== tie.StartNote) return;`) to avoid re-printing the accidental. That's right for an ordinary tie, where the second note is the same spelling held on — but an enharmonic tie (F#–G♭) connects two *different* spellings of one sounding pitch, so the continued note genuinely has its own accidental to show. The blanket suppression is what turned the G♭ into a bare G.

**Fix:** only suppress when the two tied notes share fundamental note *and* accidental; otherwise fall through and let the continued note draw its own accidental as normal. Same-spelling ties are completely unchanged (they still take the early return).

**Verification:** I extended the `#1694` render test to assert the tied end note keeps its accidental (`GraphicalTie.EndNote.DrawnAccidental === AccidentalEnum.FLAT`) — it fails on the old blanket-suppression behavior and passes now. Rendering the fixture confirms F#–G♭ now shows the flat on the second note while the start note keeps its sharp.

Pushed as 36c6e8b. Let me know if you'd prefer the spelling comparison to also cover the double-accidental edge cases (e.g. Fx–G) — happy to broaden it.

sschmidTU (changes requested):
The missing enharmonic accidental is fixed, but now there's a new regression/breaking change:
When a last note in a measure with a natural accidental is tied to the first note in the next measure that would normally have an implied accidental from the key signature, it's not necessary to display a natural accidental on the second note, because the natural is implied from the tie.

E.g. in `test/data/TelemannWV40.102_Sonate-Nr.1.2-Allegro-F-Dur.xml` measure 9 to 10, we have a Bb key signature, a B natural with natural accidental at the end of measure 9 tied to a B (natural) in measure 10. The most recent change displays a natural accidental for the measure 10 note.
Before:
image

After:
image

This is not wrong for the musical meaning, but it's also not necessary, and we didn't display it before, so it's a breaking change. If we had always displayed it it wouldn't be a problem, but we'd like to avoid unnecessary breaking changes.
MuseScore also does not display the natural accidentals here. They should probably only be displayed if they're explicitly given in the MusicXML (as a kind of courtesy accidental), though that should be rare, as tied notes are intuitively (for the musician) / implicitly (from notation rules) understood to carry over the accidental.

Same applies to `ScottJoplin_EliteSyncopations.xml` measure 83 to 84, but it should be fixed by the same code fix that will be needed for the Telemann.

I found this with the visual regression tests by the way, which you can use on your own commits to check regressions.
(npm run build && npm run generate:blessed on the develop branch, then npm run test:visual:build on the PR branch, assuming npm install isn't necessary because dependencies are unchanged)

ymxlx:
Good catch, and thanks for naming the exact measures — that made it quick to pin.

**Why it regressed:** my first pass compared *both* letter name and accidental enum to decide "same spelling". But a tie always joins one sounding pitch, so the only thing that can vary is the spelling, and a same-letter tie can still differ in the accidental enum — the explicit `B♮` at the end of Telemann m.9 carries `NATURAL`, while the tied `B` in m.10 comes through as `NONE`. That enum mismatch made my check treat it as a respelling and fall through, drawing the unnecessary natural.

**Fix (d768eb0):** compare by **letter name (`FundamentalNote`) only**. Same letter ⇒ the same written note held across the tie ⇒ stay suppressed, exactly as before. A different letter ⇒ a real enharmonic respelling (F#–Gb) ⇒ the continued note draws its own accidental. So same-letter ties — including natural-to-natural across a barline — are back to the original behavior.

**Verified:**
- Telemann m.9→10 and Joplin m.83→84: continued tied notes no longer draw a natural (checked every tie in both samples — zero continued naturals).
- The F#–Gb case still draws its flat.
- Added a Telemann-based regression test asserting no continued tie note redraws a natural, next to the existing enharmonic test.

I verified the two regressions and the enharmonic case at the model level (`GraphicalTie.EndNote.DrawnAccidental`) via a headless render rather than the full visual-regression suite — my local environment doesn't have Chrome/Firefox for karma — so a confirmation against your visual baselines would be very welcome. Thanks again for the thorough review.

sschmidTU (approved):
Good, no more visual regressions, thanks!
LGTM, merging.

Issue #1439, opened 2023-08-10
BUG: Octave shift crashes on render when some instrument is hidden


rlisbeth:
Hello OSMD team,

I think I have spotted another bug in the octave shift code : sometimes I get a crash when I render my .mxl file which contains piano scales on 4 octaves with left and right hand in different Instruments and the right hand is hidden. This crash is quite random, doesn't happen every time, but I guess it is related to the width of the viewport since I manage to get it systematically when resizing the window.

I provide a minimal code for you to reproduce the bug in a [zip file](https://github.com/opensheetmusicdisplay/opensheetmusicdisplay/files/12311587/osmd-issue.zip). The code is pretty simple :

```html
<div id="osmdContainer" />
<script src="opensheetmusicdisplay.min.js"></script>
<script>
    const osmd = new opensheetmusicdisplay.OpenSheetMusicDisplay("osmdContainer");
    osmd.setOptions({
        backend: "svg",
        drawTitle: true,
        drawingParameters: "compacttight" // don't display title, composer etc., smaller margins
    });
    osmd
        .load("./Scale_sixteenth.mxl")
        .then(
            function () {
                osmd.sheet.Instruments[0].Visible = false
                osmd.render();
            }
        );
</script>
```

I have built a debug version of osmd (included in zip) to get a more readable stacktrace:

```
VexFlowMusicSheetCalculator.ts:988 Uncaught TypeError: Cannot read properties of undefined (reading 'Measures')
    at VexFlowMusicSheetCalculator.calculateSingleOctaveShift (VexFlowMusicSheetCalculator.ts:988:80)
    at VexFlowMusicSheetCalculator.calculateOctaveShifts (MusicSheetCalculator.ts:3404:34)
    at VexFlowMusicSheetCalculator.calculateMusicSystems (MusicSheetCalculator.ts:948:18)
    at VexFlowMusicSheetCalculator.calculate (MusicSheetCalculator.ts:294:14)
    at GraphicalMusicSheet.reCalculate (GraphicalMusicSheet.ts:195:25)
    at OpenSheetMusicDisplay.render (OpenSheetMusicDisplay.ts:230:22)
    at OpenSheetMusicDisplay.ts:741:26
    at resizeEnd (OpenSheetMusicDisplay.ts:776:17)
```
If you don't get the crash at first launch, please try to resize the window. My guess is that the crash could be related to the number of lines breaks used to display the sheet.

Thanks for your amazing work !

[...]

PR #1696, opened 2026-07-15
fix: don't crash on unterminated octave-shift (Fraction.lte on undefined endTimeStamp)

recrack:
## Problem

`osmd.render()` throws `TypeError: Cannot read properties of undefined (reading 'realValue')` and aborts rendering the whole sheet when the MusicXML contains an **octave-shift start without a matching stop**. Such files are produced in the wild by OMR tools (e.g. Audiveris exports `<octave-shift type="down">` and never closes it).

```
TypeError: Cannot read properties of undefined (reading 'realValue')
    at Fraction.lte
    at VexFlowMusicSheetCalculator.calculateSingleOctaveShift
    at MusicSheetCalculator.calculateOctaveShifts
    at MusicSheetCalculator.calculateMusicSystems
```

Minimal trigger inside any part (never followed by a `stop`):

```xml
<direction placement="above">
  <direction-type>
    <octave-shift type="down" number="1" size="8"/>
  </direction-type>
</direction>
```

## Cause

For an unterminated shift, `octaveShift.ParentEndMultiExpression` is undefined, so `endTimeStamp` is undefined. The code already clamps `endMeasure` to the last rendered measure for this case, but the "latest staff entry at or before the end timestamp" fallback loop then calls `entry.relInMeasureTimestamp?.lte(endTimeStamp)` — and `Fraction.lte` dereferences its argument unconditionally (unlike `Fraction.Equals`, which is null-safe).

## Fix

Guard the fallback search loop on `endTimeStamp`. When it is undefined the loop is skipped and the existing `endMeasure.staffEntries[last]` fallback right below handles the unterminated case, so the shift is drawn up to the last rendered measure instead of crashing.

See also earlier octave-shift robustness work in #1439 and #1376.

## Verification

- Reproduced with a 13-page Audiveris-exported score (2 unterminated `type="down"` shifts): crash on 2.0.0 → renders fine with this patch (headless jsdom + browser).
- Removing the two unterminated octave-shift elements from the same file also renders fine on unpatched 2.0.0, confirming the trigger.
- `npm run lint` passes; webpack build succeeds.

### DIFF
--- a/src/MusicalScore/Graphical/VexFlow/VexFlowMusicSheetCalculator.ts
+++ b/src/MusicalScore/Graphical/VexFlow/VexFlowMusicSheetCalculator.ts
@@ -1290,7 +1290,9 @@ export class VexFlowMusicSheetCalculator extends MusicSheetCalculator {
         startStaffEntry = startMeasure.staffEntries[0];
       }
       let endStaffEntry: GraphicalStaffEntry = endMeasure.findGraphicalStaffEntryFromTimestamp(endTimeStamp);
-      if (!endStaffEntry) {
+      if (!endStaffEntry && endTimeStamp) {
+        // endTimeStamp can be undefined for an unterminated octave-shift (start without stop,
+        //   e.g. from OMR-generated MusicXML), see #1439 / #1376 for similar cases.
         // No exact match (e.g. pending stop with computed inclusive end).
         // Find the latest staff entry at or before the end timestamp.
         for (let i: number = endMeasure.staffEntries.length - 1; i >= 0; i--) {


sschmidTU (approved):
LGTM, merging.
Thanks for the PR, nice find!
I'll add a new test sample after merging.
No visual regressions.

If possible, in future PRs and issues, instead of MusicXML excerpts, share full MusicXML files, so that people don't have to copy the MusicXML excerpt, find a suitable existing MusicXML file to copy, then copy and paste the excerpt into at the right spot, create a new filename and MusicXML work-title element, and so on.

sschmidTU:
ps: I'll need to remember to change the changelog message to fix(OctaveShift) instead of fix(VexFlowMusicSheetCalculator), forgot to amend the commit message for this part while squashing

Issue #910, opened 2020-10-29
Notes misaligned after hidden dotted rest


pjcdawkins:
OSMD version: 0.8.6

XML: https://fourscoreandmore.org/apps/download/scores/chorales/J_S_Bach_BWV_2_6-cut2-fromATB-exercise.xml?download=1
(raw: [J_S_Bach_BWV_2_6-cut2-fromATB-exercise.xml.txt](https://github.com/opensheetmusicdisplay/opensheetmusicdisplay/files/5457550/J_S_Bach_BWV_2_6-cut2-fromATB-exercise.xml.txt))

Displayed with OSMD: https://fourscoreandmore.org/apps/score/chorales/J_S_Bach_BWV_2_6-cut2-fromATB-exercise.xml
![Screenshot from 2020-10-29 15-38-20](https://user-images.githubusercontent.com/1465106/97596628-d15e4000-19fc-11eb-8e3f-e543aba579be.png)
(super impressed with the OSMD project by the way!)

As you can see the harmony in parts 2-4 (measures 2, 4) is misaligned after the rests, resulting in unusual chords for Bach.

Desired behaviour (displayed with MuseScore):
![Screenshot from 2020-10-29 15-37-32](https://user-images.githubusercontent.com/1465106/97596504-b12e8100-19fc-11eb-92cb-dfceb3624042.png)

The rests are created with music21 in @MarkGotham's script here: https://github.com/fourscoreandmore/fourscoreandmore.org/blob/0626d13a235c7b468bf9eeb9551e59a224b953c5/cut-outs-backend/app/TheoryExercises.py#L68

and it seems they look like this

```xml
      <forward>
        <duration>30240</duration>
      </forward>
```

pjcdawkins:
I'm not certain it's only with dotted rests, but it seems that way based on a few other scores, and I wondered if it's a similar problem to #484

sschmidTU:
Interesting. This doesn't happen if you remove the pickup measure, so i don't think it's related to issue 484,
probably an error in Vexflow's formatter.js.
![image](https://user-images.githubusercontent.com/33069673/97787044-f646e480-1baf-11eb-98b5-722594198621.png)
[...]

PR #1697, opened 2026-07-18
fix(layout): align voices when a note's type and dots disagree with its duration

### LINKED ISSUE #1698: Simultaneous notes drawn as sequential when a note's <type> + <dot>s disagree with its <duration> (e.g. invisible padding rests)
**OSMD version:** reproduced on 1.9.7, 1.9.9, 2.0.0 (npm) and current `develop`.

## Describe the bug

When a note or rest is encoded with a `<type>` (plus `<dot>`s) that does not match its actual `<duration>`, every note that follows it in the same voice is drawn horizontally misaligned against the other voices. Two notes that share the same onset end up in separate columns, looking like *sequential* notes instead of a simultaneous interval.

A common real-world source of this encoding is an **invisible padding rest**: e.g. `<rest/>` with `print-object="no"`, `type=half` + `<dot/>` (which implies 3 quarters) but `<duration>` of only 2.5 quarters. Notation editors export this pattern when a padding gap has no clean single-type representation. MuseScore renders such files correctly aligned, since it positions by actual duration.

## Minimal reproduction

Single 3/4 measure, `divisions=2`. Voice 1: invisible rest (`duration=5` = 2.5 quarters, but written `type=half` + dot = 3 quarters), then an eighth F4 on beat 3.5. Voice 2: dotted quarter, quarter, then an eighth G4 also on beat 3.5. The F4 and G4 are simultaneous and should be rendered as a second interval in one column.

```xml
<?xml version="1.0" encoding="UTF-8"?>
<score-partwise version="3.1">
  <part-list>
    <score-part id="P1"><part-name>Music</part-name></score-part>
  </part-list>
  <part id="P1">
    <measure number="1">
      <attributes>
        <divisions>2</divisions>
        <key><fifths>0</fifths></key>
        <time><beats>3</beats><beat-type>4</beat-type></time>
        <clef><sign>G</sign><line>2</line></clef>
      </attributes>
      <!-- voice 1: invisible rest lasting 2.5 quarters, but written as dotted half (3 quarters) -->
      <note print-object="no">
        <rest/>
        <duration>5</duration>
        <voice>1</voice>
        <type>half</type>
        <dot/>
      </note>
      <note>
        <pitch><step>F</step><octave>4</octave></pitch>
        <duration>1</duration>
        <voice>1</voice>
        <type>eighth</type>
        <stem>down</stem>
      </note>
      <backup><duration>6</duration></backup>
      <!-- voice 2: dotted quarter, quarter, eighth (adds up to 3 quarters) -->
      <note>
        <pitch><step>C</step><octave>4</octave></pitch>
        <duration>3</duration>
        <voice>2</voice>
        <type>quarter</type>
        <dot/>
        <stem>up</stem>
      </note>
      <note>
        <pitch><step>C</step><octave>4</octave></pitch>
        <duration>2</duration>
        <voice>2</voice>
        <type>quarter</type>
        <stem>up</stem>
      </note>
      <note>
        <pitch><step>G</step><octave>4</octave></pitch>
        <duration>1</duration>
        <voice>2</voice>
        <type>eighth</type>
        <stem>up</stem>
      </note>
    </measure>
  </part>
</score-partwise>
```

**Expected:** F4 and G4 rendered as one vertical event (second interval, adjacent noteheads offset by about one notehead width, opposite stems).

**Actual (measured with 2.0.0):** the invisible rest gets vexflow ticks for a *dotted half* (12288 ticks = 3 quarters) instead of its actual duration (10240 ticks = 2.5 quarters). The following F4 therefore lands on vexflow tick 3.0 while the simultaneous G4 of voice 2 is on tick 2.5 — they get separate tick contexts and the F4 is drawn ~44px to the right of the G4, past it, as if it were a later note.

## Root cause

`VexFlowConverter.StaveNote()` builds the vexflow duration string from the XML `<type>` + dots whenever `TypeLength` differs from the actual note length. VexFlow then accumulates the voice's tick positions from those durations, so the voice drifts by the type/duration difference. OSMD's own model timestamps are correct (that's why the ghost-note gap filling in `getRestFilledVexFlowStaveNotesPerVoice` doesn't kick in — the model sees no gap), but nothing re-synchronizes the vexflow ticks.

Notably, `VexFlowMeasure.graphicalMeasureCreatedCalculations()` already contains an exact tick correction (via `VF.Fraction` and `graphicalNoteLength`) for this same class of problem — but only for **tuplets**.

Related issue with the same symptom but a different mechanism (`<forward>` ghost notes, already fixed via #1062): #910.

## Proposed fix

PR #1697 generalizes the existing tuplet tick correction to any note whose vexflow ticks disagree with its actual graphical length (whole-measure rests and grace notes excluded). With that change the rest gets 10240 ticks and F4/G4 share one tick context, rendering as a proper second interval.

gifflet:
Fixes #1698.

## The problem

When a note or rest is encoded with a `<type>` (plus `<dot>`s) that does not match its actual `<duration>`, the VexFlow `StaveNote` is built from the type and dots, so the tick counting of that voice drifts out of sync with the other voices. Every subsequent note of the voice is then drawn horizontally misaligned: two simultaneous notes end up in separate tick contexts / columns, looking like *sequential* notes to the reader (and confusing musicians, since the playback cursor correctly treats them as one onset).

A common real-world source of this encoding is an **invisible rest filling an irregular gap** — e.g. a rest with `print-object="no"`, `type=half` + `<dot/>` (3 quarters) but a `duration` of only 2.5 quarters. Notation editors export this pattern when the padding rest has no clean single-type representation. Other renderers (e.g. MuseScore) align these scores correctly because they position by actual duration.

Related (same symptom, different mechanism, already fixed): #910.

## The fix

`VexFlowMeasure.ts` already rewrites VexFlow ticks exactly (via `VF.Fraction` and `graphicalNoteLength`) **for tuplets**. This PR generalizes that correction to any note whose VexFlow ticks disagree with its actual graphical length:

- whole-measure rests keep their vexflow whole-rest ticks (nothing follows them in the voice; correcting them would only churn spacing),
- grace notes are excluded,
- the tuplet branch behaves exactly as before.

14 insertions, 3 deletions, single file.

## Test sample

Adds `test/data/test_voice_alignment_type_duration_mismatch.musicxml`: measure 1 reproduces the bug (invisible dotted-half rest with duration 2.5 quarters, voice 2 with a simultaneous eighth at the last onset); measure 2 is a control with the same music but correctly-encoded rests. **Both measures must render identically** — without the fix, measure 1 draws the two simultaneous eighths (F4/G4) in separate columns.

Happy to adjust guards/naming or extend the sample if you prefer a different scope.

### DIFF
--- a/src/MusicalScore/Graphical/VexFlow/VexFlowMeasure.ts
+++ b/src/MusicalScore/Graphical/VexFlow/VexFlowMeasure.ts
@@ -1424,15 +1424,26 @@ export class VexFlowMeasure extends GraphicalMeasure {
                 // different voices use different normal-type values in their tuplets, the calculated
                 // ticks can differ even for notes at the same timestamp. We use graphicalNoteLength
                 // (which represents the actual musical duration) to calculate correct tick values.
+                // The same correction is applied when the XML <type> (+<dot>s) disagrees with the
+                // actual <duration> (e.g. an invisible rest with type half + dot, but a duration of
+                // only 2.5 quarters): otherwise the vexflow tick counting of the voice goes out of
+                // sync and all following notes of the voice are misaligned with the other voices.
                 if (voiceEntry.notes.length > 0 && voiceEntry.notes[0].sourceNote) {
                     const sourceNote: Note = voiceEntry.notes[0].sourceNote;
-                    if (sourceNote.NoteTuplet) {
-                        const graphicalLength: Fraction = voiceEntry.notes[0].graphicalNoteLength;
+                    const graphicalLength: Fraction = voiceEntry.notes[0].graphicalNoteLength;
+                    const vfTicks: VF.Fraction = vexFlowVoiceEntry.vfStaveNote.getTicks();
+                    // whole measure rests keep their vexflow "w" ticks: nothing follows them
+                    //   in the voice, and correcting them would change spacing unnecessarily.
+                    const isWholeMeasureRest: boolean = sourceNote.IsWholeMeasureRest ||
+                        graphicalLength.RealValue === this.parentSourceMeasure.ActiveTimeSignature.RealValue;
+                    const ticksMismatch: boolean =
+                        Math.abs(vfTicks.value() - graphicalLength.RealValue * VF.RESOLUTION) > 0.001;
+                    if (sourceNote.NoteTuplet ||
+                        (ticksMismatch && !isWholeMeasureRest && !voiceEntry.parentVoiceEntry?.IsGrace)) {
                         // Calculate ticks using VexFlow Fraction to preserve precision.
                         // graphicalLength.RealValue is the note length as a fraction of a whole note.
                         // VF.RESOLUTION (e.g., 16384) is the number of ticks for a whole note.
                         // We use Fraction arithmetic to avoid floating-point precision issues.
-                        const vfTicks: VF.Fraction = vexFlowVoiceEntry.vfStaveNote.getTicks();
                         vfTicks.numerator = graphicalLength.Numerator * VF.RESOLUTION;
                         vfTicks.denominator = graphicalLength.Denominator;
                         // Simplify the fraction to reduce large numbers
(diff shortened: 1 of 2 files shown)


sschmidTU (commented):
LGTM, merging.
Confirmed same display as MuseScore, no harmful visual regressions.

Before:
image

After:
image



Visual regression tests did show this also fixes two issues in Beethoven Moonlight Sonata and Norma fantasy, which is great.

[visual_diff_voice_alignment_fix_pr1697.zip](https://github.com/user-attachments/files/30272200/visual_diff_voice_alignment_fix_pr1697.zip)

Moonlight:
Before:

Image

After:

Image

Norma:
Before:

Image
After:
Image

sschmidTU (approved):

Issue #1681, opened 2026-06-06
Performance improved by 1.8x average (range 1.5x-2.3x) by new geometric SkyBottomLine calculation


sschmidTU:
# ~1.8× faster rendering (1.5–2.3×): skylines are now calculated geometrically, without getImageData

**TL;DR:** `osmd.render()` is now **~1.8× faster on average** (×1.5–×2.3 depending on the score) in real browsers - a large orchestral score drops from **1.6s to 0.7s**.
The skyline calculation, previously the biggest performance cost, is **2–4× faster** (up to 7× vs the WebGL backend on short scores), while producing layouts almost identical to before to within a twentieth of a pixel on average.
As a bonus, staff lines now always render crisp instead of varying in apparent weight from system to system, as they snap to crisp subpixel position.

## Why rendering was slow

To place graphical elements like lyrics, dynamics and ornaments while avoiding collisions and to space systems, OSMD computes a *skyline* and *bottom-line* for every staffline: the upper and lower point/line of everything drawn in it.
Since 2017, this worked by brute force ([#937](https://github.com/opensheetmusicdisplay/opensheetmusicdisplay/issues/937)):

1. draw every measure a second time onto a hidden canvas,
2. read **all of its pixels** back with `getImageData()`,
3. scan every pixel column for the topmost and bottommost ink.

Step 2 is the problematic one: on GPU-accelerated canvases, `getImageData()` stalls the pipeline and copies the framebuffer from the GPU back to the CPU — for every measure of the sheet. Profiling showed the skyline pass taking up to ~96% of render time.

And there was no way around this other than not using `getImageData()` (and thus not getting the necessary pixel data), except using WebGL shaders, which are basically just a fancier and sometimes slightly faster way to `getImageData()`, and on some systems and browsers WebGL was actually slower, so we inelegantly enabled it depending on environment.

The later batch and WebGL backends made the readback cheaper, but still rasterized, transferred and scanned every pixel — and rebuilt the entire WebGL pipeline (context, shader compilation) on **every** render, including every zoom step and resize.
Now, we don't need WebGL at all anymore, as the new method is always the fastest.

<img width="1964" height="394" alt="image" src="https://github.com/user-attachments/assets/a25e4866-1af5-4caa-8359-63ee0c50ad90" />

*Skyline (red), bottomline (blue)*

## What changed

The trick: **VexFlow still "draws" every measure — but onto math instead of pixels.**

[...]

PR #1703, opened 2026-07-31
perf: scope correctNotePositions to its own measure (render 27-57% faster)

ishinomaru:
`VexFlowMeasure.correctNotePositions()` runs once per graphical measure, but its loop walks `Voice.VoiceEntries` — every entry that voice has in the **entire score**. So each measure re-writes the vertical position of every note in the piece, and everything after the first pass writes the same value again.

This changes the loop to walk the measure's own `staffEntries`. `render()` gets 27–57% faster on larger scores, and the SVG output is byte-identical.

## Numbers

Median of 3 runs each, headless Chrome, synthetic scores (4 quarter notes per measure per part):

| Score | Before | After |
|---|---:|---:|
| 6 parts × 50 measures | 685 ms | 501 ms |
| 6 parts × 100 measures | 1222 ms | 854 ms |
| 6 parts × 200 measures | 3793 ms | 1626 ms |
| 24 parts × 200 measures | 11342 ms | 4986 ms |

Scaling exponent over a 4× longer score (`log(t₂₀₀/t₅₀)/log 4`): **1.24 → 0.85**.

The effect is bigger on slow hardware — on a low-end Android phone the same change cut `render()` by 40% for a 102-measure / 6-part score.

## Why it was quadratic

```ts
for (const voice of this.getVoicesWithinMeasure()) {
    for (const ve of voice.VoiceEntries) {   // <-- whole score, not this measure
```

`getVoicesWithinMeasure()` correctly narrows to the voices in this measure, but `Voice.VoiceEntries` expands right back out to the full score.

Profiling 6 parts × 200 measures (4,800 notes), one `render()`. Every one of these calls came from this method and nowhere else:

| | calls | per note |
|---|---:|---:|
| `EngravingRules.GNote` | 1,920,000 | 400 |
| `VexFlowGraphicalNote.notehead` | 3,840,000 | 800 |

400 visits per note is exactly the score length in measures. The method accounted for 41% of render self time with a scaling exponent of 2.08, while every other method measured sat near 1.0. (Skyline calculation is down to 1.5% since #1681 — that change is what left this as the dominant term.)

Walking `staffEntries` also drops the `rules.GNote()` lookups entirely, since `graphicalVoiceEntries` already holds the `GraphicalNote` objects.

## Verification

Both revisions built with `npm run build:webpack` and rendered in headless Chrome; the SVG is serialised and hashed (FNV-1a), so a match means identical output. **95 scores, all identical, zero differences:**

- 94 from `test/data`, picked for constructs this could break: tablature, drums, grace notes, multiple voices, chords, invisible notes, in-measure clef changes, multi-rests, octave shifts, tuplets, fingerings, arpeggios, plus `ActorPreludeSample`, Haydn, Gounod, Joplin, Bach, Clementi
- 1 purpose-built guitar + tablature score, see note 3 below

Also identical with `renderSingleHorizontalStaffline: true` on the four synthetic scores above — the 24 × 200 one matched across 30,704,121 characters of SVG.

`npm test` is unchanged: 361 passing, 2 skipped, and one slur test (`SlurFlattenToObstacle, width-driven`) failing — it fails identically on unmodified `develop`, with the same numbers to 15 decimal places, when run in Chrome instead of the configured Firefox.

## Three things worth your judgement

1. **`continue` instead of `return` on a missing `vfnote`.** The existing `return` abandons the whole loop — which, in the whole-score loop, meant the rest of the score. Skipping just that note looks like the intent, and it lines up with the existing `TODO fix multi rest measure not removed` comment. A variant keeping `return` also produces identical output on all 95 scores, if you prefer the smaller semantic change.

2. **The new `lastNote.vfnote` check is required.** `notehead()` with no argument dereferences `this.vfnote[0]`, and a staff entry holding only grace notes never gets a `StaveNote`. The old code got away with it because it bailed out of the entire loop at the first such note; skipping per entry exposes it, and `draw()` has no `try`/`catch`.

3. **For a part with both a standard and a tablature staff, a different `GraphicalNote` object gets adjusted.** `NoteToGraphicalNoteMap` holds the tab-side one (created last, overwrites the id), so `rules.GNote()` returned that while walking `staffEntries` reaches the standard-staff one. VexFlow owns the drawn coordinates, so the SVG is unaffected — I checked with a score using `<staves>2</staves>` plus a 6-line tab staff, since every tablature fixture in `test/data` is tab-only and never exercises this path. But if anything downstream reads `GraphicalNote.PositionAndShape` for these notes, it now sees the offset on the standard-staff object. That seems more correct to me, but you would know if something depends on the current arrangement.

The tab-measure branch has the same `Voice.VoiceEntries` problem and can take the same treatment. I left it alone rather than change tab layout without fixtures to check against.

### DIFF
--- a/src/MusicalScore/Graphical/VexFlow/VexFlowMeasure.ts
+++ b/src/MusicalScore/Graphical/VexFlow/VexFlowMeasure.ts
@@ -768,27 +768,39 @@ export class VexFlowMeasure extends GraphicalMeasure {
             }
             return; // don't do the below y position adaptations meant for non-tab notes
         }
-        for (const voice of this.getVoicesWithinMeasure()) {
-            for (const ve of voice.VoiceEntries) {
-                for (const note of ve.Notes) {
-                    if (note.isRest()) {
+        // Iterate this measure's own staff entries. Going through
+        // Voice.VoiceEntries instead would walk every entry that voice has in
+        // the whole score, from a method that runs once per measure, so each
+        // measure would re-position every note in the piece (quadratic in
+        // measure count, and every pass after the first writes the same value).
+        for (const gse of this.staffEntries) {
+            for (const gve of gse.graphicalVoiceEntries) {
+                const notes: GraphicalNote[] = gve.notes;
+                if (notes.length === 0) {
+                    continue;
+                }
+                const lastNote: VexFlowGraphicalNote = notes[notes.length - 1] as VexFlowGraphicalNote;
+                if (!lastNote.vfnote) { // notehead() below reads its vfnote[0] with no argument
+                    continue;
+                }
+                for (const graphicalNote of notes) {
+                    const gNote: VexFlowGraphicalNote = graphicalNote as VexFlowGraphicalNote;
+                    if (gNote.sourceNote.isRest()) {
                         continue;
                         // rest positions are already fine.
                         // Doing the below for rests messes up the y-position calculation / bbox for non-rest note for some reason.
                         //   they were not in the voice's voice entries until #1612, so it didn't matter.
                     }
-                    const gNote: VexFlowGraphicalNote = this.rules.GNote(note) as VexFlowGraphicalNote;
-                    if (!gNote?.vfnote) { // can happen were invisible, then multi rest measure. TODO fix multi rest measure not removed
-                        return;
+                    if (!gNote.vfnote) { // can happen were invisible, then multi rest measure. TODO fix multi rest measure not removed
+                        continue;
                     }
                     const vfnote: VF.StemmableNote = gNote.vfnote[0];
-                    // if (note.isRest()) // TODO somehow there are never rest notes in ve.Notes
-                    // TODO also, grace notes are not included here, need to be fixed as well. (and a few triple beamed notes in Bach Air)
+                    // TODO grace notes are not included here, need to be fixed as well. (and a few triple beamed notes in Bach Air)
                     let relPosY: number = 0;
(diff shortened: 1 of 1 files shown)


phonicscore (approved):
(approved by @sschmidTU )

sschmidTU:
LGTM, merging.
No significant visual regressions. (mostly some improved bounding boxes of grace notes)

Good catch, thanks for the fix!

Could you post the `<staves>2</staves>` sample with tab+classical staff here (zip it or rename to .txt)? That would be helpful. Though if that's private and not easy to create a short extract of, I can also create a new sample.

I'll add some follow-up commits like also applying this to tab notes, and removing the stale TODO grace notes comment.

sschmidTU:
Performance improvement table:
image

csv:
[pr1703_perf_table.csv](https://github.com/user-attachments/files/30754406/pr1703_perf_table.csv)

Interestingly, the Actor prelude has only a ~7% speedup despite being one of the samples with the most notes, but apparently it's so dense with content that this bottleneck was relatively smaller for it.
But in general, as you said, the more notes a piece has, the bigger the speedup.

The three judgement issues you mentioned:
1. Yes, we'll keep continue, it's more correct.
2. Keep the lastNote.vfnote check. It's safer and might be needed with dual staff/tab scores
3. Yes, sounds more correct now, but I'll have to check with a tab+classical sample we don't have yet.

ishinomaru:
@sschmidTU 
Attaching the sample (renamed to .txt, GitHub doesn't take .musicxml). It's hand-written from scratch, so feel free to add it to test/data as-is.
[synth_guitar_plus_tab.musicxml.txt](https://github.com/user-attachments/files/30754830/synth_guitar_plus_tab.musicxml.txt)

Issue #676, opened 2020-02-19
Visual Regression Tests implemented


sschmidTU:
I just implemented Visual Regression Testing, allowing you to quickly generate png images, and automatically detect visual changes between OSMD renders of all samples. Now we don't have to manually check lots of samples anymore.

![image](https://user-images.githubusercontent.com/33069673/74853278-ffec2d00-533d-11ea-9fdc-faf8dbd6aa2a.png)

Grey is what both images have in common (unchanged).
Red is what the blessed image has and is missing in the current image.
Black is what the current image has and didn't exist in the blessed image.

This way I discovered that I introduced a bug that prevented notes from being invisible/transparent (the duplicate rests), while some stems were invisible (see above).

The first test was updating Vexflow from 1.2.89 to 1.2.90, which didn't cause any visual changes, good to know!

So when you make changes to OSMD, you can automatically find out whether there were good or bad side effects.

The procedure is explained in visual_regression.sh: (Linux/Mac only for now)
```sh
# do OSMD visual regression tests: (Linux/Mac only for now)
# local OSMD webserver has to be running
npm start

# set your OSMD to the previous state to compare to, e.g. by checking out develop in a new folder
# this can take 1-2 minutes
# we may soon put these blessed images in a git submodule for you to simply pull instead of having to generate them
npm run generate:blessed

# then apply your changes and do this:
npm run generate:current

# run visual regression tests, creating images in visual_regression/diff.
# this can take 3+ minutes
npm run test:visual
```

Now you'll get reports for visual differences in the console, and there will be diff and current/blessed images in the data/images/diff folder if there were changes above threshold.

You can also run visual tests for single samples:
```sh
node test/Util/generateDiffImagesPuppeteerLocalhost.js ./test/data ./visual_regression/current mySampleFileNameRegExPrefix.*
sh ./test/Util/visual_regression.sh ./visual_regression mySampleFileNamePrefix
# note that javascript expects a RegEx, while the shell script accepts a prefix x that will match 'x*' in the shell. We may use RegExes for both later on.
```

ps: the code is adapted with few core changes from Vexflow, so thanks to them! The bigger challenge was generating pngs in the first place.

sschmidTU:
[...]

PR #1709, opened 2026-08-06
Visual regression tests: Speed up and make cross-platform (visualRegression.mjs, works on Windows/any platform)

sschmidTU:
Node + node-canvas reimplementation of test/Util/visual_regression.sh, which needed bash + ImageMagick and only ran on Linux/macOS (e.g. in a VM on Windows). It produces the same visual_regression/diff/ output: results.txt, warnings.txt, <name>.png (red-highlight diff), <name>_Blessed.png and <name>_Current.png.

- The diff images are pixel-identical to ImageMagick's `compare -highlight-color '#ff000050'` (verified maxChannelDelta 0): unchanged content dimmed 80% toward white, changed pixels = red #ff0000 composited at alpha 80/255 over the current pixel.
- results.txt reports the differing-pixel count per sample (0 = identical), sorted biggest-first, instead of ImageMagick's PHASH (impractical to reproduce in JS; a pixel count is strictly more sensitive).
- **Performance gain: Byte-identical images skip decoding, so typical runs finish in seconds instead of minutes.**
In a (slightly performance-capped) Linux VM, the old visual regression script took 2m31s real and 1m15s user on my machine. The new .mjs script took 1.2 seconds.

npm scripts: test:visual, test:visual:build and test:visual:singletest now use the .mjs by default (cross-platform).
The previous bash+ImageMagick versions are kept as test:visual:old, test:visual:build:old and test:visual:singletest:old.

---

The process remains the same (as described in the usage section of the new script):
checkout develop
npm run build; npm run generate:blessed
checkout your branch/changes
npm run test:visual:build
or
npm run build
npm run generate:current
npm run test:visual
(this does the same as test:visual:build)

---

The visual regression tests test and render all test samples, and can thus find any, often tiny, differences in rendering between two commits or states of code, usually develop and your current changes.

See the visual regression issue #676 for more details and example images.
image

### DIFF
--- a/package.json
+++ b/package.json
@@ -36,11 +36,14 @@
     "generate:current:debug": "node ./test/Util/generateImages_browserless.mjs ../../build ./test/data ./visual_regression/current png 0 0 allSmall --debugosmdtesting",
     "generate:current:singletest": "node test/Util/generateImages_browserless.mjs ../../build ./test/data ./visual_regression/current png 0 0 ^Beethoven --osmdtestingsingle",
     "generate:blessed": "node ./test/Util/generateImages_browserless.mjs ../../build ./test/data ./visual_regression/blessed png 0 0 allSmall --osmdtesting",
-    "test:visual": "bash ./test/Util/visual_regression.sh ./visual_regression",
+    "test:visual": "node ./test/Util/visualRegression.mjs ./visual_regression",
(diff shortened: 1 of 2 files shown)


sschmidTU:
Merging.
Already tested multiple times, though I'll run both the new and old visual regression tests for a while to make sure they give the same results.

PR #1715, opened 2026-08-06
perf(Formatter): Dedupe the duration key computation in evaluate()

sschmidTU:
Small optimization saving a precomputed value for later, performance gain likely below significance threshold.

Formatter.evaluate() (VexFlowPatch) computed each note's duration key, note.getTicks().clone().simplify().toString() - a Fraction clone + simplify + string allocation - twice per note: once in the mean-spacing loop and again in the deviation loop. The key is a pure function of the note, so compute it once in the first loop and stash it on the note's persistent formatter metrics (getFormatterMetrics() returns the object created once in the Tickable ctor), then reuse it in the second loop. The value written to metrics.duration is unchanged, so output is byte-identical.

### DIFF
--- a/src/VexFlowPatch/src/formatter.js
+++ b/src/VexFlowPatch/src/formatter.js
@@ -556,9 +556,14 @@ export class Formatter {
 
     this.voices.forEach(voice => {
       voice.getTickables().forEach((note, i, notes) => {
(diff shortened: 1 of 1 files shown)


sschmidTU:
Merging, visual regression tests clean.

PR #1722, opened 2026-08-13
fix: include octave-change in PlaybackTranspose

dotkebi:
## Problem

MusicXML defines the sounding pitch of a transposing instrument as:

```
sounding pitch = written pitch + <chromatic> + 12 * <octave-change>
```

`InstrumentReader` only reads `<chromatic>` into `Instrument.PlaybackTranspose`, so the `<octave-change>` part of a `<transpose>` element is silently dropped. Instruments that transpose by a full octave in addition to their chromatic interval end up an octave off for playback consumers.

Concrete cases:

| Instrument | `<chromatic>` | `<octave-change>` | Expected | Current |
|---|---|---|---|---|
| Bb clarinet | -2 | — | -2 | -2 ✅ |
| Bb tenor sax | -2 | -1 | **-14** | -2 ❌ (sounds an octave too high) |
| Piccolo | — | +1 | **+12** | 0 ❌ |

Bass clarinet is affected the same way as tenor sax. Parts that carry only `<octave-change>` with no `<chromatic>` currently yield `0`, because the value is read inside an `if (chromaticNode)` branch.

This affects consumers that use `PlaybackTranspose` to derive sounding pitch (MIDI export, playback engines); the rendered notation itself is unchanged.

## Fix

Read both elements and combine them:

```ts
this.instrument.PlaybackTranspose =
  (isNaN(chromatic) ? 0 : chromatic) + 12 * (isNaN(octaveChange) ? 0 : octaveChange);
```

Parts with only `<chromatic>` keep exactly their previous value, so existing scores are unaffected. The assignment now also runs when only `<octave-change>` is present.

## Tests

Adds `test/data/test_transpose_octave_change.musicxml` with three parts covering all cases, and three assertions in `MusicSheetReader_Test.ts`:

- chromatic only → `-2` (regression guard: previous behaviour preserved)
- chromatic + octave-change → `-14`
- octave-change only → `+12`

The fixture is picked up automatically by the existing `test/data/*.musicxml` karma pattern.

## Note on an unrelated test failure

When running the suite locally (macOS, Chrome 151, ChromeHeadless) one unrelated test fails:

```
widest slur's flattened arc (3.0) should be far below unflattened (4.6):
expected 3.0405566236394037 to be below 3.0012267258723133
```

This is **not caused by this PR**. It reproduces on a clean checkout of the base commit, and also on `08379287` — the commit that introduced that test — with a bit-identical value. It looks environment-dependent (the flattening does work: 4.6 → 3.04, a 34% reduction against an asserted 35%), so it presumably passes on CI. Mentioning it only so it is not mistaken for fallout from this change; happy to open a separate issue if it is useful.

### DIFF
--- a/src/MusicalScore/ScoreIO/InstrumentReader.ts
+++ b/src/MusicalScore/ScoreIO/InstrumentReader.ts
@@ -820,9 +820,15 @@ export class InstrumentReader {
     }
     const transposeNode: IXmlElement = attrNode.element("transpose");
     if (transposeNode) {
+      // MusicXML: sounding pitch = written pitch + chromatic + 12 * octave-change.
+      // E.g. Bb clarinet: chromatic -2. Bb tenor sax: chromatic -2, octave-change -1.
       const chromaticNode: IXmlElement = transposeNode.element("chromatic");
-      if (chromaticNode) {
-        this.instrument.PlaybackTranspose = parseInt(chromaticNode.value, 10);
+      const octaveChangeNode: IXmlElement = transposeNode.element("octave-change");
+      const chromatic: number = chromaticNode ? parseInt(chromaticNode.value, 10) : NaN;
+      const octaveChange: number = octaveChangeNode ? parseInt(octaveChangeNode.value, 10) : NaN;
+      if (!isNaN(chromatic) || !isNaN(octaveChange)) {
+        this.instrument.PlaybackTranspose =
+          (isNaN(chromatic) ? 0 : chromatic) + 12 * (isNaN(octaveChange) ? 0 : octaveChange);
       }
     }
     const clefList: IXmlElement[] = attrNode.elements("clef");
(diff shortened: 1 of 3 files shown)


sschmidTU (approved):
LGTM, merging.

Thanks for the PR! You're right, specific instruments were played one octave off, like Bb tenor sax, and also the guitar in some samples, which is significant.

I updated the table of pitches before and after with some added samples:

| Sample / part | Written | Played now | Correct sounding |
|---|---|---|---|
| PR fixture, Bb tenor sax | C4 (60) | Bb3 (58) | Bb2 (46) |
| PR fixture, piccolo | C5 (72) | C5 (72) | C6 (84) |
| Actor Prelude, Piccolo | B5 (83) | B5 (83) | B6 (95) |
| Actor Prelude, Bass Clarinet in Bb | G#3 (56) | F#3 (54) | F#2 (42) |
| test_tabs_simple, guitar | E3 (52) | E3 (52) | E2 (40) |

About the test failure: that only happens in Chromium based browsers, CI tests use Firefox so it passed. But I'll fix the tests in a separate commit with wider margins so it also passes in Chrome.

PR #1723, opened 2026-08-14
feat(Fingering): Link each fingering label to its note (GraphicalLabel.sourceNote)

isc:
Since #1685, `calculateFingerings()` stacks the fingerings of a staff entry in the pitch order of their notes, which is deliberately *not* the order they were read in. So the index of a label in `GraphicalStaffEntry.FingeringEntries` says nothing about which note it belongs to, and the `GraphicalLabel` carries no reference back to the note either.

That leaves a consumer who wants to act on one rendered fingering — edit its text in place, re-position it, hit-test it — with no way to tell whose fingering it is except to re-implement the stacking order outside OSMD and keep that copy in sync with any later change to the heuristics here.

I know that cost first-hand: my app did exactly that, and when #1685 landed the copy silently went out of step with OSMD for two months. Editing a fingering in a two-voice chord rewrote a different note's label, because the app's ordering still assumed the old voice order.

## The change

Set `GraphicalLabel.sourceNote` where the label is created, from the `fingering.sourceNote` that is already in scope:

```ts
gLabel.sourceNote = fingering.sourceNote;
gse.FingeringEntries.push(gLabel);
```

Purely additive. The field is `undefined` for every other kind of label, and `FingeringEntries` keeps both its type (`GraphicalLabel[]`) and its order, so nothing downstream changes.

## Alternatives I considered, in case you would rather have one of them

- **A `GraphicalFingeringEntry` wrapper**, pairing the label with its note the way `GraphicalLyricEntry` and `GraphicalChordSymbolContainer` do. That is the more idiomatic shape for this codebase, and I would happily write it — but it changes the type of the public `FingeringEntries` array, so it felt like a call for a major rather than something to slip into 2.x.
- **`Clickable.dataObject`**, which `GraphicalLabel` already inherits and which nothing in `src/` currently assigns. It would cost no new field at all. I went with a typed field instead because `dataObject: Object` gives a consumer no type safety, and because a fingering label arguably "represents" its `TechnicalInstruction` rather than the note — I did not want to spend that generic slot on a narrower meaning. Glad to switch if you would prefer the existing field populated.
- **Setting it for other label kinds too.** Lyrics can already reach their note through `GraphicalLyricEntry`, and chord symbols have no single note, so only fingerings actually gain anything here.

The name follows `GraphicalNote.sourceNote` and `TechnicalInstruction.sourceNote` rather than the PascalCase fields on `GraphicalLabel` itself (`SVGNode`, `ColorXML`), since `Notehead.SourceNote` suggests PascalCase marks an accessor here rather than a plain field. Happy to flip it.

## Test

Extends `VexFlowMeasure_Test.ts` using the sample already added for #1685, asserting the pairing note-by-note for all four stacks — `["E4=1", "G4=3", "C5=5"]` and so on — so a later change to the stacking order cannot silently break the link.

Full suite green locally (366 passing, Firefox headless), lint clean.

### DIFF
--- a/src/MusicalScore/Graphical/GraphicalLabel.ts
+++ b/src/MusicalScore/Graphical/GraphicalLabel.ts
@@ -24,8 +24,9 @@ export class GraphicalLabel extends Clickable {
     /** The note this label was created for, if any. Set for fingerings (GraphicalStaffEntry.FingeringEntries),
      *  which are stacked in the pitch order of their notes rather than the order they were read in,
      *  so that the note a fingering belongs to can be told without re-deriving that order.
+     *  Named like GraphicalNote.sourceNote, which holds the same kind of reference.
      */
-    public SourceNote: Note;
+    public sourceNote: Note;
 
     /**
      * Creates a new GraphicalLabel from a Label
--- a/src/MusicalScore/Graphical/GraphicalLabel.ts
+++ b/src/MusicalScore/Graphical/GraphicalLabel.ts
@@ -1,5 +1,6 @@
 import { TextAlignmentEnum } from "../../Common/Enums/TextAlignment";
 import { Label } from "../Label";
+import { Note } from "../VoiceData/Note";
 import { BoundingBox } from "./BoundingBox";
 import { Clickable } from "./Clickable";
 import { EngravingRules } from "./EngravingRules";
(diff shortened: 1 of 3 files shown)


sschmidTU (approved):
LGTM, merging.
No visual regressions.

Nice utility addition, thanks!
You're right, it was difficult or impossible to find which note a fingering GraphicalLabel belonged to, especially if there are multiple notes with the same fingering.

(circular references are not an issue outside of module imports if it's the graphical -> non-graphical direction: Module import cycles only bite when a module dereferences an import while modules are still being evaluated, such as class A extends B across the cycle or a top-level constant computed from the other module. Uses inside methods and constructors are lazy and fine, and type annotations don't exist at runtime.

Also, graphical -> non-graphical means we have a short-lived pointer (graphic can be recalculated) towards a long-lived data model (which only gets calculated once on load()). the other direction would be more problematic as it would mean pointers become stale)

PR #1724, opened 2026-08-18
Render all children in combined dynamics

### LINKED ISSUE #1705: Combined dynamics such as `<sf/><mp/>` render only the first child element
## Description

I'm exporting MusicXML from a Finale conversion tool and ran into an issue with dynamic markings that have no dedicated MusicXML element of their own, such as `sfmp` and `ffz`. MusicXML represents these as a single `<dynamics>` element containing several children. OSMD renders only the first child and discards the rest.

[ffz.xml](https://github.com/user-attachments/files/30639673/ffz.xml)

The attached `ffz.xml` has one part and two bars, with three such markings:

| Location | Children | Should read |
|---|---|---|
| bar 1, beat 1 | `<sf/><mp/>` | sfmp |
| bar 1, beat 3 | `<other-dynamics>s</other-dynamics><f/><other-dynamics>m</other-dynamics><p/>` | sfmp |
| bar 2, beat 1 | `<ff/><other-dynamics>z</other-dynamics>` | ffz |

Bar 1 deliberately spells the *same* marking two ways: once from the composite `sf` and `mp` symbols, and once from four separate letters. Both are the same dynamic and should render identically.

**Steps to reproduce**

1. Open the attached `ffz.xml` in OSMD (the demo page is sufficient).
2. Look at the three dynamic markings.

**Expected:** each marking renders in full, as `sfmp`, `sfmp`, `ffz`.

**Actual:**

* bar 1, beat 1 renders `sf`; the `<mp/>` is dropped.
* bar 1, beat 3 renders **nothing at all**.
* bar 2, beat 1 renders `ff`; the `z` is dropped.

The bar 1, beat 3 case is the one I find most concerning: the marking disappears silently, with nothing logged.

## What the spec says

For reference, the [MusicXML 4.0 spec](https://www.w3.org/2021/06/musicxml40/musicxml-reference/elements/dynamics/) states, "The `<other-dynamics>` element allows other dynamic marks that are not covered here. Dynamics elements may also be combined to create marks not covered by a single element, such as `<sf/><mp/>`." That is exactly the use case `ffz.xml` is exercising. The marking at bar 1, beat 1 is literally `<sf/><mp/>`.

The same page gives the content model for `<dynamics>` as "Zero or more of the following" (`maxOccurs="unbounded"` in the schema), so multiple children are not an edge case but the declared grammar.

The documentation also explains why the elements are named as they are: "To avoid inconsistencies between and amongst the letter abbreviations for dynamics (what is sf vs. sfz, standing alone or with a trailing dynamic that is not always piano), we use the actual letters as the names of these dynamic elements." These are symbol-level elements, so a marking drawn from several symbols is correctly written as several children of one `<dynamics>` element. One advantage of this is that it represents the exact glyph sequence of the dynamic.

## Code analysis

*The following code analysis was produced by Claude (Opus 5), reading the current `develop` branch. I have reviewed it, but the code reading is Claude's rather than mine.*

**1. The reader keeps only the first child.** `ExpressionReader.interpretInstantaneousDynamics` (`src/MusicalScore/ScoreIO/MusicSymbolModules/ExpressionReader.ts`):

```ts
let expressionText: string = dynamicsNode.elements()[0]?.name;
if (expressionText === "other-dynamics") {
    expressionText = dynamicsNode.elements()[0].value;
}
```

Exactly one `InstantaneousDynamicExpression` is created per `<dynamics>` node, built from `elements()[0]`. The remaining children are never examined.

**2. The marking is then reduced to an enum.** The `InstantaneousDynamicExpression` constructor (`src/MusicalScore/VoiceData/Expressions/InstantaneousDynamicExpression.ts`):

```ts
this.dynamicEnum = DynamicEnum[dynamicExpression.toLowerCase()];
```

The source text is not retained anywhere on the expression.

**3. The renderer regenerates the display string from that enum.** `VexFlowInstantaneousDynamicExpression` (`src/MusicalScore/Graphical/VexFlow/VexFlowInstantaneousDynamicExpression.ts`):

```ts
get Expression(): string { return DynamicEnum[this.mInstantaneousDynamicExpression.DynEnum]; }
```

`DynamicEnum` covers `pppppp` through `ffffff`, plus `mp`, `mf`, `sf`, `sff`, `sfp`, `sfpp`, `fp`, `rf`, `rfz`, `sfz`, `sffz`, `fz`, and `other`. It has no `s`, `m`, `z`, `r`, or `n`.

Together these account for all three observed results:

* bar 1, beat 1: `elements()[0]` is `<sf/>`, giving `"sf"`, which resolves to `DynamicEnum.sf` and renders as `sf`.
* bar 2, beat 1: `elements()[0]` is `<ff/>`, giving `"ff"`, which renders as `ff`.
* bar 1, beat 3: `elements()[0]` is `<other-dynamics>s</other-dynamics>`, giving `"s"`. `DynamicEnum["s"]` is `undefined`, so `Expression` returns `DynamicEnum[undefined]`, which is also `undefined`, and nothing is drawn.

**A caution for whoever fixes this:** changing only the reader to concatenate the children would make matters worse. `"sfmp"` is not in `DynamicEnum` either, so every combined marking would then render nothing rather than partially. A working fix needs the expression to retain its source text alongside `DynamicEnum` (which can remain for playback volume), with the renderer using that retained text.

Note also that the same `undefined` path affects the single-child form `<other-dynamics>ffz</other-dynamics>`, which is what Finale, Sibelius, and MuseScore all emit for these markings. The underlying problem is therefore not limited to combined dynamics.

Finally, the rendering engine is not the constraint here. VexFlow's `TextDynamics` is documented as rendering "any dynamics string that contains a combination of the following letters: P, M, F, Z, R, S", mapping each letter to a SMuFL glyph. OSMD currently draws dynamics as a bold-italic `GraphicalLabel` rather than through `TextDynamics`, so proper glyph rendering is available should it be wanted.

aribradshaw:
Fixes #1705.

## What changed

- Read every instantaneous-dynamic child from a MusicXML direction instead of stopping after the first.
- Preserve the child order for rendering compound dynamics such as `sf mp`.
- Keep the first dynamic as the playback enum so existing playback behavior remains stable.
- Add a focused MusicXML fixture and regression tests for rendering order and playback selection.

## Why

The expression reader previously retained only the first child element in a compound dynamics element. The remaining markings were discarded before graphical rendering.

## Validation

- `npm run lint`
- The two new issue #1705 regression tests pass.
- Headless browser suite: 366 passed and 2 skipped. One unrelated, browser-sensitive slur geometry assertion failed under Microsoft Edge.

### DIFF
--- a/src/MusicalScore/Graphical/VexFlow/VexFlowInstantaneousDynamicExpression.ts
+++ b/src/MusicalScore/Graphical/VexFlow/VexFlowInstantaneousDynamicExpression.ts
@@ -1,5 +1,5 @@
 import { GraphicalInstantaneousDynamicExpression } from "../GraphicalInstantaneousDynamicExpression";
-import { InstantaneousDynamicExpression, DynamicEnum } from "../../VoiceData/Expressions/InstantaneousDynamicExpression";
+import { InstantaneousDynamicExpression } from "../../VoiceData/Expressions/InstantaneousDynamicExpression";
 import { GraphicalLabel } from "../GraphicalLabel";
 import { Label } from "../../Label";
 import { TextAlignmentEnum } from "../../../Common/Enums/TextAlignment";
@@ -28,6 +28,6 @@ export class VexFlowInstantaneousDynamicExpression extends GraphicalInstantaneou
     }
 
     get Expression(): string {
-        return DynamicEnum[this.mInstantaneousDynamicExpression.DynEnum];
+        return this.mInstantaneousDynamicExpression.DynamicExpression;
     }
 }
--- a/src/MusicalScore/ScoreIO/MusicSymbolModules/ExpressionReader.ts
+++ b/src/MusicalScore/ScoreIO/MusicSymbolModules/ExpressionReader.ts
@@ -639,24 +639,22 @@ export class ExpressionReader {
                 this.directionTimestamp = Fraction.createFromFraction(inSourceMeasureCurrentFraction);
             }
             const numberXml: number = this.readNumber(dynamicsNode); // probably never given, just to comply with createExpressionIfNeeded()
-            let expressionText: string = dynamicsNode.elements()[0]?.name; // elements can in rare cases still be empty even though hasElements=true, see #1269
-            if (expressionText === "other-dynamics") {
-                expressionText = dynamicsNode.elements()[0].value;
-            }
+            const dynamicsElements: IXmlElement[] = dynamicsNode.elements();
+            // A single <dynamics> element may contain multiple symbols, e.g. <sf/><mp/>.
+            const expressionText: string = dynamicsElements
+                .map((element: IXmlElement): string => element.name === "other-dynamics" ? element.value : element.name)
+                .join("");
             if (expressionText) {
+                // Keep the previous first-child enum behavior for playback while rendering the complete marking.
+                const firstDynamicText: string = dynamicsElements[0].name === "other-dynamics"
+                    ? dynamicsElements[0].value
+                    : dynamicsElements[0].name;
+                const dynamicEnum: DynamicEnum = DynamicEnum[firstDynamicText?.toLowerCase()];
                 // ToDo: make duplicate recognition an afterReadingModule, as we can't definitively check here if there is a repetition:
                 // Compare with the active dynamic expression and only add it if there is a change in dynamic
                 // Exception is when a repetition starts here, where the "repeated" dynamic might be desired.
                 // see PR #767 where this was removed
                 if (currentMeasure.Rules?.IgnoreRepeatedDynamics) {
-                    let dynamicEnum: DynamicEnum;
-                    try {
-                        dynamicEnum = DynamicEnum[expressionText];
-                    } catch (err) {
-                        const errorMsg: string = ITextTranslation.translateText("ReaderErrorMessages/DynamicError", "Error while reading dynamic.");
-                        this.musicSheet.SheetErrors.pushMeasureError(errorMsg);
-                        return;
-                    }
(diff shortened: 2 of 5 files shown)


aribradshaw:
Hi, this PR’s CI workflow is currently marked `action_required` and appears to be waiting for maintainer approval. When convenient, could someone approve it? Thanks!

sschmidTU (approved):
LGTM, merging.

There are a few dynamics now newly rendered in our test samples:
Schubert An die Musik m14 cresc:
image

Norma Fantasy m1 marcato (`test_clef_change_on_invisible_note_norma_fantasy_1605.musicxml`):
image


There are a few small nitpicks but I can do them as follow-up commits:
* IgnoreRepeatedDynamics dedupe should now compare DynamicExpression, otherwise `<sf/><p/>` following `<sf/><mp/>` would be dropped as a repeat.
* In the else branch, activeInstantaneousDynamic is still built from the joined text, so its enum is undefined for "sfmp" while the real expression carries sf. Pass dynamicEnum there too.
* The second dynamic mark in the test xml is written in such a (incorrect) way that osmd doesn't assign a volume (because "s" is not a dynamic mark): `<other-dynamics>s</other-dynamics><f/><other-dynamics>m</other-dynamics><p/>` - but that's fine. Could be asserted in the test, on the other hand we maybe don't want to prescribe this as correct, to also detect this as dynamics in the future.
* Follow-up: MusicXML 4.0's `<n/>, <pf/>, <sfzp/>` are missing from DynamicEnum, so they now render as text but have no volume. And the Braille plugin renders DynEnum only, so "sfmp" comes out as "sf".

The Edge test failure is a threshold that only worked under Firefox (used in CI), fixed today in 93a109c9.

PR #1725, opened 2026-08-19
fix(stems): don't force a lone secondary voice stem-down (#1719)

### LINKED ISSUE #1719: Secondary voice is forced stem-down when no other voice is active
# Secondary voice is forced stem-down when no other voice is active

## Description

OSMD forces every note in a secondary voice to use downward stems, even when that voice is the only voice active at those timestamps and the MusicXML does not specify a stem direction.

The attached `layer1layer4.xml` is a minimal reproduction. It contains the same pitches and rhythms in two consecutive measures:

- Measure 1 uses voice 1.
- Measure 2 uses voice 7.
- The voices never overlap.
- Neither measure contains a `<stem>` element.

Voice 7 is intentional and valid. MusicXML defines `<voice>` as a string identifier and gives its numeric value no ordering or stem-direction semantics.

## Steps to reproduce

1. Open the attached `layer1layer4.xml` in OSMD.
2. Compare the stem directions in measures 1 and 2.

## Actual result

The notes in measure 1 receive normal automatic stem directions. The identical notes in measure 2 are all forced stem-down solely because voice 7 was encountered after voice 1. Here is the result:

Image

## Expected result

Measure 2 should use the same pitch-based automatic stem calculation as measure 1. A secondary voice should be forced down only while another voice is active at the same staff position, unless MusicXML explicitly supplies `<stem>down</stem>`.

## Relevant MusicXML

The file declares that stem directions are not comprehensively encoded:

```xml
<supports type="no" element="stem" />
```

The first measure contains:

```xml
<note>
  <pitch>
    <step>F</step>
    <octave>4</octave>
  </pitch>
  <duration>16</duration>
  <voice>1</voice>
  <type>half</type>
</note>
```

The second measure contains the identical note in a different voice:

```xml
<note>
  <pitch>
    <step>F</step>
    <octave>4</octave>
  </pitch>
  <duration>16</duration>
  <voice>7</voice>
  <type>half</type>
</note>
```

## Code analysis

*Code analysis performed with OpenAI Codex (GPT-5.6-sol) on August 8, 2026, against OSMD develop commit bce4d08f0ec03658ae4ecf4ba34e49b7c9f9348d.*

[layer1layer4.xml](https://github.com/user-attachments/files/30866857/layer1layer4.xml)

In `MusicSheetCalculator.calculateStemDirectionFromVoices()`, a main voice is forced up only when multiple voice entries occur at the same staff entry. A `LinkedVoice`, however, is forced down unconditionally:

That asymmetry appears to cause the result. The secondary voice remains a `LinkedVoice` throughout the instrument, so it receives `WantedStemDirection = Down` even at timestamps where it is the only active voice.

One possible fix would be to force a linked voice down only when its `ParentSourceStaffEntry` contains multiple voice entries. Otherwise, leave `WantedStemDirection` undefined so the normal engraving logic can determine the stem direction. Beamed groups may need the existing propagation logic to carry the selected direction through the beam.

## Attachment

[`layer1layer4.xml`](https://github.com/opensheetmusicdisplay/opensheetmusicdisplay/blob/bce4d08f0ec03658ae4ecf4ba34e49b7c9f9348d/src/MusicalScore/Graphical/MusicSheetCalculator.ts#L4069-L4097)

youdie006:
Fixes #1719

## Repro
A secondary (linked) voice that is the **only** voice at its staff entry, in a file with no `<stem>` elements, is forced stem-**down** -- while an identical note in the main voice auto-stems normally.

`test/data/test_secondary_voice_stem_direction.musicxml` is a minimal reproduction: two consecutive measures with the identical F4 half note (F4 is below the treble middle line, so its automatic stem points **up**), no `<stem>` anywhere:
- Measure 1 uses voice 1 (main voice) -> auto-stems up.
- Measure 2 uses voice 7 (a distinct voice number, parsed as a `LinkedVoice`) -> forced down.

The voices never overlap, so each note is alone at its staff entry.

## Root cause
`MusicSheetCalculator.calculateStemDirectionFromVoices()` treats the two voice kinds asymmetrically. The main-voice branch only forces a direction **when another voice is present** at the same staff entry (`ParentSourceStaffEntry.VoiceEntries.length > 1`), but the `LinkedVoice` branch forced `StemDirectionType.Down` **unconditionally**:

```ts
if (voiceEntry.ParentVoice instanceof LinkedVoice) {
    voiceEntry.WantedStemDirection = StemDirectionType.Down;   // no clash guard
} else {
    if (voiceEntry.ParentSourceStaffEntry.VoiceEntries.length > 1) {
        voiceEntry.WantedStemDirection = StemDirectionType.Up;  // guarded
    }
}
```

A linked voice stays a `LinkedVoice` for the whole instrument, so it got `Down` even at timestamps where it is the sole active voice.

## Fix
Give the `LinkedVoice` branch the same `VoiceEntries.length > 1` clash guard. When the linked voice is alone at its staff entry, `WantedStemDirection` is left `Undefined`, so the normal pitch-based auto-stemming applies -- the same as the main voice. The main-voice branch is unchanged, and the caller only reaches this code when the MusicXML supplies no `<stem>`, so explicit stem directions are never overridden.

Per MusicXML, `<voice>` is a string identifier with no ordering or stem semantics, and this file declares `<supports type="no" element="stem"/>`, so automatic stemming is the intended default.

## Test
`VexFlowMeasure_SecondaryVoiceStemDirection_Test.ts` loads the fixture, calculates, and asserts the lone measure-2 linked-voice entry keeps `StemDirectionType.Undefined` and matches the identical main-voice measure. Red before the fix (`WantedStemDirection === Down`), green after; the sibling unison/stem tests still pass.

Reported by @rpatters1 (with a clear root-cause analysis and minimal `layer1layer4.xml`) -- thank you.

*AI-assisted: this fix and test were prepared with AI assistance and reviewed by me.*

### DIFF
--- a/src/MusicalScore/Graphical/MusicSheetCalculator.ts
+++ b/src/MusicalScore/Graphical/MusicSheetCalculator.ts
@@ -4082,8 +4082,13 @@ export abstract class MusicSheetCalculator {
             }
         } else {
             if (voiceEntry.ParentVoice instanceof LinkedVoice) {
-                // Linked voice: set stem down:
-                voiceEntry.WantedStemDirection = StemDirectionType.Down;
+                // Linked voice: set stem down, but only while another voice is present at
+                // the same staff entry (same clash guard as the main-voice branch below).
+                // A linked voice alone at its staff entry has no voice to avoid, so leave
+                // WantedStemDirection undefined and let pitch-based auto-stemming apply.
+                if (voiceEntry.ParentSourceStaffEntry.VoiceEntries.length > 1) {
+                    voiceEntry.WantedStemDirection = StemDirectionType.Down;
+                }
             } else {
                 // if this voiceEntry belongs to the mainVoice:
                 // check first that there are also more voices present:
--- a/test/MusicalScore/Graphical/VexFlow/VexFlowMeasure_SecondaryVoiceStemDirection_Test.ts
+++ b/test/MusicalScore/Graphical/VexFlow/VexFlowMeasure_SecondaryVoiceStemDirection_Test.ts
@@ -1,0 +1,63 @@
+/* eslint-disable @typescript-eslint/no-unused-expressions */
+import {GraphicalMusicSheet} from "../../../../src/MusicalScore/Graphical/GraphicalMusicSheet";
+import {IXmlElement} from "../../../../src/Common/FileIO/Xml";
+import {MusicSheet} from "../../../../src/MusicalScore/MusicSheet";
+import {MusicSheetReader} from "../../../../src/MusicalScore/ScoreIO/MusicSheetReader";
+import {VexFlowMusicSheetCalculator} from "../../../../src/MusicalScore/Graphical/VexFlow/VexFlowMusicSheetCalculator";
+import {TestUtils} from "../../../Util/TestUtils";
+import {GraphicalStaffEntry} from "../../../../src/MusicalScore/Graphical/GraphicalStaffEntry";
+import {GraphicalVoiceEntry} from "../../../../src/MusicalScore/Graphical/GraphicalVoiceEntry";
+import {StemDirectionType} from "../../../../src/MusicalScore/VoiceData/VoiceEntry";
+import {expect} from "chai";
+
+describe("VexFlow Measure - Secondary Voice Stem Direction (issue #1719)", () => {
+
+   const path: string = "test_secondary_voice_stem_direction.musicxml";
+
+   // Measures 1 and 2 carry the identical F4 half note (no <stem>). Measure 1 uses
+   // voice 1 (main voice); measure 2 uses voice 7, which OSMD parses as a LinkedVoice.
+   // Each measure has one voice entry per staff entry, so neither note collides with
+   // another voice and both should auto-stem (pitch-based), leaving WantedStemDirection
+   // Undefined. The bug forced the lone linked voice in measure 2 stem-down.
+   function graphicalMusicSheet(): GraphicalMusicSheet {
+      const score: Document = TestUtils.getScore(path);
+      expect(score).to.not.be.undefined;
+      const partwise: Element = TestUtils.getPartWiseElement(score);
+      expect(partwise).to.not.be.undefined;
+      const reader: MusicSheetReader = new MusicSheetReader();
+      const calc: VexFlowMusicSheetCalculator = new VexFlowMusicSheetCalculator(reader.rules);
+      const sheet: MusicSheet = reader.createMusicSheet(new IXmlElement(partwise), path);
+      const gms: GraphicalMusicSheet = new GraphicalMusicSheet(sheet, calc);
+      calc.calculate();
+      return gms;
+   }
+
+   // The single voice entry of measure `measureIndex` (staff 0), which holds the lone note.
+   function loneVoiceEntry(gms: GraphicalMusicSheet, measureIndex: number): GraphicalVoiceEntry {
+      const staffEntries: GraphicalStaffEntry[] = gms.MeasureList[measureIndex][0].staffEntries;
+      expect(staffEntries.length, `measure ${measureIndex + 1} should have one staff entry`).to.equal(1);
+      const gves: GraphicalVoiceEntry[] = staffEntries[0].graphicalVoiceEntries;
+      expect(gves.length, `measure ${measureIndex + 1} staff entry should have one voice entry`).to.equal(1);
+      return gves[0];
+   }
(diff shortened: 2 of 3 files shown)


sschmidTU:
This is a valid use case, but I would gate it behind an optional EngravingRule that's disabled by default.
It's standard musical practice to use a different stem direction for secondary voices.
MuseScore renders your sample the same way OSMD does before the PR, with the second note downwards:
image

It's a valid option to not force stems of these notes downwards when they don't conflict with other voices, but shouldn't be the default imo.

Your non-optional version would also change stems in an undesirable way in Scott Joplin Entertainer imo:
Before:
image

After:
image

This would be a breaking change, and negative imo.

That's one more reason I would gate your preference by an EngravingRule.
I've created a patch for your branch that you can apply:
[pr1725rulegated.patch](https://github.com/user-attachments/files/31830010/pr1725rulegated.patch)

put it in the main folder and apply it like this:
```sh
git apply --check pr1725-rule-gated.patch && git apply pr1725-rule-gated.patch
```

It would be great if you can apply, commit and push this patch to your pr branch, then I think this looks good for merging. @youdie006

rpatters1:
I submitted the issue, and I am confused as to why this fix would break the Joplin example. The only situation where it should apply is if there are no other voices in the measure on the staff. Clearly there are other voices in the Joplin example. So either the PR has not fully captured the conditions of the issue or this is not an actual breaking change.

The choice of "primary" vs "secondary" is an invention of OSMD. I'm not saying it is wrong. I'm saying there is nothing in the musicxml spec that specifies whether a voice is primary or secondary. If there is only one voice in a measure on a staff, there is no reason it should not get normal stem treatment: there is no musicxml basis upon which to do otherwise. (Though I grant the spec allows OSMD to do anything it chooses for unspecified stem directions.)

I was surprised to see MuseScore forces the stems down too. (I thought I had tried it before submitting the issue.) But at least MuseScore pretends to justify this by fabricating a full measure rest in the "primary" voice that is not in the xml. FWIW, Dorico imports the stems in the second bar in the same direction as the first, without fabricating a full measure rest.

If this lives behind an option, sobeit. I am mainly commenting because the Joplin example should not be affected by this PR. If it is, there is more wrong here than just a missing option.

sschmidTU:
edit: I realize you're not the submitter of the PR
Have you taken a look at the patch?
You can render without the patch, it affects the Joplin, you don't have to speculate whether it's affected or not, I also shared screenshots and measure number.
The patch removes the regression on Joplin and is not a lot of code, so it should be pretty easy to analyze.

This is the part of the patch that fixes this:
```diff
-                // Linked voice: set stem down, but only while another voice is present at
-                // the same staff entry (same clash guard as the main-voice branch below).
-                // A linked voice alone at its staff entry has no voice to avoid, so leave
-                // WantedStemDirection undefined and let pitch-based auto-stemming apply.
-                if (voiceEntry.ParentSourceStaffEntry.VoiceEntries.length > 1) {
+                // Linked (secondary) voice: stem down. With AutoStemSecondaryVoicesWhenAloneInMeasure, a secondary
+                //   voice that is the only voice with entries in this measure on this staff has no voice to avoid and
+                //   keeps WantedStemDirection undefined, i.e. gets pitch-based stems (#1719). Checking only the same
+                //   staff entry would not do: under a sustained main-voice note the secondary voice's notes would
+                //   flip to pitch-based stems mid-measure (e.g. The Entertainer m.56).
+                if (!this.rules.AutoStemSecondaryVoicesWhenAloneInMeasure || this.otherVoicePresentInMeasure(voiceEntry)) {
                     voiceEntry.WantedStemDirection = StemDirectionType.Down;
```

I mean, that's basically all the patch does.

youdie006:
Applied, committed and pushed - thanks for writing the patch.

@rpatters1 you were right that the PR hadn't captured the condition. My guard checked only the same
staff entry, so under a sustained main-voice note the secondary voice's later entries saw no other
voice at *their* staff entry and flipped to pitch-based stems mid-measure. That's the Entertainer
m.56 regression. sschmidTU's `otherVoicePresentInMeasure` checks the whole measure, which is what
the issue actually describes.

The rule is `AutoStemSecondaryVoicesWhenAloneInMeasure`, default false, so nothing changes for
existing users.

`eslint .` and `tsc --noEmit` are clean on the branch. I didn't render the Joplin file myself - I
have no browser here - so the visual confirmation is still yours.

sschmidTU (approved):
LGTM now, merging.
Thanks for applying the patch!
No visual regressions

PR #1730, opened 2026-09-07
fix(Unison): Carry the hidden unison exception to the notehead and the tuplet

### LINKED ISSUE #1729: Hidden unison notes: the beam exception does not reach the notehead or the tuplet number
## Summary

MuseScore writes a unison across two voices by hiding one of the two notes with `print-object="no"`, so that a single notehead serves both. OSMD already has a predicate for exactly this case, `Note.sharesNoteheadWithVisibleUnisonNote()`, and uses it to spare such a note's beam and stem — without it, a beamed hidden note loses its stem and the beam hangs in the air.

That exception is applied in two places and missing in two others, so the note ends up half-included in what it belongs to:

| | applies the unison exception |
|---|---|
| beam — `MusicSheetCalculator.ts:2269` | ✅ |
| stem colour — `VexFlowVoiceEntry.ts:189` | ✅ |
| **notehead colour** — `VexFlowVoiceEntry.ts:90` | ❌ bare `PrintObject` |
| **tuplet membership** — `MusicSheetCalculator.ts:2274` | ❌ bare `PrintObject` |

The two gaps give two visible symptoms, both below. (Line numbers on `develop` @ 000cd4e5.)

## Repro

One bar, minimal: a sustained F#3 minim in voice 1, and a triplet in voice 2 whose first quaver is the same F#3, hidden. This is Debussy's Arabesque no. 1, left hand, bar 3; the same encoding appears in Clair de lune and in Liszt's Liebestraum no. 3, all MuseScore exports.

<details>
<summary>MusicXML for the bar above</summary>

```xml
<?xml version="1.0" encoding="UTF-8" standalone="no"?>
<!DOCTYPE score-partwise PUBLIC "-//Recordare//DTD MusicXML 3.0 Partwise//EN" "http://www.musicxml.org/dtds/partwise.dtd">
<score-partwise version="3.0">
  <part-list><score-part id="P1"><part-name/></score-part></part-list>
  <part id="P1">
    <measure number="1">
      <attributes>
        <divisions>6</divisions>
        <key><fifths>4</fifths><mode>major</mode></key>
        <time><beats>2</beats><beat-type>4</beat-type></time>
        <clef><sign>F</sign><line>4</line></clef>
      </attributes>
      <!-- voice 1: the sustained bass, the head both voices share -->
      <note>
        <pitch><step>F</step><alter>1</alter><octave>3</octave></pitch>
        <duration>12</duration><voice>1</voice><type>half</type><stem>down</stem>
      </note>
      <backup><duration>12</duration></backup>
      <!-- voice 2: the triplet, its first note hidden behind the minim's head -->
      <note print-object="no">
        <pitch><step>F</step><alter>1</alter><octave>3</octave></pitch>
        <duration>4</duration><voice>2</voice><type>eighth</type>
        <time-modification><actual-notes>3</actual-notes><normal-notes>2</normal-notes></time-modification>
        <stem>up</stem><beam number="1">begin</beam>
        <notations><tuplet type="start" bracket="no"/></notations>
      </note>
      <note>
        <pitch><step>A</step><octave>3</octave></pitch>
        <duration>4</duration><voice>2</voice><type>eighth</type>
        <time-modification><actual-notes>3</actual-notes><normal-notes>2</normal-notes></time-modification>
        <stem>up</stem><beam number="1">continue</beam>
      </note>
      <note>
        <pitch><step>C</step><alter>1</alter><octave>4</octave></pitch>
        <duration>4</duration><voice>2</voice><type>eighth</type>
        <time-modification><actual-notes>3</actual-notes><normal-notes>2</normal-notes></time-modification>
        <stem>up</stem><beam number="1">end</beam>
        <notations><tuplet type="stop"/></notations>
      </note>
    </measure>
  </part>
</score-partwise>
```

</details>
## Symptom 1 — the notehead is not drawn

Where VexFlow cannot merge the two heads (a quaver head under a minim head), the hidden head is laid out beside the visible one and painted fully transparent. The stem and the beam are drawn, so the beam ends on a bare stem with nothing under it — the note is drawn half-visible: stem yes, beam yes, head no.

OSMD today: the triplet's beam rises from a bare stem, no notehead beside the minim's open head

The head is not missing from the DOM, only from the ink — `<g class="vf-notehead"><path fill="#00000000" …>`, laid out at its own x. Here is the same render with that single `fill` changed to black, nothing else, which is how MuseScore shows the bar in its own editor:

The same bar with the hidden head drawn: a filled head beside the open one, the stem between them

`VexFlowVoiceEntry.color()` sets `noteheadColor = transparentColor` on `!note.sourceNote.PrintObject` alone (line 90), while the stem a hundred lines below (line 189) gets the unison exception.

## Symptom 2 — the tuplet number ignores the hidden note

`MusicSheetCalculator.ts:2274` gates tuplet membership on `note.PrintObject` alone, immediately after line 2269 gates the beam on the widened condition:

```ts
if (note.NoteBeam !== undefined && (note.PrintObject || note.sharesNoteheadWithVisibleUnisonNote())) {
    if (!(note instanceof TabNote) || this.rules.TabBeamsRendered) {
        this.handleBeam(graphicalNote, note.NoteBeam, openBeams);
    }
}
if (note.NoteTuplets.length > 0 && note.PrintObject) {
```

So the hidden note never reaches `handleTuplet()`, the `VF.Tuplet` is built from two stave notes instead of three, and the number is centred over those two.

Below: the fixture (head inked, so all three heads are visible), then the control — the same bar with nothing hidden.

Hidden first note: the 3 sits over the second and third notes, right of the beam's centre

Nothing hidden: the 3 sits over the middle note, at the beam's centre

Everything else is identical between the two renders — notehead x, stems at 182.1 / 229.8 / 277.5, beam 181.3 → 278.2. Only the number moves, its centre going from 229.8 to 253.7: 23.9px, exactly half the note spacing, and exactly the midpoint of notes 2 and 3.

## Suggested rule

Carry the existing exception to both remaining sites: when a hidden note's beam and stem are drawn because it shares a head with a visible unison, it should also get its notehead and its place in the tuplet — a note drawn whole or not at all.

The tuplet side looks like a one-line change at 2274, matching 2269.

The notehead side needs one caveat: it must only apply where the head has a place of its own. Where VexFlow does put the two heads at the same x, inking the hidden one would overprint the visible head — a filled quaver head hiding an open minim head, which is worse than the missing head. Keying on the head's displacement relative to the visible unison head (rather than on the unison alone) covers both.

If you would rather honour `print-object="no"` literally for the notehead — Verovio renders that the way OSMD does today — that is a perfectly good answer for symptom 1, and we will keep the workaround on our side (after each render, we move the transparent path into the visible note's notehead group and give it the neighbour's fill). Symptom 2 seems harder to defend either way: hidden or not, the note is one of the tuplet's three.

**Happy to open the PR** — say which of the two you would take, and I will send it with test cases built on the bar above.

isc:
Fixes #1729.

`Note.sharesNoteheadWithVisibleUnisonNote()` already spares the beam and the stem of a note MuseScore hid with `print-object="no"` so that a visible unison note in another voice carries the shared notehead. Two other sites still gated on `PrintObject` alone, so the note ended up half-included in what it belongs to. This carries the exception to both.

Both symptoms on the test sample added here (Debussy's Arabesque no. 1 m.3) — `develop`, then this PR:

OSMD develop: the triplet's beam rises from a bare stem, no notehead beside the half note's open head, and the 3 sits over the second and third notes

With the fix: a filled head beside the open one, and the 3 centred over all three notes

### The notehead — `VexFlowVoiceEntry.color()`

Where Vexflow merges the two heads into one column (`StaveNote.format` → `mergeableUnison`), the visible note's head stands in for the hidden one and nothing is missing — that is the Moonlight Sonata case the predicate was written for. Where it can't, the hidden head is laid out beside the visible one and was painted fully transparent, so the beam ended on a bare stem with nothing under it.

A merge only ever happens between heads of the same shape (`mergeableUnison` refuses a half or whole note against anything else, and mismatched dots), which is what makes an eighth hidden under a half note stagger. So the head is now colored under exactly the condition its stem already uses — head and stem are drawn together, or neither is — and in the merged case the inked head lands on the visible one: same glyph, same position, no visual change. I checked that on `test_unison_notehead_moonlight_sonata_measure37`: the SVG comes out byte-identical apart from that one `fill`, the two head paths being identical strings at identical coordinates.

That makes the caveat I raised in the issue unnecessary — no displacement test is needed, because the only case where the head is not displaced is the case where inking it changes nothing.

`notehead="none"` is a deliberate request for no notehead and still stays hidden.

### The tuplet — `MusicSheetCalculator.handleVoiceEntry()`

Hidden or not, the note is one of the tuplet's notes. Left out of `handleTuplet()`, the `VF.Tuplet` was built from the remaining ones and its number centered over those — half a note spacing right of the beam's center in the test sample. One line, matching the beam gate two lines above it.

### Test sample

`test/data/test_unison_notehead_tuplet_arabesque_measure3.musicxml` is the minimal bar from the issue: a sustained F#3 half note in voice 1 and a triplet in voice 2 whose first eighth is the same F#3, hidden. It has both symptoms at once. The same encoding appears in Clair de lune and in Liszt's Liebestraum no. 3.

The new test in `VexFlowMeasure_Test.ts` asserts the hidden note's notehead is not transparent and that the `VF.Tuplet` contains all three notes; it fails on `develop` on the first assertion, and the tuplet one is visible in the render above as the `3` moving back to the beam's center.

Suite is green locally (403 tests).

### DIFF
--- a/src/MusicalScore/Graphical/MusicSheetCalculator.ts
+++ b/src/MusicalScore/Graphical/MusicSheetCalculator.ts
@@ -2271,7 +2271,11 @@ export abstract class MusicSheetCalculator {
                         this.handleBeam(graphicalNote, note.NoteBeam, openBeams);
                     }
                 }
-                if (note.NoteTuplets.length > 0 && note.PrintObject) {
+                // Same for its tuplet: hidden or not, the note is one of the tuplet's notes, so the tuplet number
+                // (and bracket) has to span it. Leaving it out built the VF.Tuplet from the remaining notes and
+                // centered the number over those, off the beam's center.
+                // E.g. Debussy Arabesque no. 1 m.3 (test_unison_notehead_tuplet_arabesque_measure3).
+                if (note.NoteTuplets.length > 0 && (note.PrintObject || note.sharesNoteheadWithVisibleUnisonNote())) {
                     // a note can be part of more than one tuplet (nested tuplets); add it to each of them
                     for (const noteTuplet of note.NoteTuplets) {
                         this.handleTuplet(graphicalNote, noteTuplet, openTuplets);
--- a/src/MusicalScore/Graphical/VexFlow/VexFlowVoiceEntry.ts
+++ b/src/MusicalScore/Graphical/VexFlow/VexFlowVoiceEntry.ts
@@ -59,6 +59,18 @@ export class VexFlowVoiceEntry extends GraphicalVoiceEntry {
         }
     }
 
+    /** Whether the note is drawn although its own notehead is hidden (print-object="no"), because it shares the
+     * notehead of a visible unison note in another voice and its stem is beamed. The stem emanates from the shared
+     * notehead and has to reach the beam, and wherever Vexflow can't merge the two heads into one column it lays
+     * the hidden note's notehead out beside the visible one, where it has to be drawn too - otherwise the beam ends
+     * on a bare stem with nothing under it. E.g. Beethoven Moonlight Sonata 1st mvt. m.37, where the two heads are
+     * merged (test_unison_notehead_moonlight_sonata_measure37), and Debussy Arabesque no. 1 m.3, where the hidden
+     * eighth note's head can't merge with a half note's and gets a column of its own
+     * (test_unison_notehead_tuplet_arabesque_measure3). */
+    private static drawnAsSharedUnisonNote(note: Note): boolean {
+        return note.NoteBeam !== undefined && note.sharesNoteheadWithVisibleUnisonNote();
+    }
+
     /** (Re-)color notes and stems by setting their Vexflow styles.
      * Could be made redundant by a Vexflow PR, but Vexflow needs more solid and permanent color methods/variables for that
      * See VexFlowConverter.StaveNote()
@@ -75,6 +87,11 @@ export class VexFlowVoiceEntry extends GraphicalVoiceEntry {
         for (let i: number = 0; i < this.notes.length; i++) {
             const note: GraphicalNote = this.notes[i];
 
+            // notehead="none" asks for no notehead at all, so it always stays hidden. print-object="no" hides it
+            // too, unless the note is drawn anyway because it shares a visible unison note's notehead: then it is
+            // drawn whole, notehead included, exactly like its stem below.
+            const noteheadVisible: boolean = note.sourceNote.Notehead?.Shape !== NoteHeadShape.NONE &&
+                (note.sourceNote.PrintObject || VexFlowVoiceEntry.drawnAsSharedUnisonNote(note.sourceNote));
             sourceNoteNoteheadColor = note.sourceNote.NoteheadColor;
             noteheadColor = sourceNoteNoteheadColor;
             // Switch between XML colors and automatic coloring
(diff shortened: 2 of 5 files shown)


sschmidTU (approved):
LGTM, merging.
No visual regressions.

There's one small edge case I can fix in a follow-up commit:
When 2 noteheads are merged, if one of them was a different color (e.g. red), merging noteheads would make the notehead black.

PR #1731, opened 2026-09-18
fix(Ties): Draw a tie that starts or ends at a notehead shared with a hidden unison note of another voice

isc:
## Summary

MuseScore writes a unison across two voices by hiding one of the two notes with `print-object="no"`, so that a single notehead serves both (the encoding #1730 dealt with). When the **visible** note of such a pair starts or ends a tie, the tie can go missing.

`GraphicalStaffEntry.findTieGraphicalNoteFromNote()` looks up the tie note by pitch and timestamp only, and returns the first match in the staff entry. If the other voice's hidden note comes first, the lookup returns that note. `handleTie()` then skips the segment, because it never draws a tie to a hidden note (`if (!startNote.sourceNote.PrintObject || !endNote.sourceNote.PrintObject) continue;`).

Example: Bach, Prelude no. 2 in C minor BWV 847, m.35.
- Voice 2, the left hand, holds a C2: a dotted eighth tied to a quarter tied to a half.
- Voice 1 moves above it, starting from the same C2, which is hidden.
- The held C2's first tie was missing.
- Its second tie was drawn, because by then voice 1 has moved on to B2.

## Solution

`findTieGraphicalNoteFromNote()` first returns the tie note's own `GraphicalNote` when the staff entry has it, found with the existing `findGraphicalNoteFromNote()`. If the staff entry doesn't, it falls back to the existing pitch match, unchanged.

## Screenshots

New test sample `test_tie_hidden_unison_bwv847_measure35.musicxml` (BWV 847 m.35, left hand, as MuseScore exports it).

Before: the held C2 is missing its first tie.
before

After:
after

## Tests

- New unit test in `VexFlowMeasure_Test.ts`. It renders the sample and checks the start note of every drawn tie: both of voice 1's B2 ties and both of voice 2's C2 ties. The test fails on develop, where only three ties are drawn, and passes with the fix. All 411 tests pass.
- **Visual-regression self-diff:** I built a develop baseline and this branch, ran `generate:blessed` / `generate:current` (all 337 samples), then `npm run test:visual`. **1 / 337 samples differ**: the new sample, by the added tie (136 px). Nothing else moves.


sschmidTU (approved):
LGTM, merging.
No visual regressions.

I'll add one follow-up commit fixing an edge case/regression not caught in the test samples:
there's a mirrored encoding to the new test sample where the hidden note is the one carrying the tie and the visible unison note is in the earlier voice.
Develop draws that tie only by accident of the pitch lookup. The PR drops it, so the quarter C2 reads as struck again.
Which head the engraver hides is arbitrary with two filled heads (MuseScore shows the same picture either way), so it shouldn't decide whether the tie is drawn.

